PR #18, opened 2021-03-30
Nexys a7 example

AWenzel83:
I created example projects for the Digilent Nexys A7 and Nexys 4 DDR boards.
Since this is my first pull request, I hope its the right way to do it


stnolting:
Awesome! Thank you very much! 👍

Issue #24, opened 2021-05-12
neorv32_wdt component not VHDL-08 compliant


tmeissner:
During experiments to synthesize the project with GHDL & Yosys I found that the Watchdog timer unit cannot analyzed when VHDL-08 standard is used. The record type for the control register uses the now reserved keyword `force`:

```vhdl
  -- control register --
  type ctrl_reg_t is record
    enable  : std_ulogic; -- 1=WDT enabled
    clk_sel : std_ulogic_vector(2 downto 0);
    mode    : std_ulogic; -- 0=trigger IRQ on overflow; 1=trigger hard reset on overflow
    rcause  : std_ulogic; -- cause of last system reset: '0' = external, '1' = watchdog
    reset   : std_ulogic; -- reset WDT
    force   : std_ulogic; -- force action
    lock    : std_ulogic; -- lock control register
  end record;
```

Changing `force` to a non-reserved word resolves the problem and neorv32 can be synthesized with Yosys & GHDL.

If you're interested I could contribute to synthesize the project with Yosys & GHDL.

Nice project by the way 🙂

stnolting:
Thanks for the hint! I will fix that.

> If you're interested I could contribute to synthesize the project with Yosys & GHDL.

That would be great! Experimenting with the GHDL Yosys plugin is already on my to-do list, but I haven't really looked at it yet...

> Nice project by the way 🙂

Thanks again 😉

tmeissner:
I have a IceZero board with a Lattice ICE40HX on it which is supported by the open source flow. I will try to implement neorv32 on it and make a contribution to the `boards/` folder.

umarcor:
I have an icestick (maybe too small), a FOMU (UP5K) and an EOS S3 (maybe too small). I have the icestick and FOMU toolchains ready. Either on Windows (https://hdl.github.io/MINGW-packages/) or through containers (https://hdl.github.io/containers). I helped setup CI in [antonblanchard/microwatt](https://github.com/antonblanchard/microwatt). I'd be happy to jump in and help enhance the CI plumbing in this regard, after Torsten is done with his contributions (so that we don't overlap).

Moreover, in the mid term, I would be interested in testing neorv32 on PYNQ-Z1, microzed and KC705.

@stnolting I pinged you in [gitter.im/vhdl/General](https://gitter.im/vhdl/General). Would be nice to have a chat!

stnolting:
@tmeissner 

[...]

PR #25, opened 2021-05-15
Add open source implementation flow for Upduino v3 board

tmeissner:
This is an attempt to add an implementation flow for the Upduino v.3 board which uses free & open source tools (GHDL, Yosys, nextpnr, Icestorm) only.

Some files were added:

* Makefile (obvious)
* sb_ice40_components.vhd (VHDL component definitions for FPGA hard macros, for Yosys/iceCube style & Radiant)
* yosys_ice40_components.vhd (Components which wrap the Radiant style macros to the Yosys (iceCube) ones) 
* system_pll.vhd (Hopefully functional similar to the verilog version)
* neorv32_upduino_v3.pcf (Constraint file for the nextpnr P&R tool)

Yosys uses the component definitions which the former Lattice tool (iceCube) used. That's the reason for the wrapper components in the yosys_ice40_components.vhd file. I tried to not change the original files, only the `CURRENT_MODE` generic was changed to a equivalent value `"1"` -> `"0b1"`.

I get the design fitted on the FPGA when I disable the TRNG (or maybe another) peripheral. If no feature is disabled, fitting fails with 101% resource allocation.

The resulting bin file is named like the existing one, but with number 2: `neorv32_upduino_v3_impl_2.bin`.

I used the `hdlc/impl` Docker image of the [hdlc](https://github.com/hdl/containers) project. It contains all necessary tools.

The PR is in draft mode and open for hints & discussion 🙂

### DIFF
--- a/boards/UPduino_v3/.gitignore
+++ b/boards/UPduino_v3/.gitignore
@@ -18,3 +18,5 @@ source/*
 !*.rdf
 !*.pdc
 !*.bin
+!*.pcf
+!Makefile
--- a/boards/UPduino_v3/Makefile
+++ b/boards/UPduino_v3/Makefile
@@ -47,6 +47,7 @@ IMPL_NAME := ${DESIGN_NAME}_impl_2
 .PHONY: syn impl bin clean
 
 # Some phony targets for convenience
+all: bin
 syn: ${IMPL_NAME}.json
 impl: ${IMPL_NAME}.asc
 bin: ${IMPL_NAME}.bin
@@ -61,10 +62,12 @@ work-obj08.cf: neorv32-obj08.cf ${UPDUINO_SRC}
 	ghdl -a --std=08 --work=work ${UPDUINO_SRC}
 
 ${IMPL_NAME}.json: work-obj08.cf
-	yosys -m ghdl -p 'ghdl --std=08 --no-formal neorv32_upduino_v3_top; synth_ice40 -abc2 -dsp -json $@'
+	yosys -m ghdl -p 'ghdl --std=08 -gyosys=true --no-formal neorv32_upduino_v3_top; synth_ice40 -abc2 -dsp -json $@' 2>&1 | tee yosys-report.txt
 
 ${IMPL_NAME}.asc: ${IMPL_NAME}.json ${DESIGN_NAME}.pcf
-	nextpnr-ice40 --ignore-loops --up5k --package sg48 --json $< --pcf ${DESIGN_NAME}.pcf --asc $@
+	nextpnr-ice40 --ignore-loops --timing-allow-fail \
+	  --up5k --package sg48 --json $< --pcf ${DESIGN_NAME}.pcf \
+	  --asc $@ 2>&1 | tee nextpnr-report.txt
 
 ${IMPL_NAME}.bin: ${IMPL_NAME}.asc
 	icepack $< $@
--- a/boards/UPduino_v3/Makefile
+++ b/boards/UPduino_v3/Makefile
@@ -1,0 +1,73 @@
+NEORV32_PKG := ../../rtl/core/neorv32_package.vhd
+
+NEORV32_SRC := \
+  neorv32_dmem.ice40up_spram.vhd \
+  neorv32_imem.ice40up_spram.vhd \
+  ../../rtl/core/neorv32_bootloader_image.vhd   \
+  ../../rtl/core/neorv32_application_image.vhd   \
+  ../../rtl/core/neorv32_boot_rom.vhd   \
+  ../../rtl/core/neorv32_bus_keeper.vhd   \
+  ../../rtl/core/neorv32_busswitch.vhd   \
+  ../../rtl/core/neorv32_cfs.vhd   \
+  ../../rtl/core/neorv32_cpu.vhd   \
+  ../../rtl/core/neorv32_cpu_alu.vhd   \
+  ../../rtl/core/neorv32_cpu_bus.vhd   \
+  ../../rtl/core/neorv32_cpu_control.vhd   \
+  ../../rtl/core/neorv32_cpu_cp_bitmanip.vhd   \
+  ../../rtl/core/neorv32_cpu_cp_fpu.vhd   \
+  ../../rtl/core/neorv32_cpu_cp_muldiv.vhd   \
+  ../../rtl/core/neorv32_cpu_decompressor.vhd   \
+  ../../rtl/core/neorv32_cpu_regfile.vhd   \
+  ../../rtl/core/neorv32_gpio.vhd   \
+  ../../rtl/core/neorv32_icache.vhd   \
+  ../../rtl/core/neorv32_mtime.vhd   \
+  ../../rtl/core/neorv32_nco.vhd   \
+  ../../rtl/core/neorv32_neoled.vhd   \
+  ../../rtl/core/neorv32_pwm.vhd   \
+  ../../rtl/core/neorv32_spi.vhd   \
+  ../../rtl/core/neorv32_sysinfo.vhd   \
+  ../../rtl/core/neorv32_top.vhd   \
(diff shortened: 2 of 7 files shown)


stnolting:
Great work and thx for contributing! 👍 

I do not have the toolchains available yet to experiment with your setup configuration. But looking t https://github.com/hdl/containers / https://github.com/stnolting/neorv32/issues/24#issuecomment-841221221 this should be no big deal.

> I get the design fitted on the FPGA when I disable the TRNG (or maybe another) peripheral. If no feature is disabled, fitting fails with 101% resource allocation.
The resulting bin file is named like the existing one, but with number 2: neorv32_upduino_v3_impl_2.bin.

Im curious. Were there any timing problems?
Do you have an ice40up5k available to test the bitstream on real hardware? If not, could you post your bitstream?
I assume it is compatible to the plain Lattice Radiant / iceCube programmer, right? 🤔

tmeissner:
> Im curious. Were there any timing problems?

There are problems to reach the 24 MHz of the original design. I assume it's because the design uses almost all (98% without TRNG) resources of the device. So, the P&R tool hasn't much margin for routing the wires. I reaches almost 20 MHz, so I've changed the PLL config to generate 20 MHz (not pushed yet).

The result table in [exemplary FPGA setups](https://github.com/stnolting/neorv32#neorv32-processor---exemplary-fpga-setups) suggests that the Radiant tools also don't reach the 24 MHz target frequency. Maybe a clock of 20 MHz would be better for both tool chains.

> Do you have an ice40up5k available to test the bitstream on real hardware? If not, could you post your bitstream?

I don't have such a device, you can download the latest bin file (20 MHz, w/o TRNG) at:
* [neorv32_upduino_v3_impl_2.bin](https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2.bin)
* [neorv32_upduino_v3_impl_2.bin.sha512](https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2.bin.sha512)

> I assume it is compatible to the plain Lattice Radiant / iceCube programmer, right?

I think it should be compatible to the result of the Lattice toolchain.

stnolting:
> The result table in exemplary FPGA setups suggests that the Radiant tools also don't reach the 24 MHz target frequency. Maybe a clock of 20 MHz would be better for both tool chains.

The f_max there refers to a different (old) setup. I really need to update that table.

> There are problems to reach the 24 MHz of the original design. I assume it's because the design uses almost all (98% without TRNG) resources of the device. So, the P&R tool hasn't much margin for routing the wires. I reaches almost 20 MHz, so I've changed the PLL config to generate 20 MHz (not pushed yet).

Good point. How about adding your yosys setup to a new boards folder like `UPduino_v2-yosys`?

> I don't have such a device, you can download the latest bin file (20 MHz, w/o TRNG) at:

Thanks for the bitstream! Ive programmed my UPduino v3 board but unfortunately nothings happened. I think there might be a problem with the PLL configuration. If the PLL does not lock the processor stays in reset all the time.

Maybe it would be better to start with a simpler setup that just uses the on-chip HF oscillator running at 12 MHz as clock source - so no PLL at all. We could use a shift register that is initialized with all zeros by the bitstream and feed the MSB to the processor reset input - this would provide a sufficient reset (just some cycles until the OSC runs stable).

tmeissner:
Okay, I currently try to build a FW with oscillator set to 12 MHz, without PLL & a shift register for reset:

```vhdl
  signal reset_reg : std_ulogic_vector(31 downto 0) := (others => '0');

begin

  HSOSC_inst : HSOSC
  generic map (
    CLKHF_DIV => "0b10" -- 12 MHz
  )
  port map (
    CLKHFPU => '1',
    CLKHFEN => '1',
    CLKHF   => hf_osc_clk
  );

  process (hf_osc_clk)
  begin
    if rising_edge(hf_osc_clk) then
      reset_reg <= reset_reg(30 downto 0) & '1';
    end if;
  end process;

  pll_rstn <= reset_reg(reset_reg'left);
  pll_clk  <= hf_osc_clk;
```

Another question: which peripheral isn't used by the program code and can be disabled to short run times? I assume TWI? Anything else?

stnolting:
> Okay, I currently try to build a FW with oscillator set to 12 MHz, without PLL & a shift register for reset

Looks good. 👍 

> Another question: which peripheral isn't used by the program code and can be disabled to short toolchain run times? I assume TWI? Anything else?

That depends on the program code you want to run. Most of the example programs require at least `UART0`. The bootloader is happier when having `MTIME` and it is always good to have some minimal IO provided by `GPIO`. The remaining peripheral modules can be safely disabled.

tmeissner:
As I wrote on Gitter, I have a new FW with 12 MHz clock direct from the oscillator and the shift reg for reset. TWI, PWM, WDT & TRNG are disabled. Should be okay for the blinky-program.

* https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2_-_12mhz_without_pll_twi_pwm_wdt_trng.bin
* https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2_-_12mhz_without_pll_twi_pwm_wdt_trng.bin.sha512

Maybe we could talk on Gitter if these don't work.

stnolting:
It works! 👍 
Well, at least somehow 😄  The bootloader status LED is flashing - but slower than expected. Also, the UART sends garbage when using the default baud rate of 19200. But it seems like this is just a simple clock configuration issue.

grafik


The HF oscillator of your setup is running at 12 MHz, right?
Did you adapt the `f_clock_c` constant in `neorv32_upduino_v3_top.vhd` or the processor's `CLOCK_FREQUENCY` generic, respectively (`CLOCK_FREQUENCY` goes to a memory-mapped register that is read by the bootloader to configure the UART baud rate dividers)?

I've tried using the doubled baud rate and also using the halved one - but still only garbage. But I will further evaluate on that.

tmeissner:
Grrr, I forgot to change the `f_clock_c` constant from my last experiment with 20 MHz 😄 So, no suprise that the UART sends garbage 😏

stnolting:
Oh, haha - I was just trying to reverse engineer the actual clock speed with my oscilloscope connected to the status LED :smile: 
Anyway, your new bitstream works like a charm now and I think this can be merged!

How about re-placing your yosys-based setup into a new folder in `boards` so we can keep the "original" radiant setup? 🤔

tmeissner:
> Oh, haha - I was just trying to reverse engineer the actual clock speed with my oscilloscope connected to the status LED 😄
> Anyway, your new bitstream works like a charm now and I think this can be merged!

Good 😄 

> How about re-placing your yosys-based setup into a new folder in `boards` so we can keep the "original" radiant setup? 🤔

Yeah, I will create a new folder and move the files there. I have waited for a working binary file. After merge I can look after the PLL problems.

stnolting:
Cool, thanks!
Feel free to modify this PR or to create a new one.

tmeissner:
I have a new test firmware with a PLL generating a 18 MHz from the 12 MHz clock of the oscillator. With this settings, timing closure was reached: `Max frequency for clock 'system_pll_inst:20': 21.88 MHz (PASS at 18.00 MHz)`. Maybe this time I've got the PLL settings right 🤞 

You can dl the firmware under:
* [neorv32_upduino_v3_impl_2_-_18mhz_without_twi_pwm_wdt_trng.bin](https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2_-_18mhz_without_twi_pwm_wdt_trng.bin)
* [neorv32_upduino_v3_impl_2_-_18mhz_without_twi_pwm_wdt_trng.bin.sha512](https://downloads.goodcleanfun.de/neorv32_upduino_v3_impl_2_-_18mhz_without_twi_pwm_wdt_trng.bin.sha512)

stnolting:
I have just tested the new 18 MHz bitstream including the PLL and it works! 👍

tmeissner:
I will create a new PR with the Yosys based flow in a separate folder.

PR #26, opened 2021-05-18
Build HTML datasheet and publish to GitHub Pages along with Doxygen output

umarcor:
In this PR, several enhancements to the CI plumbing are contributed:

- The two existing workflows for the PDF datasheet and the Doxygen output are merged into a single workflow. That allows visualizing the dependencies between them (see below).
- asciidoctor is used for building the datasheet in HTML too. In order to do so, most of the content from `neorv32.adoc` is moved to `content.adoc`. A different entrypoint is used for the HTML build (`index.adoc`).
- The HTML datasheet is published to GitHub Pages along with the Doxygen output (in subdir `sw`). That is, the artifacts from the two jobs are downloaded, reorganised and pushed.
- The shields/badges in the README are updated, and some of them are added to the HTML datasheet.

See [umarcor.github.io/neorv32](https://umarcor.github.io/neorv32/) and [umarcor.github.io/neorv32/sw](https://umarcor.github.io/neorv32/sw).

@stnolting, in the current state, the PDF datasheet will be uploaded to branch gh-pages (https://github.com/umarcor/neorv32/tree/gh-pages). That might be not desirable, since it is also added to branch master, and it is not good practice to add PDFs to git repositories. Alternatively, a GitHub Release might be used for keeping the latest PDF(s) available. What do you think?

PR #31, opened 2021-05-20
Add UPduino v3 variant using open source toolchain (2nd try)

tmeissner:
This is an attempt to add an implementation flow for the Upduino v.3 board which uses free & open source tools (GHDL, Yosys, nextpnr, Icestorm) only.

I used the [hdlc](https://github.com/hdl/containers) Docker image of the hdlc project. It contains all necessary tools.

The resulting bin file is left because nextPNR is still trying to route all the wires 😄


stnolting (approved):
Looks really great! 👍 🚀

umarcor:
> FPGA binary file added. I've noticed that there is also an free alternative to the programming tool, but I don't find it in the docker image I used. @umarcor do you know more about that?

Do you mean iceprog from the icestorm set of tools? That is included in the `hdlc/prog` image, but not in others. That is because using board programming tools from containers is not straightforward. On GNU/Linux, it's rather easy because you can just share the device as a volume (only need to pass the specific device as an argument to the makefile). However, on Windows and macOS, USB devices are not directly accesible by the containers. USB/IP needs to be used, and no open source solution exists. Moreover, there are quite a lot of programming tools, almost one for each vendor, family or board. Therefore, I focused my efforts on other areas and I recommend users to install programming tools locally. You can see that many of those tools are supported in https://hdl.github.io/MINGW-packages/#_packagestools but shown as To Do in https://hdl.github.io/containers/#_tools_and_images.

Nonetheless, I have used containers such as hdlc/prog for programming boards both from GNU/Linux and Windows, so I can confirm it works, in case you want to try it.

> And: @umarcor maybe we (you 😄 ) can try to add a GHA flow for this?

As commented in https://github.com/stnolting/neorv32/pull/26#issuecomment-845545669, I updated the CI and Makefiles in fomu-workshop in preparation for reusing that knowledge here. Hence, we (you :wink:) can start by copying the following step to a workflow (maybe create a new workflow for this): https://github.com/umarcor/fomu-workshop/blob/ci/containers/.github/workflows/test.yml#L69-L83. You need to replace `args: ./.github/hdl-tests.sh` with the entrypoint to this task. You can remove `GHDL_PLUGIN_MODULE: ghdl` because that's not used by your makefile. 

For instance:

```yml
name: Implementation
on:
  push:
  pull_request:
  schedule:
    - cron: '0 0 * * 5'
  workflow_dispatch:
      
jobs:

  all-in-one:
    name: '🛳️ All-in-one'
    runs-on: ubuntu-latest
    steps:

    - name: '🧰 Repository Checkout'
      uses: actions/checkout@v2

    - name: '⚙️ Generate bitstream' 
      uses: docker://ghcr.io/hdl/debian-buster/impl
       with:
         args: make -C boards/UPduino_v3_ghdl-yosys-nextpnr/ all

    - name: '📤 Upload Artifact: HTML'
      uses: actions/upload-artifact@v2
      with:
        name: UPduino_v3_ghdl-yosys-nextpr
        path: boards/UPduino_v3_ghdl-yosys-nextpnr/neorv32_upduino_v3_impl_1.bin
```

boards/UPduino_v3_ghdl-yosys-nextpnr/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	ghdl -a --std=08 --work=neorv32 ${NEORV32_PKG} ${NEORV32_SRC}
+
+work-obj08.cf: neorv32-obj08.cf ${UPDUINO_SRC}
+	ghdl -a --std=08 --work=work ${UPDUINO_SRC}
+
+${IMPL_NAME}.json: work-obj08.cf
+	yosys -m ghdl -p 'ghdl --std=08 --no-formal neorv32_upduino_v3_top; synth_ice40 -abc2 -dsp -json $@' 2>&1 | tee yosys-report.txt
+
+${IMPL_NAME}.asc: ${IMPL_NAME}.json ${DESIGN_NAME}.pcf
+	nextpnr-ice40 --ignore-loops --timing-allow-fail \
```
umarcor, line 67: Does the design require combinational loops for it to work?

boards/UPduino_v3_ghdl-yosys-nextpnr/sb_ice40_components.vhd
```diff
@@ -0,0 +1,137 @@
...
+    RESETB          : in  std_logic;
+    LATCHINPUTVALUE : in  std_logic;
+    SDO             : out std_logic;
+    SDI             : in  std_logic;
+    SCLK            : in  std_logic
+  );
+  end component;
+
+
+  component  RGB
```
umarcor, line 54: Declaring RGB as an alias of SB_RGBA_DRV (or the other way) should avoid the duplication.

boards/UPduino_v3_ghdl-yosys-nextpnr/system_pll.vhd
```diff
@@ -0,0 +1,58 @@
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+library iCE40UP;
+use iCE40UP.components.all; -- for device primitives and macros
+
+entity system_pll is
```
umarcor, line 8: Is this `system_pll` entity defined/used somewhere else? If so, would it make sense to define the architecture only here and have the entity in some common location? This is not specific to this PR, but an open question in preparation for upcoming contributions. Precisely, the Upduino and Fomu boards have the same device, but are different boards. Hence, it would make sense to have device specific HDL sources on the one hand, and board specific ones on the other. See, for instance, `board` and `device` folders in https://github.com/dbhi/vboard/tree/main/vga.

boards/UPduino_v3_ghdl-yosys-nextpnr/Makefile
stnolting: Basically: No.
But the true-random number generator (TRNG) uses combinatorial loops for the entropy generation.

tmeissner: Yes, the option is needed when TRNG is enabled. Without it, nextPNR can't calculate the timing of the design.

boards/UPduino_v3_ghdl-yosys-nextpnr/sb_ice40_components.vhd
tmeissner: Ups, the duplication is an artifact of my 1st attempt. The Radiant components aren't needed anymore as this is Yosys only.

boards/UPduino_v3_ghdl-yosys-nextpnr/system_pll.vhd
tmeissner: In this design, the `system_pll.vhd` component is a simple wrapper for the PLL macro. It could be replaced by directly using the PLL macro in the top level design, maybe the better solution.

tmeissner: Done

boards/UPduino_v3_ghdl-yosys-nextpnr/sb_ice40_components.vhd
tmeissner: Unused Radiant components are removed now.

tmeissner:
I have done the suggested changes and updated the resulting binary. One side-effect: by removing the `-abc2` option of the `synth_ic40` command, a few less cells are used and the router finishes in 1/3 of the time before 🎉  (but still 347.30 sec on my Mac using docker).1

I would prefer to merge this first and possibly adding CI later.

umarcor:
I don't think it's necessary to add the bitstream. As soon as CI is added, the bitstream will be available as an artifact and in the nightly pre-release. Nevertheless, I see it was added in `UPduino_v2` too, so it's acceptable I guess.

> I would prefer to merge this first and possibly adding CI later.

Sounds good to me.

stnolting:
Ready to merge! :D

The new hardware modules of the on-chip debugger are missing (`rtl/core/neorv32_debug_*.vhd`) - but that should not be relevant as they are no included into the design by default and will be treated as "black boxes", right? 🤔
(I don't want to touch the files in your PR as I am still working through your setup...)

tmeissner:
I have rebased the branch to master and updated the UPduino top level. Now the branch is up-to-date and can be merged.

PR #38, opened 2021-05-23
[docs] add favicon

umarcor:
This PR adds a favicon to the HTML site, in order to make it easier to spot the NEORV32 tabs in the browser.


stnolting:
Now that's cool! 😎 
Just one thing: could you move the icon into `docs/figures`? Or is there some reason it has to be / should be in `docs`? 🤔

umarcor:
> Just one thing: could you move the icon into `docs/figures`? Or is there some reason it has to be / should be in `docs`? 🤔

It's up to you. The icon can be anywhere in the codebase and anywhere in the generated site. Do you want to move into figures in both cases? Or do you want to move the source into figures, but keep it in the root when deployed? Let me know and I'll update this PR.

The reason for keeping it out from figures is because, strictly, it's not a figure. It's an image, but not part of the documentation per se. By the same token, I tried to keep `docinfo.html` in `doc`, instead of `doc/src_adoc`. However, in that case, the location needs to be the same as `index.adoc`.

stnolting:
Actually, it is not really critical *where* it is. I thought it might be nice to have it in `docs/figures` so one could use it in the asciidoctor theme file and maybe place it in the pdf's header 🤔

umarcor:
That is the case now :wink:. I found a better approach. Instead of writing the docinfo file, just setting the `:favicon:` attribute.

PR #39, opened 2021-05-23
Add GitHub Actions workflow 'Implementation'

umarcor:
As discussed in #31, this PR adds a workflow named Implementation for generating the UPduino v3 bitstream using the open source tools. Precisely, a single container from [hdl/containers](https://github.com/hdl/containers) is used. That's why the job is named "All-in-one" (see https://hdl.github.io/containers/#_all_in_one_images). The bitstream, along with the yosys and nextpnr reports (logs) are uploaded as an artifact. However, the artifact is not published anywhere else. That is, users willing to download the bitstream/reports need to go to https://github.com/stnolting/neorv32/actions/workflows/Implementation.yml.

BTW, I renamed workflow "Processor Check" to "Processor" in order to reduce the width of the labels in the readme. Moreover, all workflows are checks in some sense, so I think the term is not very descriptive.

### DIFF
--- a/.github/workflows/Implementation.yml
+++ b/.github/workflows/Implementation.yml
@@ -1,0 +1,33 @@
+name: Implementation
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 5'
+  workflow_dispatch:
+
+jobs:
+
+  all-in-one:
+    runs-on: ubuntu-latest
+    name: '🛳️ All-in-one'
+
+    steps:
+
+    - name: '🧰 Repository Checkout'
+      uses: actions/checkout@v2
+
+    - name: '⚙️ Generate bitstream'
+      uses: docker://ghcr.io/hdl/debian-buster/impl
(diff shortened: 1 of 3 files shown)


stnolting:
> As discussed in #31, this PR adds a workflow named Implementation for generating the UPduino v3 bitstream using the open source tools. Precisely, a single container from hdl/containers is used. That's why the job is named "All-in-one" (see https://hdl.github.io/containers/#_all_in_one_images). The bitstream, along with the yosys and nextpnr reports (logs) are uploaded as an artifact. However, the artifact is not published anywhere else. That is, users willing to download the bitstream/reports need to go to https://github.com/stnolting/neorv32/actions/workflows/Implementation.yml.

Great! Thank you very much!

> BTW, I renamed workflow "Processor Check" to "Processor" in order to reduce the width of the labels in the readme. Moreover, all workflows are checks in some sense, so I think the term is not very descriptive.

I agree :wink:

PR #40, opened 2021-05-24
Use asciidoctor-diagram with wavedrom support

umarcor:
For some time, I wanted to learn how to use asciidoctor-diagram with wavedrom support, so I gave it a try.

For illustration purposes, in the last commit of this PR `wavedrom.adoc` is added and it is included in `content.adoc`. It shows a timing diagram, a bit-field diagram and a logic circuit, all generated through wavedrom. It also shows how to include a wavedrom figure from a JSON5 file, instead of writing the wave/diagram in the adoc source.

It works both for the HTML and the PDF outputs. See page 7 of [NEORV32.pdf](https://github.com/stnolting/neorv32/files/6529157/NEORV32.pdf).

Nevertheless, this PR is not ready to be merged because the `wavedrom.adoc` file is just a demo. @stnolting, let me know if you find this feature interesting. You can make a copy of the `wavedrom.adoc` file, and I'll remove it from this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ docs/doxygen_build
 
 # HTML datasheet
 /docs/index.html
+
+# asciidoctor cache
+/docs/.asciidoctor
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,28 @@
+# Generate PDF datasheet
 pdf:
 	cd docs; \
 	[ -f revnumber.txt ] && REVNUMBER='-a revnumber='"$$(cat revnumber.txt)" || unset REVNUMBER; \
 	asciidoctor-pdf $$REVNUMBER \
 	  -a pdf-theme=src_adoc/neorv32-theme.yml \
+	  -r asciidoctor-diagram \
 	  src_adoc/neorv32.adoc \
 	  --out-file NEORV32.pdf
 
+# Generate HTML datasheet
 html:
 	cd docs; \
 	[ -f revnumber.txt ] && REVNUMBER='-a revnumber='"$$(cat revnumber.txt)" || unset REVNUMBER; \
 	asciidoctor $$REVNUMBER \
+	  -r asciidoctor-diagram \
 	  src_adoc/index.adoc \
 	  --out-file index.html
 
+# Generate revnumber.txt for overriding the revnumber attribute in 'pdf' and/or 'html'
 revnumber:
 	if [ `git tag -l | grep nightly` ]; then git tag -d nightly; fi
-	git describe --long --tags  | sed 's#\([^-]*-g\)#r\1#;' > docs/revnumber.txt
+	git describe --long --tags | sed 's#\([^-]*-g\)#r\1#;' > docs/revnumber.txt
 	cat docs/revnumber.txt
 
+# Build 'pdf' and 'html' in an 'asciidoctor-wavedrom' container
 container: revnumber
-	docker run --rm -v /$(PWD)://documents/ asciidoctor/docker-asciidoctor make pdf html
+	docker run --rm -v /$(PWD)://documents/ btdi/asciidoctor make pdf html


stnolting:
This is great! So we could remove the actual wavedrom images and replace them with the according JSON files and let the CI job do all the rendering? 👍 

I also did not know that wavedrom can make nice bit field diagrams. It would be great to use them in the documentation when talking about bits in control register and stuff like that. Tables are good, but bit field images might be more intuitive and understandable.

umarcor:
> This is great! So we could remove the actual wavedrom images and replace them with the according JSON files and let the CI job do all the rendering? 👍

That is the case. Strictly, we don't need the plugin. We could keep the JSON sources, generate the SVG/PNG with wavedrom-cli and use those in the adoc sources. However, the plugin handles that for us. Including the fact that images are embedded in the PDF but not in the HTML (that's optional). Moreover, the plugin allows writing the waves in code blocks, which is useful for small figures. Overall, I find it to be a very nice integration.

> I also did not know that wavedrom can make nice bit field diagrams. It would be great to use them in the documentation when talking about bits in control register and stuff like that. Tables are good, but bit field images might be more intuitive and understandable.

Exactly. At first, I wasn't unsure about the utility in this case. However, I found that you have some bitfield descriptions as tables, indeed. My concern about this bitfields generated by wavedrom is links/references. I don't know whether those are supported. I see how they would work with the HTML output (SVGs), but not in the PDF. Therefore, it might be desirable to keep both visualisations: the bitfield *and* the table. The bitfield allows easily understanding the layout, while the table describes the details and provides refs.

So, will you let me know when you make a copy of `wavedrom.adoc` and `waves/demo.json5`? Then, I'll remove the last commit from this PR, you can merge and we can start using the feature in multiple branches (no need to do everything at once).

umarcor:
> let the CI job do all the rendering?

It's not only the CI job. If you have the tools installed locally, you can do `make pdf html`. If you don't, but you have docker, `make container` will run everything inside a container and generate the HTML and PDF outputs for you.

stnolting:
I just want to check if I get this right...

The whole wavedrom script can be placed *in line* int he adoc sources, right?
So no need to have separate wavedrom files?

The rendered SVG images are optionally embedded into the HTML result?
So, *where* are they / where will they be actually?

stnolting:
I think it would not harm to have your changes to the root makefiles merged into the repository - even there are no in-line wavedrom code snippets yet.

I am wondering: What is the purpose of the `demo.json5` file? just an example?

umarcor:
> I think it would not harm to have your changes to the root makefiles merged into the repository - even there are no in-line wavedrom code snippets yet.

I cleaned this branch. It should be harmless to merge, but it provides the feature.

> I just want to check if I get this right...

> The rendered SVG images are optionally embedded into the HTML result?
> So, _where_ are they / where will they be actually?

Clone branch `feat/docs/wavedrom-demo` from my fork (which contains the examples removed from this PR) and execute `make container`. You will understand it instantly. If the figure has some specific name, `docs/figures/NAME.svg` is created; otherwise, `docs/figures/diag-HASH.svg`. Then, those images are used in the HTML and/or the PDF. Yet, generation and usage of the images is done internally by asciidoctor.

For embedding the images in the HTML, see https://docs.asciidoctor.org/asciidoctor/latest/html-backend/manage-images/#embed-images-with-the-data-uri-attribute.

> The whole wavedrom script can be placed _in line_ int he adoc sources, right?
> So no need to have separate wavedrom files?

Yes. Both use cases are supported: inline, or in a separated file.

> I am wondering: What is the purpose of the demo.json5 file? just an example?

It is just an example for showcasing that you can import from a file, instead of writing inline. You can see that the content is almost the same as the inline timing diagram.

Why would you have the waves defined in separated files? Because those can be potentially generated automatically. Say you use an VCD to wavedrom converter for extracting part of a simulation waveform. You don't want to use sed or manually replace the content of the wave in the adoc files. By the same token, you might want to use those waveform files somewhere else. For instance, they could be used as golden examples for comparing a simulation output. I'm not suggesting to do it, just explaining the possibility.

Anyway, if you are going to define them manually, and they are small, it makes sense to do it inline.

umarcor:
Moreover, if you check TerosHDL, you will find that they support describing the waveforms or bit-fields in the VHDL sources. They then parse/extract that for generating the documentation. Therefore, it would potentially be possible not to write the bit-fields in the adoc source nor manually, but having them generated from VHDL. Again, this is not something I'm suggesting to use here, just explaining why importing from a file makes sense.

stnolting:
> Why would you have the waves defined in separated files? Because those can be potentially generated automatically. Say you use an VCD to wavedrom converter for extracting part of a simulation waveform. You don't want to use sed or manually replace the content of the wave in the adoc files. By the same token, you might want to use those waveform files somewhere else. For instance, they could be used as golden examples for comparing a simulation output. I'm not suggesting to do it, just explaining the possibility.

Hmm... good point. I have to think about that :wink:

> Moreover, if you check TerosHDL, you will find that they support describing the waveforms or bit-fields in the VHDL sources. They then parse/extract that for generating the documentation. Therefore, it would potentially be possible not to write the bit-fields in the adoc source nor manually, but having them generated from VHDL. Again, this is not something I'm suggesting to use here, just explaining why importing from a file makes sense.

That's a nice feature. I always wanted to check out TerosHDL - maybe now it's the time 😄

PR #41, opened 2021-05-24
[docs] set revnumber through git describe

### LINKED ISSUE #33: Add commit SHA to the datasheet(s)
Since the datasheet is uploaded automatically from CI, multiple artifacts/assets might be difficult to tell apart. It would be desirable to have the commit SHA added to the frontpage/subtitle of the PDF and HTML datasheets, so that users can know which specific version of the repo they correspond to. It can also be useful for developers to understand bug reports or feature requests from users with probably outdated versions of the documentation.

However, I did not search how to do it with asciidoc(tor) yet. I'm creating this issue as a remainder.

umarcor:
Close #33

In this PR, `git describe` is used for setting a unique revnumber for each commit. Since the container does not include git, a `docs/revnumber.txt` file is created outside. If it exists, it's used for overriding the revnumber attribute through the CLI. Otherwise, the revnumber set in the adoc sources is used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,20 @@
 pdf:
-	asciidoctor-pdf -a pdf-theme=docs/src_adoc/neorv32-theme.yml docs/src_adoc/neorv32.adoc --out-file docs/NEORV32.pdf
+	cd docs; \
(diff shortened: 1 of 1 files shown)


stnolting:
👍 👍 👍

PR #61, opened 2021-06-10
Reorganise setups

umarcor:
In this PR, Arty and Nexys example setups/scripts/READMEs are updated. Now, when `create_project.tcl` is sourced, the bitstream is generated straightaway. By the way, the duplicated scripts for Nexys boards are merged into a single one. 

Then, subdirs `setups/vivado`, `setups/quartus` and `setups/radiant` are created. The "board" subdirs are reorganised/moved accordingly. The structure we have corresponds to the tools, not necessarily to the boards. For instance, UPduino is supported both in `setups/radiant/UPduino_v3` and `setups/osflow/UPduino_v3`.

A `setup/vivado/README.md` is added for reducing the duplication in Arty and Nexys READMEs. Both batch and GUI modes are explained.


stnolting:
Very nice! Thank you for cleaning up 😉

umarcor:
Nice! Note that the Fomu PR is "broken" now. Do not worry. I'll fix (rebase) that :wink:

stnolting:
> Nice! Note that the Fomu PR is "broken" now.

Right. But I thought it does not matter since FOMU is still under construction.
I'm sorry if this has caused you extra work. :worried:

umarcor:
> Right. But I thought it does not matter since FOMU is still under construction.
> I'm sorry if this has caused you extra work. 😟

No worries! That was expected :wink:

Issue #62, opened 2021-06-10
Processor Boot Concept


stnolting:
This is a "design note" inspired by #56 (especially https://github.com/stnolting/neorv32/pull/56#issuecomment-858008281 and https://github.com/stnolting/neorv32/pull/56#issuecomment-858149579).

Currently, the processor provides 3 boot options:

1. Use the bootloader (`BOOTLOADER_EN => true`) to upload executables via UART or from external SPI flash. In this scenario the internal IMEM is implemented as **_UNinitialized_ RAM**.
2. No bootloader (`BOOTLOADER_EN => true`). IMEM is implemented as **_initialized_ RAM** (initialized with actual executable). The content of the RAM can be modified by software as it is a RAM.
3. No bootloader (`BOOTLOADER_EN => true`). IMEM is implemented as **_initialized_ ROM** (`MEM_INT_IMEM_ROM => true`, initialized with actual executable).

I do not see any practical use case for option 2, so I think this could be removed?! 🤔 

If option 2 is not needed anymore, the boot configuration could be simplified by replacing `BOOTLOADER_EN` and `MEM_INT_IMEM_ROM` generics by a single boot config generic (what name? what type?).

umarcor:
I'm a graphical person, so I did the following diagram:

![neorv32_boot](https://user-images.githubusercontent.com/38422348/121559487-e3827100-ca16-11eb-8886-ed7b1ddce08c.png)

- Gray boxes are ROMs and **need** to be initialised in the bitstream, since there is no explicit mechanism for loading them at runtime.
- Orange boxes can be either ROMs or RAMs, and can be empty or initialised. Arrows show which can be loaded at runtime.
- Blue boxes are RAMs, and can be empty or initialised.

|  | B | BD | I | ID | BI | BI-U | BI-S | BID | BID-U | BID-S |
|---|---|---|---|---|---|---|---|---|---|---|
| BOOTLOADER_EN    | true | true | false | false | true | true | true | true | true | true |
| MEM_INT_IMEM_EN  | false | false | true | true | true | true | true | true | true | true |
| MEM_INT_IMEM_ROM | - | - | true | true | true | false | false | true | false | false |
| MEM_INT_DMEM_EN  | false | true | false | true | false | false | false | true | true | true |
| IO_UART0_EN      | - | - | - | - | - | true | - | - | true | - |
| IO_SPI_EN        | - | - | - | - | - | - | true | - | - | true |

[...]

PR #63, opened 2021-06-10
Add Fomu

umarcor:
This is a subset of #56, so we can merge it and focus on separated discussions about further enhancements. 

The MinimalBoot and UP5KDemo examples in this PR do work, in the sense that the blinks produced by the bootloader are seen. However, the software application does not work because SPRAM is used for IMEM. Loading through UART or SPI was not implemented yet. We will address those challenges after #62.

The software in the Minimal examples works because the default IMEM implementation is used in ROM mode, instead of SPRAM. However, 97% of the resources is used, so something might be incorrectly inferred. We will address that in upcoming PRs too.


stnolting:
Awesome! 🚀

PR #65, opened 2021-06-10
Add VUnit

umarcor:
This is another subset of #47. As discussed, the existing testbench is copied and it's modified for supporting VUnit's Python runner. In following PRs, the proposal is to add advanced verification features and Verification Components to the VUnit testbench(es). As we add those, we can remove the equivalent hand-written test code. In this sense, testing of the UART1 is disabled in the `neorv32_tb.simple.vhd`, but it is tested in the `neorv32_tb.vhd` (run with VUnit). Moreover, the Processor workflow is modified, in order to run VUnit tests using the VUnit/vunit_action Action.


stnolting:
This looks quite complete, right?! :thinks:

umarcor:
It is. If you are good with it, go ahead and merge!

PR #67, opened 2021-06-11
rename logical library 'iCE40UP' to 'iCE40'

umarcor:
Although we've been using the sb_ice40_components source for UP5K devices only, some of those can be used with other ICE40 devices too. This PR renames the logical library name from `iCE40UP` to `iCE40`.

By the way, `.gitignore` is updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,4 @@ sw/image_gen/image_gen.exe
 *.o
 
 # example bitstreams
-/examples/*.bit
+/setups/examples/*.bit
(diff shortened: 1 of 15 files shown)


stnolting:
> Although we've been using the sb_ice40_components source for UP5K devices only, some of those can be used with other ICE40 devices too. This PR renames the logical library name from iCE40UP to iCE40.

Very good idea!

stnolting:
Lattice Radiant cannot resolve `iCE40` library if not explicitly assigned.
Reverting the library fixes in the plain-radiant example setup should not break anything, right? 😅

umarcor:
I see. So, when using radiant all the components libraries are automatically added to the project under the hood. You don't see them, but you can instantiate them. That makes sense. Sorry about breaking that 😟

stnolting:
No worries! ;)

PR #68, opened 2021-06-11
Tinyfpga-BX

umarcor:
This PR is based on #67.

Support for TinyFPGA-BX is added to `setups/osflow` and `setups/examples/Makefile`.

A MinimalBoot BoardTop is added. Due to memory inference issues, the bootloader is enabled, IMEM is disabled, and a custom implementation is provided for DMEM. Compared to the Fomu examples, the LP8K FPGA on the TinyFPGA-BX does *not* have SPRAM. I tried using the default dmem implementation, but that would fail. Therefore, I picked the alternative implementation discussed in #56. Since imem is used as a ROM, the default implementation is used.

This example can be synthesized and the resulting bitstream can be programmed. However, something seems wrong, since the software is not executed. I don't have a TinyFPGA-BX myself, so @juanmard is helping with testing. We tried setting the LED output in the BoardTop to a constant value. Hence, we can confirm that the toolchain works and synthesis is correct. However, the bootloader does not produce the expected blinks.

I will mark this PR as a draft until we have some working blink and until we discuss how to deal with the implementation of memories.


stnolting:
I like that! :+1:
I think the tinyFPGA boards are a great platform for beginner (just like FOMU): just the FPGA without any over-engineered on-board peripherals. Plus they are very affordable.

> A MinimalBoot BoardTop is added. Due to memory inference issues, the bootloader is enabled, IMEM is disabled, and a custom implementation is provided for DMEM. Compared to the Fomu examples, the LP8K FPGA on the TinyFPGA-BX does not have SPRAM. I tried using the default dmem implementation, but that would fail. Therefore, I picked the alternative implementation discussed in #56. Since imem is used as a ROM, the default implementation is used.

👍 

> This example can be synthesized and the resulting bitstream can be programmed. However, something seems wrong, since the software is not executed. I don't have a TinyFPGA-BX myself, so @juanmard is helping with testing. We tried setting the LED output in the BoardTop to a constant value. Hence, we can confirm that the toolchain works and synthesis is correct. However, the bootloader does not produce the expected blinks.

Hmm 🤔 
If it is not a reset/clock issue (which I doubt) it might be a problem with ROM implementation (again).
Does the status LED even light up?

juanmard:
> Hmm 🤔
> If it is not a reset/clock issue (which I doubt) it might be a problem with ROM implementation (again).
> Does the status LED even light up?
 
Hi, @stnolting
The status LED is down and not blinking.
I did some testing with the PLL clock hardware flashing independently of risc-v gpio and it works.
Surprisingly I tried the bootloader (should blink) and does not work either.
My next tests will go focused on obtaining information from UART with USB external conversor. ;-)

stnolting:
Hmm strange...

Some ideas:
* Make sure you are using the default bootloader image (`rtl/core/neorv32_bootloader_image.vhd`) from the repo's master branch
* Also make sure the bootloader is enabled: `BOOTLOADER_EN => true`
* Implement UART0 even if it is not connected to external pins: `IO_UART0_EN => true`
* Implement MTIME (this will make the bootloader's status LED go _flashing_): `IO_MTIME_EN => true`
* If using a PLL (also for reset) make sure it gets locked (-> route lock signal to external pin)

juanmard:
> Some ideas:
>     * Make sure you are using the default bootloader image (`rtl/core/neorv32_bootloader_image.vhd`) from the repo's master branch

It's ok, but not work yet.

> 
>     * Also make sure the bootloader is enabled: `BOOTLOADER_EN => true`
>
I did it... not work. 😞 
 
>     * Implement UART0 even if it is not connected to external pins: `IO_UART0_EN => true`
not work again...

>     * Implement MTIME (this will make the bootloader's status LED go _flashing_): `IO_MTIME_EN => true`
nothing... 😕

> 
>     * If using a PLL (also for reset) make sure it gets locked (-> route lock signal to external pin)
I've test it... it seem to be ok, I got _pll_rstn_ on a external pin and LED light on, so PLL is "locked" fine.

Next step: To connect a conversor and get information from UART. 😉 
Thanks @stnolting!!

stnolting:
I am sorry to hear that.

Maybe it is just a reset issue? Try using an external pin/button (low-active) instead of a PLL-locked approach for generating the processor reset.

I could also be a problem with the FPGA block RAM initialization. I am working on a new memory system right now (see #62). Hopefully, I can upload that tonight. :wink:

juanmard:
> Maybe it is just a reset issue? Try using an external pin/button (low-active) instead of a PLL-locked approach for generating the processor reset.

I did this test with a manual reset... same behaviour, not blink. 😕 

> I could also be a problem with the FPGA block RAM initialization. I am working on a new memory system right now (see #62). Hopefully, I can upload that tonight. 😉

I tested this new implementation... and nothing seems to change. 😕

stnolting:
Oh damn! 😅 

Which file are you using as top entity / processor top entity?

juanmard:
> Which file are you using as top entity / processor top entity?

**neorv32_TinyFPGA_BX_BoardTop_MinimalBoot.vhd** in *setups/examples*

stnolting:
@juanmard 

I finally managed to update the whole memory/boot concept. Maybe you can try updating to the latest version of this project. I also added some more documentation and updated the user guides. Maybe the sections regarding memory configuration and boot configuration can help?!
* https://stnolting.github.io/neorv32/#_memory_configuration
* https://stnolting.github.io/neorv32/#_boot_configuration

Here is a guide that is intended to work even on smallest platforms that do not feature any kind of UART connection:
https://stnolting.github.io/neorv32/ug/#_installing_an_executable_directly_into_memory

> neorv32_TinyFPGA_BX_BoardTop_MinimalBoot.vhd in setups/examples

I am not sure why your setup is not working 🤔 
But I will have a closer look at all the top templates and example setups again.

umarcor:
Some results:

# 1e240bcbc16d65849e916ca6b339feb7d7dc9e72

```
Info: Device utilisation:
Info:            ICESTORM_LC:  4499/ 7680    58%
Info:           ICESTORM_RAM:    20/   32    62%
Info:                  SB_IO:     5/  256     1%
Info:                  SB_GB:     8/    8   100%
Info:           ICESTORM_PLL:     1/    2    50%
Info:            SB_WARMBOOT:     0/    1     0%

Max frequency for clock 'pll_clk': 32.66 MHz
```

# d61d4e6eef90f1807f9baaee29e505b9b9b990bb

The same as the previous one, using the default `rtl/core/neorv32_dmem.vhd`, instead of `setups/osflow/devices/ice40/neorv32_dmem.alt.vhd`.

```
Info: Device utilisation:
Info:            ICESTORM_LC: 68714/ 7680   894%
Info:           ICESTORM_RAM:    12/   32    37%
Info:                  SB_IO:     5/  256     1%
Info:                  SB_GB:     8/    8   100%
Info:           ICESTORM_PLL:     1/    2    50%
Info:            SB_WARMBOOT:     0/    1     0%
```

# 86cc1e311b92c8f47f1e7994c72c1443c6c80d38

The same as 1e240bcb, before the updates of last two days.

```
Info: Device utilisation:
Info:            ICESTORM_LC:  4545/ 7680    59%
Info:           ICESTORM_RAM:    20/   32    62%
Info:                  SB_IO:     5/  256     1%
Info:                  SB_GB:     8/    8   100%
Info:           ICESTORM_PLL:     1/    2    50%
Info:            SB_WARMBOOT:     0/    1     0%

Max frequency for clock 'pll_clk': 35.17 MHz
```

---

So, the main conclusion is that the updated default dmem does still not work with GHDL and Yosys. A custom implementation is required. Unlike Fomu, the instantiation of SPRAM is not possible here.

`setups/osflow/devices/ice40/neorv32_dmem.alt.vhd` produces an acceptable synthesis result, but the bitstream does not work. It did not work for @juanmard, at least. It would be good to have any other TinyFPGA-BX owner try/test it.

stnolting:
> Some results:
> It seems there was some optimisation these days (from 59% to 58%).

Awesome! 58% logic utilization looks good! still room for custom logic 👍

> So, the main conclusion is that the updated default dmem does still not work with GHDL and Yosys. A custom implementation is required. Unlike Fomu, the instantiation of SPRAM is not possible here.

Damn! 
I still can't believe that this is caused by the byte-enable stuff. 🤔
I checked your alternate memory version again and I think there are - besides the byte-enable description - two main differences:
* the current DMEM uses `signal`s for the actual memory; the _alt_ version uses a `variable`
* the current DMEM uses an explicit "_either_ read or write" concept while the _alt_ version uses a "read during write" concept

Do you think any of these might be relevant?!
...maybe I should really drop this idea of an "universal generic memory"... 😅

stnolting:
> setups/osflow/devices/ice40/neorv32_dmem.alt.vhd produces acceptable synthesis result, but the bitstream does not work. It did not work for @juanmard, at least. It would be good to have any other TinyFPGA-BX owner try/test it.

A friend of mine is using a TinyFPGA (but I am not sure which one.. there are version witch mach FPGAs, right?). I will try to borrow one so I can be of better help :wink:

stnolting:
I just tested the default DMEM + IMEM again (each 4kB) for the FOMU minimal boot setup. This is what I get:
```
Info: Device utilisation:
Info: 	         ICESTORM_LC:  4588/ 5280    86%
Info: 	        ICESTORM_RAM:    20/   30    66%
```

The `ICESTORM_LC` result makes sense:
* 8x512 bytes for IMEM
* 8*512 bytes for DMEM
* 4*512 bytes for the CPU register file

Am I missing here something here?!

umarcor:
> Awesome! 58% logic utilization looks good! still room for custom logic 👍

It does! 

> I still can't believe that this is caused by the byte-enable stuff. 🤔

According to https://github.com/stnolting/neorv32/pull/56#issuecomment-861664885, there is no problem with the default imem ROM. So, the problem is the RAM, either dmem or imem.

> I checked your alternate memory version again and I think there are - besides the byte-enable description - two main differences:
> 
> * the current DMEM uses `signal`s for the actual memory; the _alt_ version uses a `variable`
> * the current DMEM uses an explicit "_either_ read or write" concept while the _alt_ version uses a "read during write" concept

I don't think that those are relevant. You can try changing the read during write, and I'd expect a similar result.

> ...maybe I should really drop this idea of an "universal generic memory"... 😅

I do think so... but even doing so, we need to find the implementation that works for TinyFPGA-BX, since the alt description passes synthesis but seems not to work.

I think the best solution will be a parameterized description which uses for generate in order to instantiate SB_RAM512X8. That is, equivalent to the SPRAM sources. However, when using SPRAM a single instance is used, always. In this case, the number of instances needs to be computed. I believe it's safe to assume that IMEM and DMEM sizes need to be multiples of 512*4=2KB, for now. That ensures that the division is clean, and the routing of enable bits is easy.

> A friend of mine is using a TinyFPGA (but I am not sure which one.. there are version witch mach FPGAs, right?). I will try to borrow one so I can be of better help 😉

That would be nice. I think we might have similar issues with other LP or HX devices.

> I just tested the default DMEM + IMEM again (each 4kB) for the FOMU minimal boot setup. This is what I get:

That does not look coherent with https://github.com/stnolting/neorv32/pull/56#issuecomment-861664885. I'm running some more tests...

> in terms of logic cells is pretty much 64k. Seems like your design tries to implement 64kB of memory somewhere. Have you adapted the default 64kB of MEM_INT_IMEM_SIZE and MEM_INT_DMEM_SIZE when not using SPRAM for the according memory?

In the TinyFPGA-BX MinimalBoot example, sizes of IMEM and DMEM are defined in the BoarTop. There is no SPRAM at all, so it does not make sense to have 64kB. Moreover, by reducing the size of DMEM, the number of LCs is reduced, as commented in https://github.com/stnolting/neorv32/pull/56#issuecomment-861664885 (*MinimalBoot (with default dmem)*).

stnolting:
> I think the best solution will be a parameterized description which uses for generate in order to instantiate SB_RAM512X8. That is, equivalent to the SPRAM sources. However, when using SPRAM a single instance is used, always. In this case, the number of instances needs to be computed. I believe it's safe to assume that IMEM and DMEM sizes need to be multiples of 512*4=2KB, for now. That ensures that the division is clean, and the routing of enable bits is easy.

Good point. If this is really a memory _design_ issue then this sounds reasonable.
Btw, there are always _two_ instances of SPRAM since these _only_ provide 16-bit width. In contrast the EBRs provide several power-of two widths. Note: A single SPRAM instance would also be possible. But then we would need some arbiter to two 2 consecutive 16-bit accesses to the memory for every 32-bit access from the CPU.

umarcor:
> Good point. If this is really a memory _design_ issue then this sounds reasonable.

Eventually, this is the secure solution. See https://github.com/ghdl/ghdl/issues/1780#issuecomment-860079143.

> Btw, there are always _two_ instances of SPRAM since these _only_ provide 16-bit width. In contrast the EBRs provide several power-of two widths. Note: A single SPRAM instance would also be possible. But then we would need some arbiter to two 2 consecutive 16-bit accesses to the memory for every 32-bit access from the CPU.

You are correct. I was focused on the fact that the size is fixed, even if an smaller DMEM_SIZE is used.

stnolting:
> Eventually, this is the secure solution. See ghdl/ghdl#1780 (comment).

I agree that synthesis might remove pure behavioral uninitialized read-only memory and write-only memories.
Maybe it's a thing with the "sub-word" access there. Even Intel Quartus, which provides good VHDL support, cannot identify this construct.

PR #69, opened 2021-06-11
VUnit checking

umarcor:
In this PR, the Processor workflow is split in two jobs. The main reason is running the VUnit tests and the shell script tests in parallel. Moreover, the installation of the RISC-V toolchain in the VUnit job might be enhanced in the future by using a container.

Other than that, as done with the main testbench, `uart_rx` is copied to `uart_rx.simple.vhd`, and it is reverted to the version before self-checking was in place. Other sources/references are updated accordingly. In upcoming PRs, `uart_rx` will be modified for using VUnit features, while `uart_rx_simple` will be used for the simple testbench.

Last, one of the enhancements from #48 was cherry-picked here. That uses VUnit's features in uart_rx.

/cc @LarsAsplund

### DIFF
--- a/.github/workflows/Processor.yml
+++ b/.github/workflows/Processor.yml
@@ -23,6 +23,7 @@ jobs:
 
   Processor:
     runs-on: ubuntu-latest
+    name: '🐧 Ubuntu'
 
     steps:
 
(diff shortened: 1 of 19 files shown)


stnolting:
Awesome! You even updated the RISC-V architecture test! :star_struck:👍

umarcor:
In fact... it was done by @LarsAsplund 😆 That was a leftover from #47...

PR #71, opened 2021-06-11
Add MSYS2 jobs to continuous integration workflows

umarcor:
In this PR, MSYS2 jobs are added to the Implementation and Processor workflows. In both cases, tools are installed on MINGW32 and MINGW64. In the Processor workflow, the VUnit testbench is executed. In the Implementation workflow, examples are synthesised.

I believe these jobs are useful for didactic purposes, since they show how Windows users can install and use the tools, for either simulation or synthesis. In fact, this is how I use NEORV32. Moreover, these jobs are are also useful for detecting regressions issues on Windows. However, they can be considered redundant, since the tasks are the same as the ones executed on containers.
Therefore, I think it might be interesting to either:

- reduce the number of examples that are tested on MSYS2.
- or, create an specific workflow for MSYS2, including both implementation and processor jobs, which is executed periodiocally (maybe once a week?).

@stnolting, what do you think?

*edit*

See execution: https://github.com/umarcor/neorv32/actions/runs/929208199 and https://github.com/umarcor/neorv32/actions/runs/929208200.


stnolting:
That's really great! Especially for Windows fanboys like me 😅 

> * or, create an specific workflow for MSYS2, including both implementation and processor jobs, which is executed periodiocally (maybe once a week?).

Even though I like having "all in one place" I think you are right regarding a new/specific workflow. 🤔

umarcor:
Done. The Windows workflow will be executed one a week, manually (through workflow_dispatch) or when new commits are pushed to master.

stnolting:
Wow, the matrix job is really great! :+1:

stnolting:
Maybe we should use a specialized trigger or even just a manual trigger. I don't think it is required to run this on _every_ new commit. The "Implementation" workflow seems sufficient to test if rtl modifications do still run.
What do you think?

umarcor:
Yeah. Up to you. The workflow dispatch and the weekly cron job  are enough. I added the push event so that you could see it run and decide what to do.

PR #75, opened 2021-06-16
iCESugar

umarcor:
This PR adds a MinimalBoot example for [iCESugar](https://github.com/wuxx/icesugar). That is a board with an UP5K FPGA, which unlike Fomu, has plenty of I/O 🤣. So, adding the MinimalBoot example using the internal oscillator was easy enough :smile:.

Adding the Minimal one with an external clock will be more challenging, so let's leave it for an upcoming PR...

Figures from https://twitter.com/juanmard/status/1404971683170197509:

/cc @juanmard


stnolting:
This is so cool!! :smile:

Thx @juanmard and @umarcor :+1:

PR #76, opened 2021-06-16
[setups/examples] add iCESugar Minimal

umarcor:
Name Minimal may not be the most descriptive. The purpose of this PR is to modify the MinimalBoot example of the iCESugar for using the "external" clock from pin 35, instead of the internal oscillator.

On the Fomu, we could use SB_PLL40_CORE for instantiating the PLL, regardless of the clk input being internal or external. That is because of the bank where the clk input is located (`IOB_11b_G5	DPIO/GBIN	1	COMP_of_IOB_10a	F4`). On the iCESugar, that is `IOT_46b_G0	DPIO/GBIN	0		B3	35`. As explained in https://github.com/wuxx/icesugar/blob/master/doc/LatticeSemi/SBTICETechnologyLibrary201708.pdf, SB_PLL40_PAD needs to be used.

Therefore, SB_PLL40_PAD was added to `setups/osflow/devices/ice40/sb_ice40_components.vhd`. Compared to iCESugar MinimalBoot, this one (iCESugar Minimal) instantiates SB_PLL40_PAD instead of SB_PLL40_CORE, and removes the internal oscillator instantiation. All extensions are disabled except Zicsr. However, the UART is preserved. That is, the ProcessorTop_MinimalBoot template is used.

Moreover, pin 35 of the UP5K is connected to a pin of the ARM on the board, through jumper J1. See https://raw.githubusercontent.com/wuxx/icesugar/master/schematic/iCESugar-v1.5.pdf. We are not sure whether the purpose is for the ARM to send a clock to the FPGA, or for the FPGA to send a clock to the ARM. The source of the firmware on the ARM seems not to be available: https://github.com/wuxx/icesugar/tree/master/firmware. Hence, @juanmard asked about the purpose of this connection in wuxx/icesugar#29. However, there is a comment in the readme: https://github.com/wuxx/icesugar/blob/master/README_en.md#icelink

> *4. the MCO can provide 12Mhz clock for FPGA as extern clock.*

Therefore, we adjusted the PLL for `icepll -i 12 -o 22`, and it works!

/cc @juanmard

### DIFF
--- a/.github/generate-job-matrix.py
+++ b/.github/generate-job-matrix.py
@@ -23,6 +23,10 @@ print('::set-output name=matrix::' + str([
   'bitstream': 'neorv32_Fomu_pvt_UP5KDemo.bit'
 }, {
   'board': 'iCESugar',
+  'design': 'Minimal',
+  'bitstream': 'neorv32_iCESugar_Minimal.bit'
+}, {
+  'board': 'iCESugar',
   'design': 'MinimalBoot',
   'bitstream': 'neorv32_iCESugar_MinimalBoot.bit'
 }]))
--- a/setups/examples/Makefile
+++ b/setups/examples/Makefile
@@ -44,7 +44,7 @@ UPduino_v3:
 Minimal:
 	$(MAKE) \
 	  DESIGN=$@ \
-	  DESIGN_SRC=$(TEMPLATES)/processor/neorv32_ProcessorTop_Minimal.vhd \
+	  DESIGN_SRC=$(TEMPLATES)/processor/neorv32_ProcessorTop_Minimal*.vhd \
 	  $(BOARD)
 
 MinimalBoot:
--- a/setups/examples/neorv32_iCESugar_BoardTop_Minimal.vhd
+++ b/setups/examples/neorv32_iCESugar_BoardTop_Minimal.vhd
@@ -1,0 +1,172 @@
+-- #################################################################################################
+-- # << NEORV32 - Example setup with an external clock, for the iCESugar (c) Board >>              #
+-- # ********************************************************************************************* #
+-- # BSD 3-Clause License                                                                          #
+-- #                                                                                               #
+-- # Copyright (c) 2021, Stephan Nolting. All rights reserved.                                     #
+-- #                                                                                               #
+-- # Redistribution and use in source and binary forms, with or without modification, are          #
+-- # permitted provided that the following conditions are met:                                     #
+-- #                                                                                               #
+-- # 1. Redistributions of source code must retain the above copyright notice, this list of        #
+-- #    conditions and the following disclaimer.                                                   #
+-- #                                                                                               #
+-- # 2. Redistributions in binary form must reproduce the above copyright notice, this list of     #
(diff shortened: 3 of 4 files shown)


stnolting:
> Therefore, SB_PLL40_PAD was added to setups/osflow/devices/ice40/sb_ice40_components.vhd. Compared to iCESugar MinimalBoot, this one (iCESugar Minimal) instantiates SB_PLL40_PAD instead of SB_PLL40_CORE, and removes the internal oscillator instantiation. All extensions are disabled except Zicsr. However, the UART is preserved. That is, the ProcessorTop_MinimalBoot template is used.

So it just takes a global clock network as input??! 🤔 

> Therefore, we adjusted the PLL for icepll -i 12 -o 22, and it works!

Awesome! 🚀 👍 

----

By the way...
All those `<RTE> Illegal instruction ...` errors are caused by the executable because it tries to use features that were not implemented due to the generics configuration. I think these are the "hardware performance monitors" (configured via `HPM_NUM_CNTS`) that are used by CoreMark to count different performance metrics.

umarcor:
> So it just takes a global clock network as input??! 🤔

Yes. It uses an external pin as the clock input. However, that is not "new" in this repo. The same applies to the Fomu Minimal (external clock) and Fomu MinimalBoot (internal oscillator) examples. Moreover, I think it always tries to use a global network for the signals that go into the PLL. I think the explanation is deeper in the architecture of the FPGA, and it's related to the partitioning of power and pin banks.

My understanding is that pins in bank 1 are somehow isolated from 0, 2 and the internal FPGA fabric. Therefore, those are internally connected to some matrix/whatever and then go to the fabric. Conversely, banks 0,2 need an explicit pad instantiation because there is no such intermediate whatever. However, `SBTICETechnologyLibrary201708.pdf` says the PAD should be used no only with bank 2 but also with bank 0, while CORE mentions bank 2 only (not 1 explicitly). That is where I got lost.

> By the way...
> All those `<RTE> Illegal instruction ...` errors are caused by the executable because it tries to use features that were not implemented due to the generics configuration. I think these are the "hardware performance monitors" (configured via `HPM_NUM_CNTS`) that are used by CoreMark to count different performance metrics.

Yeah. I'm honestly not paying too much care to the software output, as long as it works 😆 

I hope @juanmard or someone else can have another take at these examples and improve them in further PRs: increasing the clock, exposing more I/O, testing multiple software applications...

stnolting:
> My understanding is that pins in bank 1 are somehow isolated from 0, 2 and the internal FPGA fabric. Therefore, those are internally connected to some matrix/whatever and then go to the fabric. Conversely, banks 0,2 need an explicit pad instantiation because there is no such intermediate whatever. However, SBTICETechnologyLibrary201708.pdf says the PAD should be used no only with bank 2 but also with bank 0, while CORE mentions bank 2 only (not 1 explicitly). That is where I got lost.

Sounds "reasonable..." 😅 

> I hope @juanmard or someone else can have another take at these examples and improve them in further PRs: increasing the clock, exposing more I/O, testing multiple software applications...

Would be great! 😉

PR #80, opened 2021-06-16
Added option to print a selected subset of information from processor…

LarsAsplund:
…_check to UART0 during simulation.

### DIFF
--- a/.github/workflows/Processor.yml
+++ b/.github/workflows/Processor.yml
@@ -82,6 +82,7 @@ jobs:
           clean_all \
           USER_FLAGS+=-DRUN_CHECK \
           USER_FLAGS+=-DUART0_SIM_MODE \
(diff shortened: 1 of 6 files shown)


stnolting:
Looks great! :+1:

However, I would like to keep the memory requirements below 16kB - there is some stuff in the test program that is not needed anymore. I think it would be ok to make this a simulation-only test program so we can get rid of some is-simulation checks making the executable a little less smaller. I would do the changes if that is ok for you.

LarsAsplund:
@stnolting That's ok with me.

stnolting:
I made some small edits here. The most importants are:
* IMEM uses the default size of 16kB for simulation
* the UART tests will output 0x00 at the _default_ baudrate (19200); so the testbench receivers will detect 0x00 twice for each UART

stnolting:
Seems to work :wink:
So I think this can be merged if you are ok with the changes @LarsAsplund

PR #49, opened 2021-05-29
Make the UART test code a VUnit verification components

LarsAsplund:
This PR builds on #48. Each commit is one step in the transformation. Have a look at the commit comments for explanation and rationale.

PR #82, opened 2021-06-18
[sim] use custom VUnit loggers for better verbosity control 

umarcor:
Close #48

This PR superseeds #48, but it needs to be merged after #80.

That is, @LarsAsplund's PR2 branch can be analysed/understood in three parts:

1. PR1 was already split and merged in previours PRs.
2. The software modifications for controlling UART verbosity are being handled in #80.
3. This PR adds VUnit loggers to the hardware testbench plumbing.

### DIFF
--- a/sim/neorv32_tb.vhd
+++ b/sim/neorv32_tb.vhd
@@ -148,14 +148,17 @@ architecture neorv32_tb_rtl of neorv32_tb is
   end record;
   signal ext_mem_a, ext_mem_b, ext_mem_c : ext_mem_t;
 
+  constant uart0_rx_logger : logger_t := get_logger("UART0.RX");
+
 begin
   test_runner : process
   begin
     test_runner_setup(runner, runner_cfg);
+    -- Show passing checks for UART0 on the display (stdout)
+    show(uart0_rx_logger, display_handler, pass);
     wait for 35 ms; -- Just wait for all UART output to be produced
     test_runner_cleanup(runner);
   end process;
-
 
   -- Clock/Reset Generator ------------------------------------------------------------------
   -- -------------------------------------------------------------------------------------------
@@ -318,7 +321,7 @@ begin
   begin
     uart0_checker: entity work.uart_rx
       generic map (
-        name => "uart0",
+        logger => uart0_rx_logger,
         expected => uart0_expectation,
         uart_baud_val_c => uart0_baud_val_c)
       port map (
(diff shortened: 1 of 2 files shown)


stnolting:
I'm looking forward to this as I am very interested in this logging feature! 👍

sim/neorv32_tb.vhd
```diff
@@ -148,15 +148,18 @@ architecture neorv32_tb_rtl of neorv32_tb is
...
 
+  constant uart0_rx_logger : logger_t := get_logger("UART0.RX");
+
 begin
   test_runner : process
   begin
     test_runner_setup(runner, runner_cfg);
-    wait for 20 ms; -- Just wait for all UART output to be produced
+    -- Show passing checks for UART0 on the display (stdout)
+    show(uart0_rx_logger, display_handler, pass);
```
LarsAsplund, line 158: Now that we will have output on both UARTs we should make them both visible

umarcor: Done.

umarcor:
I rebased on top of master. I think this is now ready to be merged.

LarsAsplund:
Hm, it's not due to this PR but probably happened when I was rebasing #80. It should be output on the UART like [this](https://github.com/LarsAsplund/neorv32/runs/2843220004?check_suite_focus=true#step:7:560) but now it's not, the text is just "printed": https://github.com/stnolting/neorv32/pull/82/checks#step:7:560

The bad thing with my testbench is that there is no proper end of test mechanism. It just runs for a while and if all UART output meet the expectation than everything is good. There is nothing checking if all output has been produced, or any output at all. The good thing is that this problem is removed in my "PR3" that needs to be rebased. I will try to do some work on that this evening. Should probably try to fix it on main first.

stnolting:
> The bad thing with my testbench is that there is no proper end of test mechanism. It just runs for a while and if all UART output meet the expectation than everything is good.

I know we already had some discussions here, but how about a simple mechanism **in the testbench** (like a specific GPIO output pattern) to indicate the processor has finishe executing the test program? It is up to the testbench to evaluate (all previous UART outputs for example) if the actual test succeeded.

umarcor:
IMHO, we can either have that fixed in #49, or afterwards. As soon as we start using Verification Components for the UART, I would expect to use them for the external wishbone interface too (thus reducing that maintenance burden from the simple tb).  Then, it will make sense to give a better thought at how to coordinate the VCs for deciding a success or failure.

PR #83, opened 2021-06-20
[setups/osflow] support optionally using Verilog sources and add Fomu MixedLanguage example

umarcor:
The purpose of this PR is to provide an example about how to do mixed language synthesis with GHDL and Yosys, including NEORV32. Example Fomu MixedLanguage is based on Fomu MinimalBoot and should be equivalent. The only difference is that SB_HFOSS and SB_PLL40_CORE are instantiated in a Verilog module, which is used in the BoardTop file.

Unfortunately, I'm not experienced enough with Yosys and Verilog, and there's something I'm not doing properly. The instantiation of SB_PLL40_CORE from Verilog works, but it complains about SB_HFOSS. I thought that adding blackbox components was not required when using Verilog. However, I had to add `sb_ice40_components.v`. Still, the SB_PLL40_CORE works only:

```
 ERROR: cell type 'SB_HFOSS' is unsupported (instantiated as 'clk_inst.HSOSC_inst')
```

I'm keeping this as a draft until we guess what's going on.


umarcor:
@rodrigomelo9, after this PR is done, it would be interesting to try adding pyFPGA/symbiflow_cli and/or pyFPGA/openflow support to this repo. See `setups/osflow/filesets.mk` and `setups/osflow/tools.mk`.

umarcor:
Thanks to @juanmard, this is now ready to merge.

I had written `SB_HFOSS` instead of `SB_HFOSC` 😖

rodrigomelo9:
Hi @umarcor I think that PyFPGA is a perfect match here because the project provides a vendor-independent implementation. I need to update PyFPGA to use symbiflow_cli under the hood and then, I will work on that.

stnolting:
> The purpose of this PR is to provide an example about how to do mixed language synthesis with GHDL and Yosys, including NEORV32

So cool! Thanks for implementing this! :+1:

PR #90, opened 2021-06-29
Make uart_rx a VUnit verification component. Step 4.

LarsAsplund:
Collect all generics into a single handle generic. A single generic is easier to pass
around and makes it easier to update the VC without changing the user code.
Also using a `check_uart` procedure to raise the abstraction of the API.


stnolting:
This is so much tidier! 😄 
Great work! :+1:

PR #93, opened 2021-06-30
[setups/vivado] arty-a7-test-setup: define multiple filesets, set board, set language

umarcor:
In this PR, the TCL script for creation of the Vivado project for Arty boards is updated.
Instead of adding files to libraries directly, filesets are defined. This is for making it easier to understand the similarities with `setups/osflow/filesets.mk`.
BTW, the board and language are explicitly defined in the project.

### DIFF
--- a/setups/vivado/arty-a7-test-setup/create_project.tcl
+++ b/setups/vivado/arty-a7-test-setup/create_project.tcl
@@ -1,6 +1,6 @@
-set board "A7-35"
+set board "arty-a7-35"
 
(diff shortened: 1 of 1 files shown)


stnolting:
Good idea! I was not aware of this TCL fileset option :+1:

umarcor:
> Good idea! I was not aware of this TCL fileset option 👍

In fact `fileset_*` are variables, arrays of strings/paths :wink:

PR #98, opened 2021-07-04
OrangeCrab

umarcor:
This is work in progress for supporting ECP5 devices, and specifically the OrangeCrab board.

@jeremyherbert, can you please have a look? Currently, the implementation is successful and a bitstream is generated in CI. However, please, do NOT try that bitstream yet. First, we need to ensure that the design is correct for your board:

- The constraints file is based on https://github.com/gregdavill/OrangeCrab-examples/tree/main/verilog.
- The nextpnr flags are based on https://github.com/antonblanchard/ghdl-yosys-blink/blob/master/Makefile.
- The OrangeCrab MinimalBoot tries to use CLK, RST_N, LED_RGB, GPIO (for UART) and USB (constants) ports.
- The CLK is connected to NEORV32 directly, without a PLL, without any adjustment. Therefore, I'd expect it to fail.
- The default memory descriptions are used, which are based on inference. That might work or it might fail.

Jeremy, can you review `neorv32_OrangeCrab_BoardTop_MinimalBoot.vhd` and enhance it so that it makes sense (the UART works)? The command for generating the bitstream is `make -C setups/examples BOARD=OrangeCrab MinimalBoot`.

If you instantiate hard IP (such as a PLL), we need to add `setups/osflow/devices/sb_ecp5_components.*` sources. That might need overriding `DEVICE_SRC`, when calling `setups/OrangeCrab/Makefile` from `setups/examples/Makefile`. Please, let me know if you need help with that.

PR #99, opened 2021-07-04
osflow rework

umarcor:
I extracted the commits from #98 which add ECP5 support but which are not specific about OrangeCrab. Then, I reworked the structure in `osflow` for reducing the board specific subdirectories, which might be misleading.

- There is a `setups/osflow/constraints` directory containing the PCF/LPF files for all boards.
- `setups/osflow/BOARDNAME/Makefile` were renamed to `setups/osflow/boards/BOARDNAME.mk`. Those `BOARDNAME.mk` now only contain the tasks which are specific to a board (such as a custom bitstream load command).
- I added `setups/osflow/boards/index.mk` and `common.mk`, which contain all the targets common to any board.

As a result, now there is no `Makefile` in `setups/osflow/**`, making it explicit that those are all resources to be used from some other entrypoint (precisely, `setups/examples/Makefile`). See https://github.com/umarcor/neorv32/tree/osflow-rework/setups/osflow


stnolting (approved):

stnolting:
Wow! Thank you so much for all that work and effort! :+1:

PR #100, opened 2021-07-06
[docs] move Makefile from project root to subdir 'docs'

umarcor:
Initially, we put the Makefile for building the docs in the root of the repo, because we expected to add entrypoints for other tasks in the same file. However, given how the rest of the repo evolved, that is probably not going to happen. We will probably have some orchestration/entrypoint in the root which wraps the different makefiles. Therefore, this PR moves the Makefile for building the documentation into subdir `docs`.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -47,7 +47,7 @@ jobs:
         fetch-depth: 0
 
     - name: '📚 Build Datasheet and User Guide (PDF and HTML)'
-      run: make container
+      run: make -C docs container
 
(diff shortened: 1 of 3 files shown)


stnolting:
I like that! :+1:
Make the root directory a little bit cleaner 😅

stnolting:
I have updated the user guide accordingly.
Everything should be sync now, or did we forget something?

umarcor:
I cannot think of any missing piece at the moment. Anyway, we can fix it afterwards in case we forgot anything. Hence, shall we merge?

PR #101, opened 2021-07-07
[setups] reorganise HDL sources

umarcor:
This PR splits the entities and architectures of IMEM and DMEM to separated files, so that the entity is not repeated in all the different variants. That allows all the designs to include the same entity files, and provide custom architectures depending on the target board/design.

This was discussed in #56.


stnolting:
:+1:

You know this is a "complicated" concept for me 😅 
I like the idea of having all core sources in one folder (currently, in `rtl/core`) and that these files are written in a platform independent style. However, I see the problems when optimizing for a certain platform you need to ignore certain files from the core folder (like IMEM and DMEM) and replace this with a platform-specific one (like the SPRAM-based IMEM and DMEM)...

(It would be cool to have an option in VHDL to "override" architecture definitions like [GCC's `weak` attribute](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html): so for example always include the default IMEM (which is declared "weak") from `rtl/core`. If we add another entity/architecture with the same name but without "weak" the synthesis/simulation tool takes that one instead - but maybe this is a Stephan-only use case 😄)

Anyway.
I like your concept. But I have some remarks:
* Shouldn't setup-specific (-> platform-specific) files be in the according `setups/...` folder?
* Why do we need an entity-only src-file? The entity itself is fixed in the core package file.

umarcor:
> You know this is a "complicated" concept for me 😅

I do know, I do understand, and I do partially agree. So, we don't need to resolve this now/yet. I created the PR because I wanted to close  #56 and preserve these commits. However, I will soon open a new PR which will conflict with this one (hopefully for good reason). We should handle that one first :wink:.

> I like the idea of having all core sources in one folder (currently, in `rtl/core`) and that these files are written in a platform independent style. However, I see the problems when optimizing for a certain platform you need to ignore certain files from the core folder (like IMEM and DMEM) and replace this with a platform-specific one (like the SPRAM-based IMEM and DMEM)...

To be strict, this is not a problem with the "typical" workflows supported by most vendor tools and even open source project management solutions. In those cases, you are forced to explicitly add each individual file and even to specify the analysis/elaboration order. However, since it is 2021 already, some of us (hardware developers) believe it's time to start using wildcards as normal people (say software developers) do 😆. That's what we do in VUnit, and that's what we want to extend to synthesis workflows. In fact, GHDL does support wildcards for both simulation and synthesis. Moreover, both GHDL and VUnit allow retrieving the list of sources in compilation order. Therefore, we just need not to be limited by the orchestration plumbing.

Currently, we have an specific list in https://github.com/stnolting/neorv32/blob/master/setups/osflow/filesets.mk#L12-L44, which might sound incoherent with my previous statements. However, we already discussed that when `filesets.mk` was created. In fact, you suggested using a wildcard and I replied it'd better keep it as a list, until we had this discussion. I knew it was not going to be an easy/fast discussion, tho. Hence, here we are now 😃 

> (It would be cool to have an option in VHDL to "override" architecture definitions like [GCC's `weak` attribute](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html): so for example always include the default IMEM (which is declared "weak") from `rtl/core`. If we add another entity/architecture with the same name but without "weak" the synthesis/simulation tool takes that one instead - but maybe this is a Stephan-only use case 😄)

It's not an Stephan-only use case. My opinion is that's a severe flaw/problem/missing feature in the language. In fact, I discussed this specific use case with both @JimLewis and @Paebbels in separated video meetings during the last month. 
Actually, I thought I had written about it already. I certainly started doing it somewhere, but I cannot find it now. I moved so many issues from TWIKi to [gitlab.com/IEEE-P1076/VHDL-Issues](https://gitlab.com/IEEE-P1076/VHDL-Issues/-/issues) and now it's difficult to find my own... Anyway, I created [IEEE-P1076/VHDL-Issues#235](https://gitlab.com/IEEE-P1076/VHDL-Issues/-/issues/235) just now. Please, go ahead an leave your reaction/thoughts there :wink:.

> * Shouldn't setup-specific (-> platform-specific) files be in the according `setups/...` folder?

Let me be socratic: which sources are setup-specific/platform-specific? which subdirs are so?
Subdirs `osflow`, `quartus`, `radiant` and `vivado` are tool-specific. That is orthogonal to the setup/platform (i.e. the board/device/design). We can have a single setup/platform (e.g. UPduino v3) and implement it using 4-5 different toolchains (radiant, osflow, edalize, litex...). Therefore, some of the sources in `setus/rtl/mem` might be reused by multiple tools. Moreover, we are mostly providing alternative but theoretically equivalent descriptions, so it is sensible to keep them close to each other for easier understanding of the whole picture.

Moreover, I would propose to move `setups/osflow/constraints` to `setups/osflow`, because constraints can also be potentially reused regardless of the toolchain.

Nonetheless, I agree that it might be confusing to have `rtl/core/mem` and `setups/rtl/mem`. Currently, the root of `setups/` is empty, except for the README. So, we might `mv setups/rtl/core/mem/*.vhd setups/`?

> * Why do we need an entity-only src-file? The entity itself is fixed in the core package file.

Is it possible to instantiate a component by providing the architecture only but not the corresponding entity?

umarcor:
> However, I will soon open a new PR which will conflict with this one (hopefully for good reason). We should handle that one first 😉.

@stnolting, see #110 and precisely https://github.com/stnolting/neorv32/pull/110/files#diff-d4927f87ba741c63622c0db399b1f2f85c0ba1ed83843a419a1e66b6255f152dR53-R78. I that context, if we had all the alternative architectures in a single directory using the naming `neorv32_?mem.VARIANT.vhd`, we might create the `Filesets.InstructionMemory` and `Filesets.DataMemory` automatically by parsing the filenames. Then, users would only need to take care about the `GetMemorySources`, where we actually define which of the variants is to be used dependending on the target board and design. As you see, the variant is selected by name.

PR #107, opened 2021-07-08
[setups/osflow] update OrangeCrab constraints file

### LINKED ISSUE #106: Constraints file for OrangeCrab r0.2
Coming from #98.

The naming of the GPIO ports in the OrangeCrab constraints file does currently not match the naming in the silkscreen. That is confusing. For instance, `*_GPIO_4` is used in the MinimalBoot example, which corresponds to pin labeled 9 on the board. This issue is for keeping track of the updates that might be required after gregdavill/OrangeCrab-examples#20 is resolved.

> https://github.com/stnolting/neorv32/pull/98#issuecomment-875487952
>
> with respect to the constraints file, I would suggest adding a note to the readme or similar about this. The main issue is that, for example, the silkscreen pin numbers do not have a pin 4 anywhere; the only numbered GPIO pins on the silkscreen are 0, 1, 5, 6, 9, 10, 11, 12, 13; the rest have names like SCL, SDA or are muxed into an external ADC. I personally would suggest deleting the lines that were commented out in my previous comment, and just using the ones that were not commented as these match the silkscreen. But I am nonetheless happy whichever way makes the most sense to everyone.

> https://github.com/stnolting/neorv32/pull/98#issuecomment-874403728
>
> Note: The constraints file for the OrangeCrab doesn't match the silkscreen for some pins (pin 9 is one of them). The correct constraints are in this pinout picture: https://gregdavill.github.io/OrangeCrab/r0.2/docs/pinout/
>
> I filed a bug for that here: gregdavill/OrangeCrab-examples#20

> https://github.com/stnolting/neorv32/pull/98#issuecomment-875669193
>
> It is particularly confusing to have GPIO as an array of 13 bits, if ~8 of them have other purposes. My proposal is:
> 
> 1. Have it clarified/fixed in [PCF files do not match the actual pinout for GPIOs gregdavill/OrangeCrab-examples#20](https://github.com/gregdavill/OrangeCrab-examples/issues/20).
> 2. Update https://github.com/gregdavill/OrangeCrab-examples/blob/main/verilog/orangecrab_r0.2.pcf and https://github.com/hdl/constraints/blob/main/board/OrangeCrab/constraints.lpf.
> 3. Update here.
> 
> In fact, my expectation is to submodule hdl/constraints at some point. That's why I'm picking the constraints files from there, and maintaining the naming scheme. Overall, maintaining constraint files and board metadata should be out of scope of this repository. We do need those resources, but the effort devoted to it in the context of NEORV32 should be minimal.

/cc @jeremyherbert

umarcor:
Close #106.

The constraints file was updated in gregdavill/OrangeCrab-examples#20 (https://github.com/gregdavill/OrangeCrab-examples/commit/6dceec16f116aab02ed8c6b43dc990d2b4404f63) and [hdl/constraints](https://github.com/hdl/constraints) (https://github.com/hdl/constraints/commit/2bf87ec96cac79ee20d375487d71cc654aa8c269). This PR updates the constraints file in this repo and the BoardTop of the MinimalBoot example accordingly.

The example does not change, the same pin is used. However, now the name of the top-level port does match the silkscreen.

FTR, the solution is almost the same as suggested by @jeremyherbert, with the addition of A0-A3.


stnolting:
That was fast! :+1:

PR #109, opened 2021-07-08
riscv-arch-test script cleanup

umarcor:
This is some cleanup of the `riscv-arch-test/run_riscv_arch_test.sh` script. It should be functionally equivalent, but easier to read/maintain.

Moreover, it seems that subdir `riscv-arch-test` is used just for simulation with GHDL. @stnolting, what do you think about moving it into `sim/riscv-arch-test`? That would keep the root of the project cleaner, while making it explicit that the riscv-arch-test are used for testing/verification. In the future, we might enhance `run_riscv_arch_test.sh` for using some VUnit features, or have it removed and handle these tests in the existing `sim/run.py` script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ sw/image_gen/image_gen.exe
 
 # example bitstreams
 /setups/examples/*.bit
+/setups/examples/*.svf
--- a/riscv-arch-test/run_riscv_arch_test.sh
+++ b/riscv-arch-test/run_riscv_arch_test.sh
@@ -1,82 +1,64 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Abort if any command returns != 0
 set -e
 
-# Project home folder
-homedir="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
-homedir=$homedir/..
+cd $(dirname "$0")
 
-# Check GCC toolchain installation
-echo "--------------------------------------------------------------------------"
-echo "> Checking RISC-V GCC toolchain..."
-echo "--------------------------------------------------------------------------"
+rm -rf work/neorv32
+mkdir -p work/neorv32
+
+cd ..
+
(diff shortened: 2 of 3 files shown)


stnolting:
> This is some cleanup of the riscv-arch-test/run_riscv_arch_test.sh script. It should be functionally equivalent, but easier to read/maintain.

Look good! :+1:
According to https://github.com/umarcor/neorv32/actions/workflows/riscv-arch-test.yml the modified script works correctly, right?

> Moreover, it seems that subdir riscv-arch-test is used just for simulation with GHDL. @stnolting, what do you think about moving it into sim/riscv-arch-test? That would keep the root of the project cleaner, while making it explicit that the riscv-arch-test are used for testing/verification.

I always wanted to put `riscv-arch-test` into the `sw` folder - because it is software :smile: My plan was to support on-board execution of the tests, too, but I have not further worked on that...

However, I really like the idea of using VUnit to accelerate simulation of the tests (#35). So, yeah, I'm fine with moving the tests to the simulation folder. :+1:

umarcor:
> I always wanted to put `riscv-arch-test` into the `sw` folder - because it is software 😄 My plan was to support on-board execution of the tests, too, but I have not further worked on that...
> 
> However, I really like the idea of using VUnit to accelerate simulation of the tests (#35). So, yeah, I'm fine with moving the tests to the simulation folder. 👍

That makes sense... How about merging this as-is for now? Then we might move the software to subdir `sw` and the test script itself to subdir `sim`, which would be the most coherent with other test setups.

umarcor:
> According to https://github.com/umarcor/neorv32/actions/workflows/riscv-arch-test.yml the modified script works correctly, right?

Yeah. I run it for testing that the colours were properly shown: https://github.com/umarcor/neorv32/runs/3024526753?check_suite_focus=true#step:5:95. That's the reason for using `printf` instead of `echo`.

PR #114, opened 2021-07-10
use RISCV_PREFIX instead of RISCV_TOOLCHAIN

umarcor:
The riscv-arch-test uses variable `RISCV_PREFIX`, but the makefiles in `sw` use `RISCV_TOOLCHAIN` for the same purpose This PR makes them consistent by using `RISCV_PREFIX` only.


stnolting (approved):

stnolting:
I never thought about this, but now it is much more consistent!
Great work! :+1:

PR #116, opened 2021-07-10
[ci] add workflow 'Containers'

umarcor:
Instead of installing the dependencies in each CI workflow and job, in this PR a 'Containers' workflow is added for building custom images to be used in this repo. Precisely, images `ghcr.io/stnolting/neorv32/sim` and `ghcr.io/stnolting/neorv32/impl` are built and pushed. Then, in other workflows, those images are used.

Workflow 'Containers' is not triggered by pushes or PRs. It is executed weekly, and it can be triggered manually.

Note that we are currently using container images already. Therefore, the only advantage of this PR for now is avoiding the installation of the RISCV toolchain. However, it sets the foundation for future enhancements, such as the ones in #110.

### DIFF
--- a/.github/impl.dockerfile
+++ b/.github/impl.dockerfile
@@ -1,0 +1,3 @@
+FROM gcr.io/hdl-containers/debian/buster/impl
+
+ENV GHDL_PLUGIN_MODULE=ghdl
--- a/.github/sim.dockerfile
+++ b/.github/sim.dockerfile
@@ -1,0 +1,16 @@
+FROM gcr.io/hdl-containers/debian/buster/sim/osvb
+
+RUN apt-get update -qq \
+ && DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends \
+  g++ \
+  git \
+  make \
+  time \
+ && apt-get autoclean && apt-get clean && apt-get -y autoremove \
+ && rm -rf /var/lib/apt/lists/* \
+ && mkdir -p /opt/riscv \
+ && curl -fsSL https://github.com/stnolting/riscv-gcc-prebuilt/releases/download/rv32i-2.0.0/riscv32-unknown-elf.gcc-10.2.0.rv32i.ilp32.newlib.tar.gz | \
+ tar -xzf - -C /opt/riscv \
+ && ls -al /opt/riscv
+
+ENV PATH $PATH:/opt/riscv/bin
--- a/.github/workflows/Containers.yml
+++ b/.github/workflows/Containers.yml
@@ -1,0 +1,40 @@
+name: Containers
+
+on:
+  push:
+    paths:
+    - '.github/*.dockerfile'
+  schedule:
+    - cron: '0 0 * * 5'
+  workflow_dispatch:
+
+jobs:
+
+  Container:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+        - impl
+        - sim
+    name: '🛳️ ${{ matrix.image }}'
+
+    steps:
+
+    - name: '🧰 Repository Checkout'
+      uses: actions/checkout@v2
+
+    - name: '⛴️ Build neorv32/${{ matrix.image }}'
+      run: docker build -t ghcr.io/stnolting/neorv32/${{ matrix.image }} - < .github/${{ matrix.image }}.dockerfile
+
+    - name: '🔑 Login to ghcr.io'
+      if: github.event_name != 'pull_request'
+      uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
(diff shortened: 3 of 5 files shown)


umarcor:
Note that CI fails because images were not pushed to `ghcr.io/stnolting/neorv32` (since I don't have permissions to do that, and a PR would never be allowed to do it). However, I tested it in my fork using `ghcr.io/umarcor/neorv32`:

- https://github.com/umarcor/neorv32/commit/49502742b1bdec7cfa1d131ffe2a33b1ac139065
- https://github.com/umarcor/neorv32/actions/runs/1018868127
- https://github.com/umarcor/neorv32/actions/runs/1018868143
- https://github.com/umarcor/neorv32/actions/runs/1018868196

rtl/templates/processor/neorv32_ProcessorTop_stdlogic.vhd
```diff
@@ -431,7 +431,7 @@ begin
   slink_conv:
   for i in 0 to 7 generate
     slink_tx_dat_o(i)     <= std_logic_vector(slink_tx_dat_o_int(i));
-    slink_rx_dat_i_int(i) <= std_ulogic_vector(slink_rx_dat_i(i));
+    slink_rx_dat_i_int(i) <= std_logic_vector(slink_rx_dat_i(i));
```
stnolting, line 434: Thanks for fixing the type bug. But shouldn't this be a conversion to `std_ulogic_vector` since `slink_rx_dat_i_int(i)` is (now) of type `sdata_8x32_t` (? array of `std_ulogic_vector`)?

umarcor: Honestly, I did not look into it thoroughly. It was surprising for me that this bug showed up when working on the tools, so I just did the minimal changes for working around it. Before fixing it, I think we should understand why it is not failing in the current CI, but it failed when I tried to use this PR.

stnolting:
I like this concept! However, I actually don't line the thing with installing the GCC toolchain. There should be a better approach like using the "official" nightly releases from https://github.com/riscv/riscv-gnu-toolchain/releases. I know we have discussed this already and I know there are some multilib issues... Anyway, I just wanted to say again, that I don't like this "curl Stephan's GCC" concept :smile:

However, the container workflow is a good idea and it might be very useful for future modules (even though I am still overwhelmed by this and I honestly cannot say that I understand all of it :smile:). :+1:

> Honestly, I did not look into it thoroughly. It was surprising for me that this bug showed up when working on the tools, so I just did the minimal changes for working around it. Before fixing it, I think we should understand why it is not failing in the current CI, but it failed when I tried to use this PR.

Let's clear that before merging.

umarcor:
> I like this concept! However, I actually don't line the thing with installing the GCC toolchain. There should be a better approach like using the "official" nightly releases from https://github.com/riscv/riscv-gnu-toolchain/releases. I know we have discussed this already and I know there are some multilib issues... Anyway, I just wanted to say again, that I don't like this "curl Stephan's GCC" concept 😄

I agree, however, it is unrelated to this PR. Here, I moved the same code that is used currently into a container. I did not change the functionality/procedures at all. This is precisely meant to make modifications easier. After this PR is merged, changing it in the dockerfile will suffice for all the (non-windows) workflows to use it.

> However, the container workflow is a good idea and it might be very useful for future modules (even though I am still overwhelmed by this and I honestly cannot say that I understand all of it 😄). 👍

The main idea is: instead of installing the dependencies each time, use containers for that. We create container images once a week only (so, we download and install the dependencies once a week). All other workflow runs will just retrieve the container image and execute the tasks inside. If you review the modifications to the workflows, I just removed all the setup/installation steps, and then used `uses: docker://` for running the remaining steps inside an specific container.

> Let's clear that before merging.

See #126.

umarcor:
I rebased this PR, and I reorganised the content. There are now three commits:

- Add workflow and dockerfiles.
- Use container in workflow Implementation.
- Use container in workflow Processor.

I did *not* use the container in workflow riscv-arch-test due to the serious performance issues discussed in #119.

@stnolting, I think this is ready to merge. It will allow to rebase the pydoit branch. Meanwhile, we can investigate what's happening with the performance of those tests.

rtl/templates/processor/neorv32_ProcessorTop_stdlogic.vhd
umarcor: This was resolved in #126.

stnolting:
> @stnolting, I think this is ready to merge. It will allow to rebase the pydoit branch. Meanwhile, we can investigate what's happening with the performance of those tests.

Sorry, somehow this got lost in my feed... 😅 

However, thanks for the contribution! ❤️

PR #117, opened 2021-07-11
mv riscv-arch-test sw/sig-arch-test

umarcor:
As discussed in #109, in this PR submodule `riscv-arch-test` is moved to `sw/sig-arch-test`, and the `run_riscv_arch_test.sh` script is moved to subdir `sim`.

READMEs and gitignore files are updated accordingly.


stnolting:
Looks good!
But why "SIG-arch-test"? I don't get it :smile:

umarcor:
> But why "SIG-arch-test"? I don't get it 😄

Just for avoiding users/developers being mislead because of having the same name repeated `sw/riscv-arch-test/riscv-arch/test`. Previously, it was `riscv-arch-test/work/riscv-arch-test` so the `work` in the middle made it explicit not to be a mistake. Now `work` is created in `sim/work`.

SIG comes from the title of https://github.com/riscv/riscv-arch-test/blob/master/README.md, which is also the name of the mailing list: https://lists.riscv.org/g/sig-arch-test. Honestly, I did not find what does SIG stand for...

stnolting:
> Just for avoiding users/developers being mislead because of having the same name repeated sw/riscv-arch-test/riscv-arch-test. Previously, it was riscv-arch-test/work/riscv-arch-test so the work in the middle made it explicit not to be a mistake. Now work is created in sim/work.

Right. We could also keep `riscv-arch-test`, have no additional `riscv-arch-test` folder and directly clone that repository in that, right?

> SIG comes from the title of https://github.com/riscv/riscv-arch-test/blob/master/README.md, which is also the name of the mailing list: https://lists.riscv.org/g/sig-arch-test. Honestly, I did not find what does SIG stand for...

Now I get it 😆 
I think SIG = special interest group

umarcor:
> Right. We could also keep `riscv-arch-test`, have no additional `riscv-arch-test` folder and directly clone that repository in that, right?

Not really... I thought about it, but there is `port-neorv32`, apart from the submoduled directory. Moreover, in branch `sw-parallel` of my fork (which is a follow-up of this PR), there is also `common.mk` (although we might move that to `port-neorv32`. Anyway, the point is that we need two directories, at least, the submodule and the port. I'm good with having both of them in `sw`, but I was not sure about doing it.

> Now I get it 😆
> I think SIG = special interest group

Thanks! That makes sense :heart:

stnolting:
> Anyway, the point is that we need two directories, at least, the submodule and the port. I'm good with having both of them in sw, but I was not sure about doing it.

Convinced! 👍 

> Maybe sw/riscv-arch-test and sw/riscv-arch-test-neorv32?

Let's keep it in one directory. I would put everything in `sw/riscv-compatibility/` but the term "compatibility" is not really welcomed by RISC-V 😅 So how about just `sw/arch-test/`, `sw/isa-test` or `sw/riscv-test/`? This way there is some "connection" to `sw/processor_check` (which should be `processor_test` when I think about it again).

umarcor:
> Let's keep it in one directory. I would put everything in `sw/riscv-compatibility/` but the term "compatibility" is not really welcomed by RISC-V 😅 So how about just `sw/arch-test/`, `sw/isa-test` or `sw/riscv-test/`? This way there is some "connection" to `sw/processor_check` (which should be `processor_test` when I think about it again).

I now moved it to `sw/isa-test` . I like it. Short (easy to type) and descriptive 😄 

With regard to `processor_check`, actually I believe it should be moved out of `sw/example` (and maybe `demo_freeRTOS` too). However, let's go one step at a time :wink:. If you are good with this, let's have it merged and I will rebase #118 (which is where we are rewroking the examples).

umarcor:
CI running: https://github.com/umarcor/neorv32/actions/runs/1023051910

stnolting:
Let me just update the documentation here (I will also fill the spare parts of `sw/README.md` - otherwise I will for sure forget about these).

stnolting:
> Let me just update the documentation here (I will also fill the spare parts of sw/README.md - otherwise I will for sure forget about these).

Ok, I'll have to do that later 🙈

PR #118, opened 2021-07-11
[sw/example] add common.mk

umarcor:
In order to reduce the maintenance burden of the `sw/example/**/makefile` files, in this PR `sw/example/common.mk` is created, and it is included by all the others.


stnolting:
Good one! I always wanted to simplify that because the current structure requires to change every single makefile if there are new flags to be added. :+1:

I will move `common.mk` to `sw/common` and adapt all the references. I think this is more straightforward. And there already is a makefile in `sw/example` so let's keep things simple.

Also I think that **all** software projects should reference the central makefile - so also the bootloader and the freeRTOS demo. There are some minor edits required in order to pass the correct flags but I will do that.

What do you think? Are you ok with this?

umarcor:
> I will move `common.mk` to `sw/common` and adapt all the references. I think this is more straightforward. And there already is a makefile in `sw/example` so let's keep things simple.

I had some further enhancements in branch https://github.com/umarcor/neorv32/tree/sw-common-mv, because I expected to have this PR merged as is, and discuss those afterwards. Anyway, I now rebased that branch on top of the commits you pushed here, and then updated this PR.

> Also I think that **all** software projects should reference the central makefile - so also the bootloader and the freeRTOS demo. There are some minor edits required in order to pass the correct flags but I will do that.

I did that in `sw/bootloader/makefile` too. Not with freeRTOS. Now, both of them use the common.mk,  after your changes.
 
> What do you think? Are you ok with this?

I would prefer some time to reply to these type of proposals.... I saw the notification, I came to the computer, but by the time I sat down you had already pushed to this PR/branch!
Nevertheless, this is not a complex conflict, so I fixed it.

The main difference is that I `mv sw/common/* sw/`. Since the are three common files only (two plus the new `common.mk`), I don't think it's worth that additional subdirectory.

stnolting:
> I would prefer some time to reply to these type of proposals.... I saw the notification, I came to the computer, but by the time I sat down you had already pushed to this PR/branch!
Nevertheless, this is not a complex conflict, so I fixed it.

Sorry for that 😅 

> The main difference is that I mv sw/common/* sw/. Since the are three common files only (two plus the new common.mk), I don't think it's worth that additional subdirectory.

I think it is cleaner to have linker script, start-up code and main makefile combined in one folder. Especially because there will be at least one `*.o` file generated there. What do you think?

umarcor:
> Sorry for that 😅

No worries :wink: This was an easy conflict 😄 

> I think it is cleaner to have linker script, start-up code and main makefile combined in one folder. Especially because there will be at least one `*.o` file generated there. What do you think?

There will be five files at most:

- `common.mk`
- `crt0.S`
- `crt0.S.o`
- `neorv32.ld`
- `README`

I believe that `common.mk` is now as important as README.md. In fact, given the content we have currently in the README, I would probably propose removing it and writing that at the top of `common.mk`. So, `common.mk` would contain the sequence of targets for most software apps, along with an explanation about which need a particular procedure. Then, there would be 2-3 files "additional" files in the root (`sw/`) which are used by all the software applications indeed.

Anyway, I will reorder the commits, so you can remove the last.

stnolting:
> I believe that common.mk is now as important as README.md. In fact, given the content we have currently in the README, I would probably propose removing it and writing that at the top of common.mk. So, common.mk would contain the sequence of targets for most software apps, along with an explanation about which need a particular procedure. Then, there would be 2-3 "additional" files in the root (sw/) which are used by all the software applications indeed.

I see what you mean.

I just wanted to keep things at a simple level. Beginners do not need to interact with the main makefile, the start-up code or the linker script. So these are "insulated" in a specific folder. Furthermore, there might be several versions of the start-up code and the linker script some day - each one optimized for a certain "abstract setup".

(Sorry for being pedantic sometimes - I just like having files only in leaf directories :smile:)

I think the README is good here, because it gives you a brief overview - especially if you are browsing through the repository on GitHub. The makefile provides a help target that should give some basic information for all target. So I would not put to much information here, because it is all in the data sheet.

umarcor:
Convinced! I will change it when I rebase this after #117 is merged.

umarcor:
This is now rebased and all the common files are kept in `sw/common`, as discussed. CI is running: https://github.com/umarcor/neorv32/actions/runs/1023399339

umarcor:
CI passed successfully.

stnolting:
Great thanks!

PR #120, opened 2021-07-11
[sim/ghdl] use --workdir=build

umarcor:
Instead of executing the simulation in the root of the repo, this PR modifies `sim/ghdl_sim.sh` for running it in subdir `sim` and generating intermediate files in `sim/build`.


stnolting:
Good idea! Let's keep the root directory as tidy as possible 😄

PR #124, opened 2021-07-12
Split ISA test suites in multiple jobs and rework makefiles

umarcor:
This is a follow-up of #117 and it's related to #119.

First, the CI workflow riscv-arch-test is updated for running four parallel jobs, each executing one test suite (I, C, M or privilege).

Then, several optimisations are applied to the makefiles:

- `common.mk` is added so that most of the entrypoints in the suites require 1-3 lines only.
- Instead of using `sed` for overwritting the parameters in the testbench source, top-level generics are used.
- Instead of analising (`ghdl -i`) for each test, the simulation script is split into `ghdl.setup.sh` and `ghdl.run.sh`.


stnolting:
I like this! It's very elegant! :+1:

(sorry for the delay - there was quite a lot to dig through :smile:)

umarcor:
> (sorry for the delay - there was quite a lot to dig through 😄)

Hey! Absolutely no worries or rush at all! Not for this PR, not for any other. We have all our lives ahead!
I recommend having a look at Sections 6 and 7 of https://opensource.guide/best-practices/

PR #125, opened 2021-07-12
Fix source/sink mixup in SLINK docs

jeremyherbert:
Also explicitly state which end controls which signal.


umarcor (approved):
Not necessarily in this PR, but I suggest adding a reference to https://developer.arm.com/documentation/ihi0051/a/Introduction/About-the-AXI4-Stream-protocol. Even though NEORV32 is not AXI4 Stream compliant, it does implement the very basic handshake mechanism.

jeremyherbert:
Added the link also.

stnolting:
Thanks for the fix + improvement! :+1:

Issue #128, opened 2021-07-14
Do we really need wrappers for the top entity?


stnolting:
I have some doubts regarding the current concept of the processor wrapper, i.e. [`rtl/templates/processor`](https://github.com/stnolting/neorv32/tree/master/rtl/templates) and [`rtl/templates/system`](https://github.com/stnolting/neorv32/tree/master/rtl/templates/system). Currently, the `rtl/templates/processor` folder mainly provides wrappers, which have a simplified interface (like only exposing the UART and some GPIOs) but always do some very specific CPU/processor configuration. It feels like these are "prebuilt processor configurations" - like the ATTINY 13A-PU is a prebuilt processor configuration of the AVR family.

I think it would be much cleaner to **always instantiate the processor top entity** `neorv32_top` in all example [`setups`](https://github.com/stnolting/neorv32/tree/master/setups).

My proposal:
1. Modify the processor's top entity so that all configuration options are **disabled by default**.
2. Remove the `rtl/templates/processor` and `rtl/templates/system` folders and all wrappers inside.
3. Provide a single folder (like `rtl/wrappers`) that only contains the following three wrappers:
  * The test setup (currently `rtl/templates/processor/neorv32_ProcessorTop_Test.vhd`) because this is used for the tutorial in the user guide.
  * The wrapper providing resolved interface signals (currently `rtl/templates/processor/neorv32_ProcessorTop_stdlogic.vhd`).
  * The AXI4-Lite wrapper (currently `rtl/templates/system/neorv32_SystemTop_axi4lite.vhd`). I know this is not a wrapper in the classical definitions, but this unit does not instantiate any further modules I think this is ok.

The processor top entity already provides default values for all incoming & outgoing signals. If we add 1.) then all instances of the processor only have to use the generics and signals that are actually required for the setup. For example, the instantiation of the processor in the "test setup" ([`rtl/templates/processor/neorv32_ProcessorTop_Test.vhd`](https://github.com/stnolting/neorv32/blob/master/rtl/templates/processor/neorv32_ProcessorTop_Test.vhd)) could be simplified to look like this:

```vhdl
  neorv32_top_inst: neorv32_top
  generic map (
    -- General --
    CLOCK_FREQUENCY              => 100000000,   -- clock frequency of clk_i in Hz
    INT_BOOTLOADER_EN            => true,        -- boot configuration: true = boot explicit bootloader; false = boot from int/ext (I)MEM
    -- RISC-V CPU Extensions --
[...]

PR #131, opened 2021-07-24
[WIP] setups cleanup (move to neorv32-setups repository)

stnolting:
This is a clean-up PR that removes `setups` and according links, workflows and documentation. All example setups have already been moved to new [`neorv32-setups`](https://github.com/stnolting/neorv32-setups) repository.

Changes made in this PR:
* removed `setups` folder; now located in [`neorv32-setups`](https://github.com/stnolting/neorv32-setups/tree/main)
* removed `rtl/templates` folder; now located in [`neorv32-setups/examples/processor_templates`](https://github.com/stnolting/neorv32-setups/tree/main/examples/processor_templates)
* removed implementation-related workflows:
  * removed `.github/generate-job-matrix.py`
  * removed `.github/workflows/Implementation.yml`
  * removed implementation-part from `.github/workflows/Windows.yml`
* added `rtl/wrappers/` that contains the `std_logic` top entity and the AXI4-Lite top entity
* added `rtl/test_setups/`: contains two minimalist test setups (one using the bootloader, one using pre-installed executable in IMEM ROM) that are used for the reworked "General Hardware Setup" section of the user guide
* modified READMEs and documentation (data sheet & user guide) according to cleaned-up repository

TODO:

* clean-up workflow triggers
* ...? 🤔


umarcor:
@stnolting, I cannot look thoroughly into this at the moment. I just want to point out that CI workflows don't need to be removed here. It is up to you to have CI in a single repo or split between the repos, regardless of where the code is located. I will write later provide examples of cross-triggering between repos and how to do it without cross-triggers.

stnolting:
> I cannot look thoroughly into this at the moment.

No stress! The text here is just a check-list for me :wink:

> I just want to point out that CI workflows don't need to be removed here. It is up to you to have CI in a single repo or split between the repos, regardless of where the code is located. I will write later provide examples of cross-triggering between repos and how to do it without cross-triggers.

I think it is cleaner to keep only the workflows here, that are actually used for the processor core. All implementation-specific workflow should go to `neorv32-setups`. Maybe we can add a trigger _there_ so the workflow gets executed every time `neorv32/rtl` changes. But that is not really urgent right now.

umarcor:
> I think it is cleaner to keep only the workflows here, that are actually used for the processor core. All implementation-specific workflow should go to `neorv32-setups`. Maybe we can add a trigger _there_ so the workflow gets executed every time `neorv32/rtl` changes. But that is not really urgent right now.

The point is that you are proposing to move away all the workflows which test synthesis! So, we are not just reorganising, but losing feedback about a fundamental feature of the project (whether it is synthesisable or "a simulation model only").

We don't need to run the Implementation workflow here. We can just execute `ghdl --synth` here, and have the board specific implementation tests executed elsewhere. However, if the templates are either moved or removed (see #128) we run out of design tops to run the synthesis here.

With regard to triggers, please see [gitter.im/hdl/community?at=6106c042c9f8852a971c798d](https://gitter.im/hdl/community?at=6106c042c9f8852a971c798d). In this case, I believe the best strategy is to have a subset of the tests in this CI, which can be executed without external dependencies (except maybe the riscv-arch-tests submodule). Then, have several tests in the setups repository, executed against the submoduled NEORV32, as well as the 'master' branch. However, I do not recommend setting cross-triggers (due to the permission issues).

stnolting:
> The point is that you are proposing to move away all the workflows which test synthesis! So, we are not just reorganising, but losing feedback about a fundamental feature of the project (whether it is synthesisable or "a simulation model only").

Right, I did not think about this enough. 🙈 
We _do_ need a synthesis target **here**! And I think we need a lot of "different" synthesiseses (plural "synthesis"). Maybe we could use your idea from #128 and run a synthesis for every "single-module-entity-wrapper". 🤔 

> With regard to triggers, please see gitter.im/hdl/community?at=6106c042c9f8852a971c798d. In this case, I believe the best strategy is to have a subset of the tests in this CI, which can be executed without external dependencies (except maybe the riscv-arch-tests submodule). Then, have several tests in the setups repository, executed against the submoduled NEORV32, as well as the 'master' branch. However, I do not recommend setting cross-triggers (due to the permission issues).

I saw your mention there but I did not have time to look through that yet.

stnolting:
Some ideas to get this resolved:

* I like the idea of having explicit setups for "test setups" only - so not tied to any specific board/platform/toolchain. That is the purpose of the `rtl/test_setups/`
* `rtl/wrappers/` should contain wrappers for system integration. I know the term "wrapper" might be misleading, so maybe something like `rtl/integration/` or `rtl/connectivity/` would be better ("system" might also be misleading here).
* Let's undo the removal of `rtl/templates/` :+1: maybe we can name that `processor_templates` to be more specific. But we should rework the templates there. The `Minimal*` can be removed I think. If we use `rtl/test_setups/` then these setups should be used for those minimal "hello world" examples. Additionally, we should clean-up the templates. Maybe we can use something like three templates there but I am not sure about the actual configuration.
* As discussed earlier, let's use some names for those templates that do not provide any connotations at all. You had a great table somewhere (I need to check for that again). Maybe we can use a simplified concept and put that table into the rtl folder's README (not in the actual documentation due to maintaining :wink:)

stnolting:
I thought about this again... Maybe it is a little bit too hasty to make such a huge reorganization (moving things to a different repository). So let's keep the repository in its current state (for now)! But I really want to do a rework of the `rtl` folder:

With regard to the changes in this PR, I would like to split the `rtl` sources into 4 groups:
1. The CPU/Processor core files (a.k.a. the `rtl/core/` directory)
2. System bridges: default top with different bus interfaces (like AXI)
3. Minimal SoC tops to be used as test setups for "first encounters"
4. Exemplary processor templates (-> pre-configured SoCs)

_(maybe goup 3 and 4 could be combined)_

This could look like this:

* `rtl/bridges/` top entity wrappers that provide different interface types/protocols
  * `neorv32_top_axi4lite.vhd` same as default top but with AXI4-lite instead of Wishbone
  * `neorv32_top_stdlogic.vhd` same as default top but using resolved entity signals
  * TODO: Intel Avalon bus, AXI4-Stream
* `rtl/core/` CPU and SoC core files (no README in here!)
* `rtl/templates/` pre-configured example SoCs
  * `neorv32_mars.vhd`
  * `neorv32_jupiter.vhd`
  * ...
* `rtl/test_setups/` minimal test setups; **intended to be used as top entity for "first encounters"** (will be used in the User Guide); only providing GPIO.output ports (8-bit), clock, reset and UART (only the bootloader setup); simplified generics: internal memory size configurations, clock speed
  * `neorv32_test_setup_approm.vhd` use internal IMEM as ROM, pre-initialized with software application
  * `neorv32_test_setup_bootloader.vhd ` use internal IMEM as RAM, software application upload vie UART + bootloader

Regarding the **templates**: the current `neorv32_ProcessorTop_Minimal*.vhd` templates would be replaced by the new test setups. But what about the actual templates in `rtl/templates/`? So the question is what pre-configured SoCs do we want to provide?

Maybe combining `rtl` groups 3 (`rtl/test_setups/`) and 4 (`rtl/templates/`) would make things easier as both folders provide "SoCs" that could be used for the example setups?! :thinking:

umarcor:
> * I like the idea of having explicit setups for "test setups" only - so not tied to any specific board/platform/toolchain. That is the purpose of the `rtl/test_setups/`

👍🏼

> * `rtl/wrappers/` should contain wrappers for system integration. I know the term "wrapper" might be misleading, so maybe something like `rtl/integration/` or `rtl/connectivity/` would be better ("system" might also be misleading here).

👍🏼

In any case, it would be desirable to have a definition/description in `rtl/README.md` which clarifies the content of each of the subdirs.

> * Let's undo the removal of `rtl/templates/` 👍 maybe we can name that `processor_templates` to be more specific. But we should rework the templates there. The `Minimal*` can be removed I think. If we use `rtl/test_setups/` then these setups should be used for those minimal "hello world" examples. Additionally, we should clean-up the templates. Maybe we can use something like three templates there but I am not sure about the actual configuration.

Agree. Let's have #136 and #137 merged first, and then we can rethink the current processor templates. We cannot remove Minimal because it is required in order to have meaningful output in the Fomu. However, we can rename/move/adapt it so that it's used for Fomu only.

> * As discussed earlier, let's use some names for those templates that do not provide any connotations at all. You had a great table ~somewhere (I need to check for that again)~ in [Do we really need wrappers for the top entity? #128](https://github.com/stnolting/neorv32/issues/128). Maybe we can use a _simplified_ concept and put that table into the rtl folder's README (not in the actual documentation due to maintaining 😉)

That sounds nice! Actually, I just reviewed #136 and my last comment was heading in that direction: https://github.com/stnolting/neorv32/pull/136#discussion_r685071582.

Let's flatten the structure, rename `processor` and `system` to what you feel more comfortable with, and have it explained in `rtl/README.md`.

---

> I thought about this again... Maybe it is a little bit too hasty to make such a huge reorganization (moving things to a different repository). So let's keep the repository in its current state (for now)! But I really want to do a rework of the `rtl` folder:

Strictly, I think the "Delayed" label was better than "Wontfix". However, it's ok to have this as wontfix, have the other PRs resolved, and we can later think about what to to put in the other repo.

I agree with reworking the rtl folder, because that's what needs to remain in this repo, regardless of other sources being moved elsewhere.

> With regard to the changes in this PR, I would like to split the `rtl` sources into 4 groups:
> 
> 1. The CPU/Processor core files (a.k.a. the `rtl/core/` directory)
> 2. System bridges: default top with different bus interfaces (like AXI)
> 3. Minimal SoC tops to be used as test setups for "first encounters"
> 4. Exemplary processor templates (-> pre-configured SoCs)
> 
> This could look like this:
> 
> * `rtl/bridges/` top entity wrappers that provide different interface types/protocols
>   * `neorv32_top_axi4lite.vhd` same as default top but with AXI4-lite instead of Wishbone
>   * `neorv32_top_stdlogic.vhd` same as default top but using resolved entity signals
> * TODO: Intel Avalon bus, AXI4-Stream
> * `rtl/core/` CPU and SoC core files (no README in here!)
> * `rtl/templates/` pre-configured example SoCs
>   * `neorv32_mars.vhd`
>   * `neorv32_jupiter.vhd`
>   * ...
> * `rtl/test_setups/` minimal test setups; **intended to be used as top entity for "first encounters"** (will be used in the User Guide); only providing GPIO.output ports (8-bit), clock, reset and UART (only the bootloader setup); simplified generics: internal memory size configurations, clock speed
>   * `neorv32_test_setup_approm.vhd` use internal IMEM as ROM, pre-initialized with software application
>   * `neorv32_test_setup_bootloader.vhd ` use internal IMEM as RAM, software application upload vie UART + bootloader

Ok. I should have read this before... So, we are exactly on the same page: have four subdirs in `rtl`.

> Regarding the **templates**: the current `neorv32_ProcessorTop_Minimal*.vhd` templates would be replaced by the new test setups. But what about the actual templates in `rtl/templates/`? So the question is what pre-configured SoCs do we want to provide? But this is not an urgent question.

Agree with the not urgency. Let's solve #136 and #137, then we'll cleanup the Minimal example and we'll see what we are left with.

stnolting:
> In any case, it would be desirable to have a definition/description in rtl/README.md which clarifies the content of each of the subdirs.

I totally agree. But let's do that in a new PR when we also split `templates` into `processor` and `wrappers`/`connectivity`/`integration`.

> We cannot remove Minimal because it is required in order to have meaningful output in the Fomu. However, we can rename/move/adapt it so that it's used for Fomu only.

I think we could use `neorv32_test_setup_approm.vhd` from #136 instead of the "Minimal" setup. Sure, we would loose the fancy PWM drive, but we could still glue the RGB LED to the GPIO port.

> Strictly, I think the "Delayed" label was better than "Wontfix". However, it's ok to have this as wontfix, have the other PRs resolved, and we can later think about what to to put in the other repo.

:smile: 👍 

> With pyVHDLModel/pyGHDL.dom, we might automatically generate the table that shows which peripherals/features are available in each "template", "system", "bridge", "name it as you want". So, we would not need to manually update it.

That would be awesome!!

If I dare to dream further: teroshdl-documenter (or something similar) to auomatically generate fancy figures of the top entities (showing the available signals and generics)... 😎 :smile:

stnolting:
-> #142, #137, #136

PR #137, opened 2021-08-07
Move osflow examples

stnolting:
This PR moves the board tops used by the open-source flows from `setups/examples` to `osflow/board_top`.

The original `setups/example` folder was meant as a board-wrappers folder that can be used by _any_ toolchain (for example by Lattice Radiant). But I think it is cleaner and easier to understand if we have these board-tops _inside_ the according toolchain folders even if there might be some redundancy (for example is ghdl-yosys and Radiant could use the same board tops).

Another thing is that Radiant and ghdl-yosys seem to expect different primitive libraries:

https://github.com/stnolting/neorv32/blob/70edf18fa3859be2d9c433a731547b1d6a0274f9/setups/examples/neorv32_UPduino_BoardTop_UP5KDemo.vhd#L39-L40

vs.

https://github.com/stnolting/neorv32/blob/70edf18fa3859be2d9c433a731547b1d6a0274f9/setups/radiant/UPduino_v3/neorv32_upduino_v3_top.vhd#L45-L46


setups/osflow/Makefile
```diff
@@ -1,5 +1,5 @@
-OSFLOW := ../osflow
-EXAMPLES := ../examples
+OSFLOW := .
```
umarcor, line 1: I think we can remove OSFLOW and BOARD_TOPS variables here. They don't add any value now.

setups/osflow/README.md
```diff
@@ -6,6 +6,7 @@ open source toolchains. Synthesis based on [ghdl-yosys](https://github.com/ghdl/
 ## Folder Structure
 
 * `boards`: board-specific makefiles for generating bitstreams
```
umarcor, line 8: Maybe specify that these are "partial makefiles". They cannot be used standalone, but are to be included in other makefiles.

umarcor (approved):

setups/osflow/Makefile
```diff
@@ -81,7 +81,7 @@ MixedLanguage:
 	$(MAKE) \
 	  DESIGN=$@ \
 	  DESIGN_SRC=$(TEMPLATES)/processor/neorv32_ProcessorTop_Minimal*.vhd \
-	  NEORV32_VERILOG_SRC='devices/ice40/sb_ice40_components.v ../examples/neorv32_Fomu_MixedLanguage_ClkGen.v' \
+	  NEORV32_VERILOG_SRC='devices/ice40/sb_ice40_components.v ../board_tops/neorv32_Fomu_MixedLanguage_ClkGen.v' \
```
umarcor, line 84: I think you need to remove `../` before board_tops. hat's the CI failure.

setups/osflow/Makefile
stnolting: Good point! :+1:

setups/osflow/README.md
stnolting: Agree :+1:

setups/osflow/Makefile
stnolting: Oh, right...
Somehow, I did not find that bug :smile:

stnolting:
In theory this is ready to merge.

But I think it would be good to do another release of the current setup before merging. Furthermore, we should wait until #139 is merged :wink:

PR #150, opened 2021-09-12
[sim] create subdir 'simple', avoid making a local copy of 'sw' and 'sim'

umarcor:
Currently, when executing riscv_arch_test, a local copy of 'rtl', 'sw' and 'sim' is done. However, only a few RTL sources need to be modified, so copying 'sw' and 'sim' is redundant. In this PR, run/build scripts are modfied for copying 'rtl' only, and using 'sw' and 'sim' from the root of the repo.


stnolting:
Great idea! :+1:

stnolting:
I just adapted the new "sim" target of the application makefile. Maybe we could add another target for running the vunit-based simulation :thinking: (-> another PR/issue).

umarcor:
> I just adapted the new "sim" target of the application makefile.

Thanks! I was not aware of that!

> Maybe we could add another target for running the vunit-based simulation 🤔 (-> another PR/issue).

Let's merge this as-is, and handle that in an upcoming PR. As a matter of fact, the VUnit run script is currently processing all the sources in subdir `sim` (https://github.com/stnolting/neorv32/blob/master/sim/run.py#L23). That means that the "simple" sources are analysed by VUnit (considered part of the project), despite them not being used/executed because the simple testbench is not a "valid" VUnit test.

Therefore, one of the purposes of this PR is to "fix" that. It's not problematic per se, but I believe it's cleaner if the simple sources are ignored by VUnit. A secondary purpose is to completely avoid copying sources for running the architecture tests. This PR is a first step, but I have another branch waiting, which is a subset of #101 indeed. So, let's merge this and discuss that after I rebase.

PR #154, opened 2021-09-16
fix typo wb_mem_a.cyc on sim example

### LINKED ISSUE #153: simulation cycle signal of instruction memory
The external Instruction memory (a) should check for the it valid cycle signal not for the b external memory signal:

https://github.com/stnolting/neorv32/blob/3be750c5e286aca58ae488a78dbecb9ade105269/sim/neorv32_tb.vhd#L544

```diff
+(wb_mem_a.cyc = '1')
-(wb_mem_b.cyc = '1')
```

If its is correct I could generate a PR.

all the best,

rafaelcorsi:
fix #153 

all the best,

### DIFF
--- a/sim/neorv32_tb.vhd
+++ b/sim/neorv32_tb.vhd
@@ -541,7 +541,7 @@ begin
 
         -- bus output register --
         wb_mem_a.err <= '0';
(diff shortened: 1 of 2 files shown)


stnolting:
Thanks for the fix! :smile:
There is the same copy-and-paste bug in `sim/simple/neorv32_tb.simple.vhd`. I will fix that and the merge everything if you are ok with it.

PR #157, opened 2021-09-17
[setups] added CRC32 CFS tutorial (link), by @motius

stnolting:
@ikstvn @turbinenreiter

This PR adds your awesome CRC32 CFS tutorial (https://github.com/motius/neorv32/tree/add-custom-crc32-module) from #147 as link to the README of the `setups` folder. Feel free to customize this PR.

It is just a first step to make it more public. In a next step we can think about adding a link to your branch (or some PDF version?!) to the data sheet / user guide.


turbinenreiter:
Link makes a lot of sense, I think the tutorial is also quite readable directly from the Github markdown rendered page. We will make sure to not purge the branch so the link stays valid :smile: 

We are happy :+1:

PR #159, opened 2021-09-18
[setups/osflow/filesets] do not provide default NEORV32_MEM_SRC, require it to be explicitly set

umarcor:
As the title explains, currently there is a default NEORV32_MEM_SRC variable in `filesets.mk`, which is overriden for some boards/examples. In this PR, that is changed so that no default NEORV32_MEM_SRC is defined. Therefore, for each board/design, the user needs to explicitly select whether to use the default description or some custom sources (such as SPRAM on UP5K).


stnolting:
Makes sense :+1:

PR #163, opened 2021-09-18
[doit] add task Documentation

umarcor:
Similarly to #162, this is a subset of #110.

Instead of calling make in the Documentation workflow, a task named Documentation is added to the `do.py` file.
For now, the following are equivalent:

- `./do.py Documentation`
- `doit -f do.py Documentation`
- `make -C docs`

In upcoming PRs, we might remove `docs/Makefile` and replace it with more specific doit tasks.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -46,8 +46,11 @@ jobs:
       with:
         fetch-depth: 0
 
+    - name: '🐍 Install doit'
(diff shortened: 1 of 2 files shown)


stnolting:
Awesome! :+1:

> In upcoming PRs, we might remove docs/Makefile and replace it with more specific doit tasks.

Let's keep that for now.

PR #164, opened 2021-09-19
[doit] add task RunRISCVArchitectureTests

umarcor:
Similarly to #162 and #163, this is a subset of #110.

Instead of calling `sim/run_riscv_arch_test.sh` directly, it is wrapped in a doit task named RunRISCVArchitectureTests. An option (`-s`|`--suite`) of type "choice" allows selecting which tests to run.

In upcoming PRs, we might remove `sim/run_riscv_arch_test.sh` and replace it with more specific doit/python code. Particularly, https://github.com/stnolting/neorv32/blob/master/sim/run_riscv_arch_test.sh#L49-L61 might be enhanced by using enums/dictionaries.

### DIFF
--- a/.github/workflows/riscv-arch-test.yml
+++ b/.github/workflows/riscv-arch-test.yml
@@ -30,17 +30,19 @@ jobs:
           echo "$GITHUB_WORKSPACE/riscv/bin" >> $GITHUB_PATH
           echo $GITHUB_WORKSPACE
 
-      - name: '🔧 Setup RISC-V GCC'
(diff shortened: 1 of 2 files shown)


stnolting:
> In upcoming PRs, we might remove sim/run_riscv_arch_test.sh and replace it with more specific doit/python code. Particularly, https://github.com/stnolting/neorv32/blob/master/sim/run_riscv_arch_test.sh#L49-L61 might be enhanced by using enums/dictionaries.

Nice idea, but not urgent at all. :+1:

PR #170, opened 2021-10-03
:bug: [rtl/system_integration] fix uart1 rx/tx signals

henrikbrixandersen:
Fix routing of the UART1 rx/tx signals in the AXI4Lite system integration top.


stnolting:
Awesome! Thanks for the fix!

I just realized there are some more signals missing (like the UART flow control signals). I will fix that.

henrikbrixandersen:
> Awesome! Thanks for the fix!

You are most welcome 👍 

> I just realized there are some more signals missing (like the UART flow control signals). I will fix that.

I saw that, thought it was on purpose. It would be nice if those were available there too.

stnolting:
I think I've finally found all the missing signals.
Thanks again!

PR #174, opened 2021-10-08
[setups/vivado] Mention the XHub Stores menu item for installing board support

henrikbrixandersen:
If board support files are not installed prior to creating the project it will fail in a non-obvious way.


stnolting:
Good point! That should be in the documentation for _future_ setups. :+1:

Right now, the Vivado setups do not depend on any pre-installed board support files (everything that is required comes with the setup files).

However, I just stumbled across https://github.com/zephyrproject-rtos/zephyr/pull/39162#pullrequestreview-774704188. Seems like the Vivado TCL files are outdated (wrong reference to `sim` folder; undeclared `fileset_neorv32` variable). I will fix that ASAP :wink:

PR #175, opened 2021-10-08
[setups/vivado] Fix fileset paths

henrikbrixandersen:
Fix the fileset paths in the Vivado project setups. Some were using the wrong path and one was being setup using an undefined variable.


stnolting:
You were faster than me (#174)! 😅 

Thanks for fixing this! :+1:

PR #178, opened 2021-10-10
[rtl/system_integration] add NEOLED FIFO depth generic to AXI4 wrapper

henrikbrixandersen:
Add missing generic for the NEOLED FIFO depth in the AXI4 system wrapper.


stnolting:
Oh these wrappers... 😅 
Thanks for fixing!

PR #186, opened 2021-10-18
fix: reduce_f won't work with single bit operands

gottschalkm:
The or_reduce_f, and_reduce_f, and xor_reduce_f won't work for single bit as the condition fails.
This produces some bugs in the interrupt controller (signal irq_fire) if only one interrupt line is used.


stnolting:
I'm still not sure how to handle all those labels - maybe we ( **I** ) need an issue/discussion to define a list of labels... 😅 
Anyway, thanks for clearing 😄

PR #189, opened 2021-10-28
[setups/ULX3S] add Wishbone interface

stnolting:
This PR is related to the latest discussion in #169. It adds the (yet unused) Wishbone interface to the ULX3S example setup.

The updated example setup now directly instantiates the processor top entity `neorv32_top` instead of one of the templates from `rtl/templates`. Therefore, I have created a new _design_ called "Default" - which might be a misleading name :thinking:. The board's top file name, architecture name, entity name and bitstream name were updated accordingly as well as the corresponding "Implementation" workflow setups.

@umarcor what do you think about this?

Note: I am not sure if this should be merged at all. I just wanted to make a _public_ test of this setup. However, a setup with a ready-to-use Wishbone subsystem might be handy.

PR #194, opened 2021-11-01
SDRAM source included and compiling, not connected

zipotron:


umarcor:
I think this is the type of content that should belong to a neorv32-examples repository, where neorv32 is submoduled.

zipotron:
Hi! I saw the "wishbone" thread and seams good idea, but I couldn't understand where should be located this "neorv32-example" directory, shall be inside "/setups/osflow"?

stnolting:
That would be a completely new GitHub repository at https://github.com/stnolting/neorv32-examples

zipotron:
> That would be a completely new GitHub repository at https://github.com/stnolting/neorv32-examples

Ahh... Well, is just an opinion, as long as we are speaking of something that important for a processor like a par of its memory, I would keep close to it, maybe different directory, because is a specific part and hardware dependent, but accessible. Anyway is just an opinion and I just arrived, then, if you decide to move to another repo, lets move it!
But please, please, please, don't call the repo "examples" because is not an example, for me is another step for try to make a functional core able to run a modern OS (I was thinking for next task how implement a MCU with virtual memory, but maybe is too big for me, I don't know). We can call it "extras" or "flavors" of "plugins" or "hardware_bindings". No?

umarcor:
> Ahh... Well, is just an opinion, as long as we are speaking of something that important for a processor like a part of its memory,

Strictly, NEORV32 does not need this memory. It can already be used on multiple boards, including ULX3S, without this specific memory setup. I don't mean it is not an interesting use case; because I agree it is nice to have it close to NEORV32 and to other examples. However, it is not as important as critical for this processor.

> I would keep close to it, maybe different directory, because is a specific part and hardware dependent, but accessible. Anyway is just an opinion and I just arrived, then, if you decide to move to another repo, lets move it!

From your or any other user's point of view, it would be almost the same. Instead of cloning repo neorv32, you'd clone neorv32-examples. There, you would get the neorv32 subdir, plus all the extras. Moreover, I believe it would be easier to understand for newcomers, because internal/testing resources of neorv32 would be explicitly located in a subdir, different from the "ready to be used" resources.

The complexity is slightly increased for maintainers/contributors/developers only. Particularly, we need to think/solve how to handle CI split across repos.

> But please, please, please, don't call the repo "examples" because is not an example, for me is another step for try to make a functional core able to run a modern OS (I was thinking for next task how implement a MCU with virtual memory, but maybe is too big for me, I don't know). We can call it "extras" or "flavors" of "plugins" or "hardware_bindings". No?

Strictly, running a modern OS with an external RAM and maybe virtual memory is out of scope of NEORV32. It can be done, and it is interesting to have a working example of such solution, but it is far beyond PicoBlaze, MicroBlaze or Cortex-M0 to Cortex-M3, which are the designs that NEORV32 roughly compares to. See:

> https://stnolting.github.io/neorv32/#_rationale
> Why a soft-core processor?
> [...] allow to implement the control flow part of certain applications (like communication protocol handling) using software like plain C.
>
> Yet another RISC-V core? What makes it special?
> [..] The project does not intend to replace certain RISC-V cores or just beat existing ones like VexRISC in terms of performance or SERV in terms of size.

The use case you are pursuing is closer to VexRiscv. Still, as said, I believe it is interesting to have use cases pushing NEORV32 in that direction through external peripherals.

I don't have an strong opinion about the name of the "complementary" repo. neorv32-extras, neorv32-bundle, neorv32-setups... all of those sound good. I will reply in #189 to discuss that more thoroughly. Nonetheless, we don't need to forget the scope. We are providing demos or examples of how things can be done. We are not providing a product to be relied on without proper understanding. So, anyone willing to use the setups for anything other than playing around, should approach them as toy examples and devote time to understand the codebase as if they had written it themselves. Using "examples" makes it explicit, along with being slightly more friendly to newcomers. I believe it is better to have "advanced examples" rather than "simple setups". Anyway, as said, I'm good with using extras, bundle, setups...

stnolting:
@umarcor 
Thanks for the wrap-up! I totally agree! :)

Regarding the OS thing:
The NEORV32 is not intended to run desktop Linux-like OS. The CPU lack the mandatory supervisor mode and the SoC lacks a memory management unit. You could emulate all of that in software - no question - but that would not be a performant setup.

If you need a capable OS - then how about an RTOS? The processor can run FreeRTOS and recently got upstream support by Zephyr OS :wink:

umarcor:
> If you need a capable OS - then how about an RTOS? The processor can run FreeRTOS and recently got upstream support by Zephyr OS 😉

I think this is very relevant. Unlike the ARM hard cores in several FPGAs/SoCs, or many of the soft SoCs seen tested on boards with ECP5 devices, NEORV32 is not suited for a Linux kernel. It is for bare metal, micropython (or similar approaches in golang/rust), RTOS such as Zephyr, etc. Nonetheless, supporting external DDR memory through a wishbone interface is interesting regardless of the software. RTL designs needing to process losts of video/audio in real-time do also need DDR memory for buffering data.

emb4fun:
Hello,

I have changed a little the SDRAM controller from nullobject and added the byteenable
functionality. Furthermore there is now a Wishbone wrapper for it. I tested this on a DE0-nano
board with the neorv32, of course. The memory area of the SDRAM is available from 0x90000000.
The first coding of the addresses is done in the wb_intercon. From the 32 bits, only 28 bits are passed on.
Furthermore, the wbs_stb signal for the slave is only generated if the range 0x90000000 to 0x9FFFFFFF is
accessed. 

I hope that the attachment of the files also worked. 

Best regards,
Michael

[sdram.vhd.txt](https://github.com/stnolting/neorv32/files/7499008/sdram.vhd.txt)
[wb_intercon.vhd.txt](https://github.com/stnolting/neorv32/files/7499009/wb_intercon.vhd.txt)
[wb_sdram.vhd.txt](https://github.com/stnolting/neorv32/files/7499010/wb_sdram.vhd.txt)

emb4fun:
Complete examples for accessing an SDRAM, directly or via Qsys, can now be found here for a DE0-Nano board:

https://github.com/emb4fun/neorv32-examples

zipotron:
Hello @emb4fun !! Sorry I reply you that late. Last two weeks, in my free time,  I was focus in other project, and even didn't login in Github.
Thanks a lot for the code, I am going a see it deeply tomorrow! Also I had a fast look of your repo and is very interesting!
I will create tomorrow a repo for the ULX3S complex setup, don't be offended if instead pushing code in your repo I create another one, is because I use the opensource toolchain, who have different structure from the Altera Suite, that you use. But with pleasure we still sharing ideas and code!
I was a bit demotivated with the SDRAM topic, but you gave me motivation.

Issue #199, opened 2021-11-04
slink variables rx_tmp_v  and tx_tmp_v  have wrong size


LeFl0w:
Hi,
both variables [rx_tmp_v ]( https://github.com/stnolting/neorv32/blob/a498d66a2e7dcd5d314bc8c2633cd6e619a7e653/rtl/core/neorv32_slink.vhd#L250) and [tx_tmp_v ]( https://github.com/stnolting/neorv32/blob/a498d66a2e7dcd5d314bc8c2633cd6e619a7e653/rtl/core/neorv32_slink.vhd#L251) have an upper bound equal to SLINK_NUM_RX-1 and SLINK_NUM_TX-1.

These variables are used as:
-           irq_rx_o <= or_reduce_f(irq_rx_en and rx_tmp_v);
-           irq_tx_o <= or_reduce_f(irq_tx_en and tx_tmp_v);

and refereing to definition of signals:
-       signal irq_rx_en   : std_ulogic_vector(7 downto 0);
-      signal irq_tx_en   : std_ulogic_vector(7 downto 0);

When doing the AND between the two of them with a SLINK_NUM_RX (or SLINK_NUM_TX) not equal to 8, the upper bound do not match.
This lead to a synthesis error on Vivdao 2021.1.

As soon as the vector is initialized entirely [rx_tmp_v := (others => '0');](https://github.com/stnolting/neorv32/blob/a498d66a2e7dcd5d314bc8c2633cd6e619a7e653/rtl/core/neorv32_slink.vhd#L263)
 and           [tx_tmp_v := (others => '0'); ](https://github.com/stnolting/neorv32/blob/a498d66a2e7dcd5d314bc8c2633cd6e619a7e653/rtl/core/neorv32_slink.vhd#L278)

it would be better to set the upper bound of both rx_tmp_v and tx_tmp_v to 7.

stnolting:
Thanks for the hint!
I will have a look and fix that.

PR #200, opened 2021-11-04
[rtl/core] fixing SLINK IRQ bug #199

stnolting:
This PR aims to fix the bug in SLINK's interrupt signal logic from #199.

@LeFl0w


stnolting:
@LeFl0w You're ok with the modifications?

LeFl0w:
I'm not sure about changes for lines 263 and 277.
You come from an iteration over only  the implemented Slink bits to an iteration over every possible links numbers (even unimplemented ones).
This will produce dead code in case of `SLINK_NUM`_ not equal to 8.

I think you could keep the previous code , as your vector is initialized entirely to a default value with `rx_tmp_v := (others => '0')` and then you iterate over only implemented slink bits.

With your new code I 'll have to analyze how `irq_rx_mode(i)`, `rx_fifo_half(i)` and `rx_fifo_avail(i)`  behave when i is not referring to an implemented slink number. In my point of view it is more complicated to analyze without simulation.

stnolting:
> I'm not sure about changes for lines 263 and 277. This will produce dead code in case of SLINK_NUM_ not equal to 8.

That's right, but the synthesis tool should trim these signals.

I noticed that the MSB-aligned bits of `irq_rx_en` and `irq_tx_en` might not be initialized (not written in the bus write access process) if less than 8 links are implemented. That's where those chances came from.

However, feel free to update this PR with another proposal :wink:

stnolting:
This is obsolete now. See #202.

PR #215, opened 2021-11-22
Added a example serial terminal program for Linux 

qwqw330:
Tried the example first with minicom, which didn't get the bootloader interrupt in the beginning. Somehow I couldn't figure out uploading files with screen either. GTKTerm seems to work perfectly fine. Maybe this saves someone else 10-30 minutes of searching.

Also fixed a minor typo in the preceding sentence.

### DIFF
--- a/docs/userguide/content.adoc
+++ b/docs/userguide/content.adoc
@@ -380,8 +380,10 @@ run custom programs on your FPGA setup without having a UART.
 
 [start=1]
 . Connect the primary UART (UART0) interface of your FPGA board to a serial port of your host computer.
-. Start a terminal program. In this tutorial, I am using TeraTerm for Windows. You can download it fore free
-from https://ttssh2.osdn.jp/index.html.en
+. Start a terminal program. In this tutorial, I am using TeraTerm for Windows. You can download it for free
+from https://ttssh2.osdn.jp/index.html.en . On Linux you could use GTKTerm, which you can get here
+https://github.com/Jeija/gtkterm.git (or install via your package manager).
+
 
 [NOTE]
 _Any_ terminal program that can connect to a serial port should work. However, make sure the program


stnolting:
Good Idea! Thanks for contributing! :+1:

Btw, great timing! :wink:
We just had the same issue with minicom on gitter https://gitter.im/neorv32/community?source=orgpage

stnolting (approved):

hipolitoguzman:
I have seen the same problem with minicom with the neorv32 in the iCEBreaker board. Minicom seems to corrupt any program bigger than 4k in size. But when using a different terminal program, such as cutecom, everything works as expected.

qwqw330:
Looks like my pull request was a bit quick. But thanks for merging anyways!
GTKTerm werks perfectly fine, but only for the blinky sketch, which is under 4k of size. Everything bigger than that fails because GTKTerm has this size limit for whatever reason here:
https://github.com/Jeija/gtkterm/blob/c14db185e93aa186db790cec58af3ec99b4f2459/src/serial.h#L32

I just downloaded GTKTerm and set that to and arbitrary value like
```#define BUFFER_EMISSION (4096*8)```                                  
which works fine. Maybe there is a better program somewhere for linux (or we fix GTKTerm)... i don't know

msaufyrohmad:
do this problem with minicom solved? i also having ERR_CHKS on binary more than 4kb

stnolting:
Hey @msaufyrohmad!

I do not think this problem is solved - at least not on the minicom side. Hence, I highly recommend to use a different terminal program. Luckily, there are lots of them out there. :wink:

msaufyrohmad:
yes i am using cutecom and all binaries can be send and execute now. thnks

stnolting:
Very good choice 😉

mlapaj:
I'm using Linux Picocom program
My shell command is (maybe someone else will be intrested):
`sudo picocom /dev/ttyUSBx -b 19200 --send-cmd="ascii-xfr -s -n"`
To send file you need to press ctrl+a and then ctrl+s and type file name.
Using this method, I have uploaded hello app and successfully executed it.

One more thing:
I spent a lot of time with getting ERR_EXE. I thought that issue is with terminal. Ultimately, I realized that I`m loading wrong binary file (main.bin instead of neorv32_raw_exe.bin) . My suggestion is to mention something about it in guide and also in datasheet where ERR_EXE is described (for example to check if header magic is correct). It will reduce time on investigation whats wrong.

Anyway, i have successfully ran hello world example. 
Thanks Stnolting for this project.

stnolting:
Thanks for the information!

> My suggestion is to mention something about it in guide and also in datasheet where ERR_EXE is described (for example to check if header magic is correct). It will reduce time on investigation whats wrong.

Good point. I know that quite a lot people struggle with this (myself included) 😅

PR #216, opened 2021-11-22
[On-Chip Debugger] Fix JTAG timing

stnolting:
This PR reworks the JTAG signal synchronization logic of the DTM (debug transfer module). The pre-PR version of the DTM did not fully comply with the JTAG interface timing requirements (sample TDO and TMS on the _rising_ edge of TCK, update TDO on the _falling_ edge of TCK).

This seems to cause some compatibility troubles with the JLINK EDU as reported by @emb4fun. It would be great if you could test the reworked DTM :wink:

This PR also includes some code clean-ups of the DTM code and also removes the number of recommended debugger wait states to allow faster JTAG operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ defined by the `hw_version_c` constant in the main VHDL package file [`rtl/core/
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:----------:|:-------:|:--------|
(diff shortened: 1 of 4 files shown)


emb4fun:
The J-Link EDU now works with the new PR up to its maximum frequency of 15MHz.
And an Ultra + it went up to 20MHz with a CPU frequency of 100MHz.

stnolting:
Great to hear! Thanks for the feedback! :+1:

PR #223, opened 2021-11-29
Fix neorv32_gpio_port_get()

hipolitoguzman:
This function should read from:
 - NEORV32_GPIO.INPUT_LO and NEORV32_GPIO.INPUT_HI

But previously it was reading from:
 - NEORV32_GPIO.OUTPUT_LO and NEORV32_GPIO.OUTPUT_HI

So it wasn't working correctly. The function neorv32_pin_get() was
reading the correct values (from INPUT instead of OUTPUT) and thus it
was working correctly.


stnolting (approved):

stnolting:
Thanks for fixing!

PR #226, opened 2021-11-30
Add TRNG for UP5KDemo

stnolting:
This PR is just an "idea"... It enables the [TRNG](https://github.com/stnolting/neorv32/blob/master/rtl/core/neorv32_trng.vhd) module for the [UP5KDemo](https://github.com/stnolting/neorv32/blob/master/rtl/processor_templates/neorv32_ProcessorTop_UP5KDemo.vhd) processor template.

For the records: The TRNG uses **latches (!)** and **combinatorial loops (!)**.
I know that both constructs are a bit "delicate" when it comes to FPGAs... So maybe it is a good idea to add the TRNG to check GHDL-yosys can handle such constructs.

This is where the latches and comb. loops are described:
https://github.com/stnolting/neorv32/blob/f8d886711cf0d9de93c28764e5d32bbcdc808c16/rtl/core/neorv32_trng.vhd#L492-L505

@umarcor what do you think? Btw, do you know if there's an option to get a "hardware utilization by entity" report from yosys?

The idea comes from a thread in the German [mikrocontroller.net](https://www.mikrocontroller.net/) forum where someone has problems with latches and GHDL/yosys. Here is the google translated thread: https://www-mikrocontroller-net.translate.goog/topic/528149?_x_tr_sl=de&_x_tr_tl=en&_x_tr_hl=de

**edit**
I have tested the bitstream. The TRNG works! ✔️

### DIFF
--- a/rtl/processor_templates/neorv32_ProcessorTop_UP5KDemo.vhd
+++ b/rtl/processor_templates/neorv32_ProcessorTop_UP5KDemo.vhd
@@ -222,7 +222,7 @@ begin
     IO_TWI_EN                    => IO_TWI_EN,      -- implement two-wire interface (TWI)?
     IO_PWM_NUM_CH                => IO_PWM_NUM_CH,  -- number of PWM channels to implement (0..60); 0 = disabled
     IO_WDT_EN                    => IO_WDT_EN,      -- implement watch dog timer (WDT)?
-    IO_TRNG_EN                   => false,          -- implement true random number generator (TRNG)?
+    IO_TRNG_EN                   => true,           -- implement true random number generator (TRNG)?
     IO_CFS_EN                    => false,          -- implement custom functions subsystem (CFS)?
     IO_CFS_CONFIG                => x"00000000",    -- custom CFS configuration generic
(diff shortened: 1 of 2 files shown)


umarcor:
> what do you think? Btw, do you know if there's an option to get a "hardware utilization by entity" report from yosys?

I would ask Tristan explicitly about the latches being supported. Maybe the behaviour of the TRNG is the expected, but the actual hardware is not exactly what you think. I don't know how to get a hardware utilization by entity. However, you might synthesise that module/component alone (as a top entity itself) and see those results. Furthermore, you can generate a diagram to see what is being synthesised. See ghdl/ghdl#1783.

stnolting:
@tgingold has closed the issue regarding latch support so I think "official latch support" is already there. :+1:

umarcor:
> @tgingold has closed the issue regarding latch support so I think "official latch support" is already there. 👍

He did! However, note that it added it as option `--latches`: https://github.com/ghdl/ghdl/commit/88425cb365578cacd46939d93f837b2ac7b0d5e8.

stnolting:
I did not saw that. Looks interesting! Maybe we should try that. The TRNG seems to be a nice test case.

PR #230, opened 2021-12-03
:bug: Fix Wishbone timeout bug

stnolting:
This PR fixes a bug in the Wishbone timeout counter.

Wishbone timeout configurations (via the `MEM_EXT_TIMEOUT` generic) that are a power of two (like 256) caused an **immediate** timeout exception since the timeout counter was one bit short.

The same problem occurred in the BUSKEEPER's timeout counter. However, this bug did not actually appeared because the default BUSKEEPER timeout is not a power of two.

Thanks to @emb4fun for identifying this bug. :+1:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ defined by the `hw_version_c` constant in the main VHDL package file [`rtl/core/
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:----------:|:-------:|:--------|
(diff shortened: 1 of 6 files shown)


emb4fun:
I had yesterday some problems with running code out of the SDRAM, TIMEOUT_ERR.
After pointing out the problem with the MEM_EXT_TIMEOUT and the latest fix it
is working now.

Before the fix it was not working even with a timeout value of 8191. Now it works
with a timeout of 255. The maximum length of the Avalon waitrequest is 1.68us here.
And with the CPU frequency of 100MHz the timeout of 255 is enough. It looks that
the NEORV32 is more stable than before.

stnolting:
Great to hear! 👍

PR #243, opened 2021-12-22
Fix compile error with questa, for issue #242

### LINKED ISSUE #242: neorv32_package: compile error with Questa
It seems that following line of neorv32_package.vhd isn't compliant to the IEE 1076-2008 standard:
```vhdl
constant def_rst_val_c : std_ulogic := cond_sel_stdulogic_f(dedicated_reset_c, '0', '-');
```
The `cond_sel_stdulogic_f()` function is used after declaring its signature before, but before defining it's body in the package body after.

```
$ vcom -2008 neorv32_package.vhd 
QuestaSim-64 vcom 2021.3 Compiler 2021.07 Jul 13 2021
Start time: 15:04:36 on Dec 22,2021
vcom -2008 neorv32_package.vhd 
-- Loading package STANDARD
-- Loading package TEXTIO
-- Loading package std_logic_1164
-- Loading package NUMERIC_STD
-- Compiling package neorv32_package
** Error (suppressible): neorv32_package.vhd(106): (vcom-1594) Cannot call subprogram "cond_sel_stdulogic_f" before it is elaborated. 
** Note: neorv32_package.vhd(2103): VHDL Compiler exiting
End time: 15:04:36 on Dec 22,2021, Elapsed time: 0:00:00
Errors: 1, Warnings: 0
```

The IEEE 1076-2008 standard in section 14.4.2.1:

> 14.4.2 Elaboration of a declaration
14.4.2.1 General
Elaboration of a declaration has the effect of creating the declared item.
>
> For each declaration, the language rules (in particular scope and visibility rules) are such that it is either
impossible or illegal to use a given item before the elaboration of its corresponding declaration. For
example, it is not possible to use the name of a type for an object declaration before the corresponding type
declaration is elaborated. Similarly, it is illegal to call a subprogram before its corresponding body is
elaborated.

Also see https://stackoverflow.com/questions/29764083/why-cant-i-call-a-function-in-a-constant-declaration-that-is-defined-in-the-sa

A simple fix would be only declaring the constant first and set it after definition of the functions body in the package body. I', preparing a PR which implements that to fix the problem.

tmeissner:
This PR fixes #242 by using a deferred constant. It shouldn't change anything except allowing to compile the package with Questa, which failed before.

### DIFF
--- a/rtl/core/neorv32_package.vhd
+++ b/rtl/core/neorv32_package.vhd
@@ -122,7 +122,8 @@ package neorv32_package is
 
   -- Internal (auto-generated) Configurations -----------------------------------------------
   -- -------------------------------------------------------------------------------------------
-  constant def_rst_val_c : std_ulogic := cond_sel_stdulogic_f(dedicated_reset_c, '0', '-');
+  constant def_rst_val_c : std_ulogic;  -- Use a deferred constant, prevents compile error with Questa
+                                        -- See IEEE 1076-2008 14.4.2.1
 
   -- Processor-Internal Address Space Layout ------------------------------------------------
   -- -------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


stnolting (approved):

stnolting:
Thanks for fixing!

PR #254, opened 2022-01-09
[setups] move to repo neorv32-setups

umarcor:
As discussed in several issues, this is a proposal to move subdir `setups` to a different repository. For now, I created [umarcor/neorv32-setups](https://github.com/umarcor/neorv32-setups). However, I would expect to transfer it to stnolting/neorv32-setups or to neorv32/setups.


stnolting:
Ok, we are doing this! :wink:

Let me just check all the changes from this PR.

umarcor:
As commented in stnolting/neorv32-setups#1, the most relevant changes are the modifications to CI. Workflows *Containers* and *Implementation* are moved to repo neorv32-setups, while *Documentation*, *Processor* and *riscv-arch-test* are kept here. The *Windows* workflow is split into *Implementation* and *Processor*.

Now, we need to deal with documentation. All the refs to the local `setups` subdir are broken. I don't know if you want to keep all the documentation subdir here, or if you want to move anything to neorv32-setups. I think there is no specific section about the setups in the datasheet or in the user guide; it's just some references. Therefore, I believe we could keep the docs untouched (just fix those refs). Later, we can discuss whether/how to add docs to the neorv32-setups repo. I'd like that to be cross-referenced to hdl/constraints, trabucayre/openFPGALoader and other projects in the ecosystem.

stnolting:
> Therefore, I believe we could keep the docs untouched (just fix those refs)

I agree. I will take care of fixing the broken links.

Btw, what about "cross repository" triggers? I think there are possibilities to trigger a workflow in one repository if there was a specific action in another repository. So something like: update rtl core in `neorv32` -> trigger implementation workflow in `neorv32-setups`

umarcor:
After a second thought, since neorv32 is submoduled in neorv32-setups, we would also need to bump the submodule in neorv32-setups when the workflow is triggered from here. Hence, it might be desirable to use dependabot in neorv32-setups instead.

stnolting:
> Hence, it might be desirable to use dependabot in neorv32-setups instead.

Sounds reasonable. I have never used depnedabot before, but I will have a look at the docs.

stnolting (approved):

stnolting:
I think this is ready to be merged.

* All references to `setups` are updated to point to the new repository.
* Seems like we cannot resolve https://github.com/stnolting/neorv32-setups/issues/11 at the moment. However, this is not a repository-related problem.
* I think we should also set a new hardware version so this "major" change also appears in the version history and changelog.

Issue #115, opened 2021-07-10
Add neorv32 to LiteX as soft-cpu option


mithro:
**Is your feature request related to a problem? Please describe.**

[LiteX](https://github.com/enjoy-digital/litex) is an excellent System on Chip build environment [that supports many different soft-CPU cores, including multiple RISC-V cores](https://github.com/enjoy-digital/litex/wiki/CPUs). It also [supports a huge number of FPGA boards](https://github.com/litex-hub/litex-boards).

It would be great to have neorv32 added as an option of soft-cpu to be used.

mithro:
There is some WIP documentation about this at https://github.com/enjoy-digital/litex/wiki/Add-A-New-CPU

umarcor:
Actually, I wanted to give Litex a try after #110. In #110, the filesets are defined in Python, which makes it easier to import them and use them in a Litex script (along with the litex-boards features).

At the moment, filesets, constraints and tools are all defined through makefiles, which would imply duplicating that in the Litex script. Actually, constraints and tools are part of Litex I think, so it would mostly be about the filesets. However, since we have ~6 different places where filesets are defined already, I'd like to avoid increasing the complexity.

stnolting:
That would be really great! :+1: :rocket:
But right now I am not familiar with the Litex build system at all. Thanks for the documentation hints - I will take a closer look :wink:

umarcor:
BTW, this is related to #105, since litedram is a Litex project. So, the easiest path for testing the DDR on the OrangeCrab, Arty, etc. is probably supporting NEORV32 in Litex first.

enjoy-digital:
As discussed in https://github.com/stnolting/neorv32/pull/264, I just experimented with NEORV32 and LiteX. Thanks to NEORV32 very clean code and documentation, it only took a couple of hours to get a first version working in simulation and on hardware:

````
$ litex_sim --cpu-type=neorv32

        __   _ __      _  __
       / /  (_) /____ | |/_/
      / /__/ / __/ -_)>  <
     /____/_/\__/\__/_/|_|
   Build your hardware, easily!

 (c) Copyright 2012-2022 Enjoy-Digital
 (c) Copyright 2007-2015 M-Labs

 BIOS built on Feb 14 2022 16:46:59
 BIOS CRC passed (0a0e47a6)

 Migen git sha1: ac70301
 LiteX git sha1: 0d218306

--=============== SoC ==================--
CPU:		NEORV32 @ 1MHz
[...]

PR #264, opened 2022-01-29
✨[Zxcfu ISA ext.] add option to implement custom RISC-V instructions

stnolting:
With this PR the NEORV32 now provides an option to add **custom RISC-V instructions**. 🚀

This PR adds a _Custom Functions Unit (CFU)_ wrapped in the `Zxcfu` ISA extension, which is a NEORV32-specific custom ISA extension. The extension's name follows the RISC-V naming scheme: 
* `Z` = this is a sub-extension
* `x` = the second letter behind the Z defines the "parent-extension" where this sub-extension belongs to: in this case it belongs to the `X` "custom extensions" extension (platform-specific extension that is not defined by the RISC-V spec.)
* `cfu` = name of the extension (**C**ustom **F**unctions **U**nit)

neorv32_processor

The CFU is implemented as a new hardware module (`rtl/core/neorv32_cpu_cp_cfu.vhd`) that is integrated right into the CPU's ALU. Thus, the CFU has direct access to the core's register file, which provides minimal data transfer latency. A special OPCODE, which has been officially reserved for custom extensions by the RISC-V spec, is used to build custom instructions. The custom instructions supported by the CFU use the **R2-type** format that provides two source registers, one destinations register and a 10-bit immediate (split into two bit-fields:

cfu_r2type_instruction

The `funct7` and `funct3` bit-fields can be used to pass immediates to the CFU for certain computations (for example offsets, addresses, shift-amounts, ...) or they can be used to select the actual custom instruction to be executed (allowing up to 1024 different instructions).

Software can utilize the custom instruction by using the provides intrinsics (defined in `sw/lib/include/neorv32_cpu_cfu.h`. These pre-defined functions implicitly set the `funct3` bit field. Each intrinsic can be treated as "normal C function" (see #263). A simple demo program using the default CFU hardware is available in `sw/example/demo_cfu`.
```c
// custom instruction prototypes
neorv32_cfu_cmd0(funct7, rs1, rs2); // funct3 = 000
neorv32_cfu_cmd1(funct7, rs1, rs2); // funct3 = 001
neorv32_cfu_cmd2(funct7, rs1, rs2); // funct3 = 010
neorv32_cfu_cmd3(funct7, rs1, rs2); // funct3 = 011
neorv32_cfu_cmd4(funct7, rs1, rs2); // funct3 = 100
neorv32_cfu_cmd5(funct7, rs1, rs2); // funct3 = 101
neorv32_cfu_cmd6(funct7, rs1, rs2); // funct3 = 110
neorv32_cfu_cmd7(funct7, rs1, rs2); // funct3 = 111
```

This new feature was highly inspired by @google's [CFU-Playground](https://github.com/google/CFU-Playground) - thanks again to @umarcor for showing me that framework. With some logic plumbing it should be possible to install the CFUs from the CFU-Playground into the NEORV32.

---------------------

### CFU vs. CFS

There are two processor-_internal_ options for custom hardware now: the **Custom Functions Subsystem (CFS)** and the **Custom Functions Unit (CFU)**.

* **Custom Functions Subsystem (CFS):** The CFS is a memory-mapped peripheral that is accessed using load/store instructions. It is intended for complex accelerators that - once triggered - perform some "long" processing in a _CPU-independent manner_ (like a complete AES encryption). The CFS also provides the option to implement custom interfaces as it has direct access to special top entity signals.
* **Custom Functions Unit (CFU):** The CFU is located right inside the CPU's pipeline. It is intended for custom instructions that implement certain functionality, which is not supported by the official (and supported) RISC-V ISA extensions. These instructions should be rather simple data transformations (like bit-reversal, summing elements in a vector, elementary AES _operations_, ...) rather than implementing a complete algorithm (even if this is also supported) since the CFU instructions are absolutely CPU-dependent and will stall the core until completed.


stnolting:
I have added a summarized comparison of the four most obvious (IMHO) options for adding custom hardware modules to the processor (user guide, section "Adding Custom Hardware Modules"). These options are:

* attach custom hardware via the external memory interface (WISHBONE)
* attach custom hardware via the stream link interface (SLINK)
* implement custom hardware via the custom functions subsystem (CFS)
* implement custom hardware via the custom functions unit (CFU)

@umarcor 

We recently had a short discussion about this topic. Could have a look at the comparison table (-> https://github.com/stnolting/neorv32/blob/zxcfu_isa_extension/docs/userguide/adding_custom_hw_modules.adoc#16-comparative-summary)? Maybe you have some ideas for additional (or better) comparison "metrics". :wink:

umarcor:
@stnolting you are so fast! I commented it mostly for gathering some knowledge and you implemented all of it in 1-2 days! That's impressive! Thank you so much!

/cc @tcal-x @mithro might be interested in knowing that NEORV32 supports CFU and might be combined with the content from [google/CFU-Playground](https://github.com/google/CFU-Playground).

docs/datasheet/cpu.adoc
```diff
@@ -684,6 +685,30 @@ high for one cycle to inform the memory system (like the i-cache to perform a fl
 Any additional flags within the `fence.i` instruction word are ignore by the hardware.
 
 
+==== **`Zxcfu`** Custom Instructions Extension (CFU)
+
+The `Zxcfu` presents a NEORV32-specific _custom RISC-V_ ISA extension (`Z` = sub-extension, `x` = platform-specific
+custom extension, `cfu` = name of the custom extension). When enabled via the `CPU_EXTENSION_RISCV_Zxcfu` configuration
+generic, this ISA extensions adds the <<_custom_functions_unit_cfu>> to the CPU core. The CFU is a module that is
+allows to add **custom RISC-V instructions** to the processor core.
```
umarcor, line 693: The CFU is a module that ~is~ allows...

docs/datasheet/cpu_cfu.adoc
```diff
@@ -0,0 +1,154 @@
+<<<
+:sectnums:
+=== Custom Functions Unit (CFU)
```
umarcor, line 3: Fascinating documentation! Congratulations! I'm looking forward to putting this under test with some custom Action Functions for ANN.

docs/datasheet/cpu.adoc
stnolting: I will fix that!

docs/datasheet/cpu_cfu.adoc
stnolting: Thank you very much! :D

tcal-x:
Hi!   That's great news!   Sorry I forgot to follow up on this.   It would be super awesome to have an alternative CPU that would connect to CFUs similar to VexRiscv, even if an adapter is needed (I haven't yet checked out the CFU interface on NEORV32 to see how similar it is).    Connecting VexRiscv to CFU is actually done in LiteX: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/vexriscv/core.py#L275-L328 .    

Does NEORV32 currently plug into LiteX?

stnolting:
> Does NEORV32 currently plug into LiteX?

Not yet, but that is already on the to-do stack (#115) :wink:

>  haven't yet checked out the CFU interface on NEORV32 to see how similar it is

According to [this](https://github.com/google/CFU-Playground/blob/7db6496273afe5174bb6d7b6516c65e6a540f898/proj/mport/cfu.v) CFU-Playground template the interface seems to be quite similar. However, I need to find some real specification for that and test some existing CFU setups (see #269).

umarcor:
FTR, there is https://github.com/umarcor/neorv32-setups/commits/umarcor/edaa, which is a `prj.py` to declare NEORV32 sources through [pyEDAA.ProjectModel](https://edaa-org.github.io/pyEDAA.ProjectModel/). Having it defined in Python should make it easier to reuse in Litex. However, I'm not sure about the steps required to support a new CPU in Litex (most of the uses I see are packaging SoC with already supported cores/modules). I've seen Migen and SpinalHDL designs begin available in the ecosystem, but not VHDL. Is there any Litex example using VHDL or do we need to convert NEORV32 to Verilog?

On the other hand, it might be interesting to add a `.core` file to [stnolting/neorv32-setups](https://github.com/stnolting/neorv32-setups). Can Litex read FuseSoC's `.core` files or does it need some specific declarative format?

/cc @enjoy-digital @olofk

enjoy-digital:
Hi @umarcor,

we currently have one VHDL CPU integrated in LiteX: Microwatt. It can used and integrated as [VHDL](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/microwatt/core.py#L232-L235) or pre-converted to  [Verilog through GHDL/Yosys](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/microwatt/core.py#L217-L231).

To add a CPU, you first need to create the LiteX wrapper around it. You can eventually use these PRs as reference: [CV32E40P](https://github.com/enjoy-digital/litex/pull/535) or [FemtoRV](https://github.com/enjoy-digital/litex/commit/b1d8fe61f8613babeb486877a094bff759e31fd0), with still local sources in a first time. Once working, we could package NEORV32 in a `pythondata-xxyy` package as we are doing for other CPUs.

@stnolting's NEORV32 looks awesome and I would be really happy to help for the integration in LiteX. This would also probably be a good stress test for the GHDL/Yosys plugin :) (Proprietary tools will accept VHDL, but conversion to Verilog is useful to simulate with litex_sim (through Verilator) or to implement it with open-source tools on hardware).

stnolting:
@enjoy-digital 

Thank you very much! I will take a closer look at the Microwatt integration and try to find out how things work :wink:

umarcor:
@enjoy-digital, please, don't do it alone. I mean, @stnolting is the author of NEORV32 as a whole, and I wrote most of the Makefiles in [neorv32-setups](https://github.com/stnolting/neorv32-setups/tree/main/osflow). See the diagram in [hdl.github.io/constraints/Usage](https://hdl.github.io/constraints/Usage.html). So, please, ask as soon as you don't understand anything about the structure.

I suggest you take a look at [processor_templates](https://github.com/stnolting/neorv32/tree/main/rtl/processor_templates) and [system_integration](https://github.com/stnolting/neorv32/tree/main/rtl/system_integration) subdirs in this repo. Rather than integrating `neorv32_cpu`, you might want to start with one of those. With regard to the `board_tops` in [neorv32-setups](https://github.com/stnolting/neorv32-setups/tree/main/osflow), maybe you don't want to use them in LiteX at all (because that's the core functionality of LiteX and you do have litex-boards already), however, they can serve for inspiration.

enjoy-digital:
> @enjoy-digital, please, don't do it alone. I mean, @stnolting is the author of NEORV32 as a whole, and I wrote most of the Makefiles in [neorv32-setups](https://github.com/stnolting/neorv32-setups/tree/main/osflow). See the diagram in [hdl.github.io/constraints/Usage](https://hdl.github.io/constraints/Usage.html). So, please, ask as soon as you don't understand anything about the structure.

Sure, that's what I mean with the integration skeleton. It's easier for me to put things in place for LiteX, when it will be done and if there are issues, I'll be able to share a simulation environment we could use to continue the bringup.

> I suggest you take a look at [processor_templates](https://github.com/stnolting/neorv32/tree/main/rtl/processor_templates) and [system_integration](https://github.com/stnolting/neorv32/tree/main/rtl/system_integration) subdirs in this repo. Rather than integrating neorv32_cpu, you might want to start with one of those. With regard to the board_tops in [neorv32-setups](https://github.com/stnolting/neorv32-setups/tree/main/osflow), maybe you don't want to use them in LiteX at all (because that's the core functionality of LiteX and you do have litex-boards already), however, they can serve for inspiration.

In fact the NeoRV32 CPU seems to be the equivalent of other CPUs integrated in LiteX so it's easier to start with this. If LiteX can also be useful to allow running the NeoRV32 Processor on different hardware or to provide some peripherals not present in NeoRV32 Processor, I'll also be happy to provide help/directions.

enjoy-digital:
@umarcor: This is working and I suggest moving the LiteX specific discussion to https://github.com/stnolting/neorv32/issues/115 instead of this closed PR :)

umarcor:
@enjoy-digital ack and agree. Thanks for you awesomely fast and effective response!

stnolting:
I think I would recommend to use the **processor** top entity ([`rtl/core/neorv32_top.vhd`](https://github.com/stnolting/neorv32/blob/main/rtl/core/neorv32_top.vhd)) for integration (or maybe some of the [template processor wrappers](https://github.com/stnolting/neorv32/tree/main/rtl/processor_templates)). The bus interfaces of the stand-alone CPU are somehow proprietary even though they were inspired by Wishbone.

You can disable all processor-internal modules (even the memories) via generics. By this configuration you can get a processor setup providing just a Wishbone-compatible bus interface.

> This is working and I suggest moving the LiteX specific discussion to https://github.com/stnolting/neorv32/issues/115 instead of this closed PR :)

I agree :wink:

PR #273, opened 2022-02-05
[rtl] add err_o signal to IMEM modules

stnolting:
This PR adds the `err_o` signal to the IMEM module(s). This signal is set if the IMEM is implemented as true ROM and software attempts to perform a write access to it, which will create a "device error".

If there is a write access to a read-only IMEM the [📚 default runtime environment trap handler](https://stnolting.github.io/neorv32/#_default_rte_trap_handlers) will output an error message (in this example: store 32-bit word to address 0x0000_0004):
```
<RTE> Store access fault [DEVICE_ERR] @ PC=0x0000016E, MTVAL=0x00000004 </RTE>
```

**edit**

This PR also add the `err_o` signal to the processor's BOOTROM.

---------

ℹ️ **Before this PR** write access to the IMEM also raised a bus exception by not asserting the IMEM's `ack_o` signal, which resulted in a "device timeout error":
```
<RTE> Store access fault [TIMEOUT_ERR] @ PC=0x0000016E, MTVAL=0x00000004 </RTE>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ The version number is globally defined by the `hw_version_c` constant in the mai
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:----------:|:-------:|:--------|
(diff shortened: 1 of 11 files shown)


stnolting:
⚠️ TODO: update the IMEM modules in neorv32-setups when this is merged

PR #282, opened 2022-03-01
Update neorv32_gptmr.c

prdwivedi:
correct timer source file name, previously clashed in doxygen output


stnolting (approved):

stnolting:
Thank you very much for fixing!

PR #307, opened 2022-04-23
Add .editorconfig

stnolting:
I just stumbled across this: https://editorconfig.org / https://github.com/editorconfig :+1:

I think it would be a good idea to add some guide lines regarding file formatting. What do you think about this?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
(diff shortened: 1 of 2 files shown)


umarcor:
I think it's a nice idea! However, note that not all popular editors support it. For instance, it's not straightforward to use it with VSCode. Therefore, it might worth to add other editor config files as well (in the future).

stnolting:
Good point!
I thought about the `.editorconfig` file/syntax as just some kind of "guide line". Most of the time I am using notepad++, which also does not support this kind of configuration (As far as I know) - but I think that would be ok.

Issue #284, opened 2022-03-05
CSRs and trap while in interrupt handler


Mecrisp:
Moin Stephan,

nice to meet you again in RISC-V territories!

We were busy on the Verilog side, too:
https://github.com/BrunoLevy/learn-fpga/tree/master/FemtoRV/RTL/PROCESSOR

But while tackling full atomic support (hey, you cheated!), I have difficulties with understanding the RISC-V specification in case an illegal instruction/ECALL/EBREAK happens while the processor is busy servicing an interrupt. I failed to find a nice, clean way for trapping while executing an interrupt handler according to the specification, as there seem to be no CSRs for a two-layer-exception.

So when entering an interrupt or exception, PC is saved in MEPC, current opcode is saved in MTVAL, MCAUSE is updated to the source of the exception, and entry point in MTVEC is put into PC. This is fine if there is either an interrupt or an exception, but I miss a second set of CSRs for a second exception layer.

Could you please give me hints or a pointer on how to resolve this conflict in a specification-conformant way? For a freestyle approach, I would just add more CSRs on my own.

Best wishes for your project,
Matthias

stnolting:
Hey Matthias!
Nice to see you again! :)

> We were busy on the Verilog side, too:
https://github.com/BrunoLevy/learn-fpga/tree/master/FemtoRV/RTL/PROCESSOR

I saw that. Great project! :+1:

> But while tackling full atomic support (hey, you cheated!),

Well, somehow... 😅 load-reservate (`lr`) and store-conditional (`sc`) are sufficient to provide atomic access in this single-core setup. For the NEORV32 these instructions are not evaluated by the memory (even though that is supported by the hardware), but by the CPU itself to make sure that lr-sc combinations fail if there is access from another "thread".

Furthermore, the atomic memory operations are hard to implement because they actually execute three operations in one: load from memory, perform some ALU operation, store to memory.

> I have difficulties with understanding the RISC-V specification in case an illegal instruction/ECALL/EBREAK happens while the processor is busy servicing an interrupt.

Short answer: You can't.

Long answer:
Complex RISC-V cores provide additional less-privileged operation modes like "user" or even "supervisor". You can delegate handling of interrupts and exception (using special CSRs like `*ideleg`) to those modes so they are handled - for example - **in** user mode. User mode has it's own set of `epc`, `cause`, `tvec` CSRs to handle them. Thus, you always have machine mode as fall-back for handling traps that occurred in a user-mode trap handler.

[...]

PR #308, opened 2022-04-26
:warning: remove CPU's A ISA extension (atomic memory access)

stnolting:
This PR **removes** the RISC-V **A** ISA extension (atomic memory access) support of the NEORV32. Furthermore, this is a clean-up of the CPU's instruction and data bus interfaces.

## Rationale

The NEORV32 A extensions only supports the load-reservate and store-conditional instruction (`lr.w`, `sc.w`). None of the atomic memory operations (AMO) are supported.

The load-reservate and store-conditional instructions are not required to implement lock-free mutexes and semaphores (the NEORV32 is a single-core setup without DMA). Furthermore, the "exclusive" memory access concept of the core is quite complex and sets high expectations regarding the memory system. Also see the discussion in #284.

However, atomic or exclusive memory accesses even for multi-core/DMA-system can still be implemented by the memory system via dedicated "reservation stations"

## RTL Changes

**Major**
* the `CPU_EXTENSION_RISCV_A` CPU and processor top generic is removed; it is also removed from all wrappers and testbenches
* the Wishbone lock signal `wb_lock_o` is removed from the processor top entity and also from all wrappers and testbenches

**Minor**
* the `i_bus_lock_o` and `d_bus_lock_o` output signals of the CPU top entity are removed
* the `i_bus_priv_o` and `d_bus_priv_o` output signals of the CPU top entity are removed and are replaced by a new "global" `priv_o` output signal
* the `i_bus_wdata_o`, `i_bus_ben_o` and `i_bus_re_o` output signals of the CPU top entity are removed (those signals were hardwired to zero)


NikLeberg:
Hi @stnolting,
pretty old pr, but I'm wondering what you meant by:

> However, atomic or exclusive memory accesses even for multi-core/DMA-system can still be implemented by the memory system via dedicated "reservation stations"

Could you explain what you mean by that? Is there some more terminology such that google finds relevant content? 😀
Or do you mean things like the [Peterson's algorithm](https://en.wikipedia.org/wiki/Peterson's_algorithm)?

PR #315, opened 2022-05-12
:bug: fix buskeeper timeout error

### LINKED ISSUE #314: Bus_keeper module incorrectly generates bus errors
**Observed behavior**
NeoRV32 crashes at random when executing from external DDR2 memory.

**Cause**
The bus_keeper module incorrectly generates an *internal* bus error on an *external* cycle when the wishbone acknowledge arrives exactly when the internal bus cycle timeout counter is at zero. Not before, not after; only exactly at zero. It seems that this happens because the bus_xip_i signal is zero in this cycle, so the cycle is incorrectly marked as internal.
 
**Screenshots**
image

stnolting:
Fixes #314. @GideonZ

The BUSKEEPER raises an _internal timeout error_ if an external access (via Wishbone) has the same access latency as there are maximum _internal_ timeout cycles (`max_proc_int_response_time_c`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ mimpid = 0x01040312 => 01.04.03.12 => Version 01.04.03.12 => v1.4.3.12
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:----------:|:-------:|:--------|
+| 13.05.2022 | 1.7.1.6 | :bug: fixed bug in **BUSKEEPER** timeout logic; [#315](https://github.com/stnolting/neorv32/pull/315) |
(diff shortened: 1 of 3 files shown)


GideonZ:
Hi Stephan!

Yes, I think you can safely merge this fix. It solves the crashes I was
seeing. I am still a bit confused to see the err_i signals fire during
normal operation. It makes it a bit difficult to trigger on such a fault in
SignalTap. But aside from that, this fix is very important, so please go
ahead and merge it.

Best regards,
Gideon

stnolting:
> Yes, I think you can safely merge this fix. It solves the crashes I was
seeing. [...] But aside from that, this fix is very important, so please go
ahead and merge it.

:+1:

> I am still a bit confused to see the err_i signals fire during
normal operation. It makes it a bit difficult to trigger on such a fault in
SignalTap.

Hmm, that's strange. Do you see this just during accesses via the external memory interface?

PR #324, opened 2022-05-21
Constructors

GideonZ:
As promised, hereby the pull request with the necessary changes to the crt0.s and linker script to support the proper initialization of C++ objects by calling the constructors. A small demo shows that the static objects have been initialized properly.

### DIFF
--- a/sw/common/crt0.S
+++ b/sw/common/crt0.S
@@ -190,6 +190,21 @@ __crt0_clear_bss_loop:
 
 __crt0_clear_bss_loop_end:
 
+// ************************************************************************************************
+// Call constructors
+// ************************************************************************************************
+__crt0_call_constructors:
+  la   s3,  __init_array_start
+  la   s4,  __init_array_end
+
+__crt0_call_constructors_loop:
+  bge  s3,  s4, __crt0_call_constructors_loop_end
+  lw   t1, 0(s3)
+  jalr ra, 0(t1)
+  addi s3,  s3, 4
+  j    __crt0_call_constructors_loop
+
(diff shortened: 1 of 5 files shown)


stnolting:
This is amazing!
I did not expect that a c++ setup would work this flawlessly - actually, I did not expect that it would work at all! 😅

Calling the constructor and destructors "manually" from crt0 (and not using `call __libc_init_array`) seems to be a nice way to avoid library code being pulled in. There shouldn't be any functional difference, right?

I will have a closer look at your code but I think I am absolutely fine with merging this as-is. :+1:

sw/common/crt0.S
```diff
@@ -190,6 +190,21 @@ __crt0_clear_bss_loop:
 
 __crt0_clear_bss_loop_end:
 
+// ************************************************************************************************
+// Call constructors
+// ************************************************************************************************
+__crt0_call_constructors:
```
stnolting, line 196: Bootloader ROM is precious and I do not think a bootloader requires constructors or destructors, so maybe we should exclude the according code parts when compiling the bootloader (using `#ifndef make_bootloader`)?!

stnolting: Added.

stnolting (approved):

stnolting:
@GideonZ you are OK with the changes?

GideonZ:
> @GideonZ you are OK with the changes?

Absolutely. I saw a few typos in the documentation that you may want to have a look at, but it's just minor. Actually I admire your diligence of updating the documentation with these changes!

PR #326, opened 2022-05-22
Fixes concurrent ecall / IRQ issue.

GideonZ:
Runs in simulation, still has to be tested on hardware. More commits might follow.
This pull request is just to let you know and avoid that we are both doing the same work at the same time.

### DIFF
--- a/rtl/core/neorv32_cpu_control.vhd
+++ b/rtl/core/neorv32_cpu_control.vhd
@@ -245,6 +245,9 @@ architecture neorv32_cpu_control_rtl of neorv32_cpu_control is
     env_start_ack : std_ulogic; -- start of trap handler acknowledged
     env_end       : std_ulogic; -- end trap handler env
     --
+    irq_start     : std_ulogic; -- start trap handler irq
+    irq_start_ack : std_ulogic; -- start of trap handler acknowledged
+    --
     instr_be      : std_ulogic; -- instruction fetch bus error
     instr_ma      : std_ulogic; -- instruction fetch misaligned address
     instr_il      : std_ulogic; -- illegal instruction
@@ -774,7 +777,7 @@ begin
     ctrl_o(ctrl_ir_funct3_2_c   downto ctrl_ir_funct3_0_c)  <= execute_engine.i_reg(instr_funct3_msb_c  downto instr_funct3_lsb_c);
     -- cpu status --
     ctrl_o(ctrl_sleep_c)         <= execute_engine.sleep; -- cpu is in sleep mode
-    ctrl_o(ctrl_trap_c)          <= trap_ctrl.env_start_ack; -- cpu is starting a trap handler
+    ctrl_o(ctrl_trap_c)          <= trap_ctrl.env_start_ack or trap_ctrl.irq_start_ack; -- cpu is starting a trap handler
     ctrl_o(ctrl_debug_running_c) <= debug_ctrl.running; -- cpu is currently in debug mode
     -- FPU rounding mode --
     if (CPU_EXTENSION_RISCV_Zfinx = true) then
@@ -909,6 +912,7 @@ begin
     -- trap environment control --
     trap_ctrl.env_start_ack     <= '0';
     trap_ctrl.env_end           <= '0';
+    trap_ctrl.irq_start_ack     <= '0';
 
     -- leave debug mode --
     debug_ctrl.dret             <= '0';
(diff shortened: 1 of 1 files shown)


GideonZ:
This change breaks the debugger, so it is not good yet.

umarcor:
@GideonZ, I changed this PR to a draft, to prevent unwanted merges. Please, feel free to mark it as "Ready for review" as soon as you feel it to be good.

GideonZ:
It may be superfluous; but this is my fix for the concurrent trap / irq issue. I tested this on the hw and the debugger works again; including single stepping.

I will be checking your fix as well; probably tomorrow.

stnolting:
I have just tested your fix. Looks good so far, but there is a problem with the `ebreak` instruction. When running the [processor_check](https://github.com/stnolting/neorv32/tree/main/sw/example/processor_check) program via GDB the program should enter debug mode when executing `ebreak` as this is interpreted as a "software" breakpoint. However, the processor from this PR seems to ignore that.

GideonZ:
Interesting. I tried this, and you are right, you cannot step over the ```ebreak```. I tried this in instruction stepping mode the debugger comes back on the same address every time and you can step as many times as you like, it doesn't hang per se. I see in SignalTap that it never leaves debug mode; PC stays in the FFFFF800 range, too. 

image

How important is it to fix my fix, if your fix already works? This is beyond my skills to fix, I think, as I don't exactly know how the debugger works.

stnolting:
I think this is because of a false `xpc` (trap return address) assignment. The EBREAK-TRAP and the SINGLE-STEP-IRQ might arrive at the same time here, but EBREAK seems to be evaluated first.

> How important is it to fix my fix, if your fix already works? This is beyond my skills to fix, I think, as I don't exactly know how the debugger works.

I have looked at your code again and there is one "thing" that hurts my eyes: :wink:
The signals `trap_ctrl.cause` and `trap_ctrl.irq_start` are written from two different if-blocks. Sure, this is fine as VHDL ensures the latter assignment gets priority, but this makes the code (at least for me 😅) hard to understand and hard to debug. Furthermore, this might have unintended side effects (on a "behavioral" level and also on resource utilization as we have an "implicit" mux here).

I did a quick synthesis using my default FPGA test setup:

|                 | :bug: `main`| #326  | #327 |
|:----------------|:-----------:|:-----:|:----:|
| Logic elements: | 6170        |  6152 | 6129 |
| Registers:      | 3209        |  3210 | 3211 |

(Timing is not affected at all.)

I really appreciate your fix (:+1:), but I think we should focus on #327. What do you think?

GideonZ:
I didn't mean to hurt your eyes. 😉 It is quite common in VHDL to start
with default assignments and then overwrite stuff in 'if' statements. This
is also the reason that I usually put the reset clause at the bottom of the
process, not at the top. Because... if there is something you do not want
to reset in the process (e.g. synchronization flops), the 'else' will cause
the flipflops that are not reset to keep their previous value during reset.
And, this causes more luts to be used, too. Anyway, this is "viel Getue um
nichts".

And sure, let's drop #326 and go for #327. In every case I prefer a fix by
the author over my own fix. I just had to find out if a fix for this
problem also fixed the random crashes I was seeing, so I went ahead to make
a temporary change. (And indeed; the random crashes are gone.)  I'll try
#327 tonight.

stnolting:
> I didn't mean to hurt your eyes. 😉 It is quite common in VHDL to start
with default assignments and then overwrite stuff in 'if' statements. This
is also the reason that I usually put the reset clause at the bottom of the
process, not at the top.

Don't get me wrong! I think this is just a matter of taste. :wink:

> Anyway, this is "viel Getue um nichts".

😅 

> And sure, let's drop [#326](https://github.com/stnolting/neorv32/pull/326) and go for [#327](https://github.com/stnolting/neorv32/pull/327). In every case I prefer a fix by
the author over my own fix. I just had to find out if a fix for this
problem also fixed the random crashes I was seeing, so I went ahead to make
a temporary change. (And indeed; the random crashes are gone.)  I'll try
[#327](https://github.com/stnolting/neorv32/pull/327) tonight.

I did some more testing and from my side everything looks good.

GideonZ:
I have to check it still, but I think so.  I have been working on other
things and didn't have much time to look into RiscV. Will try to look into
it tonight or tomorrow morning.

stnolting:
Great, thank you very much! :+1:

PR #329, opened 2022-05-29
:bug: fix debugger single-instruction stepping mode

stnolting:
This PR fixes the single-instruction (`stepi`) stepping issue explained in https://github.com/stnolting/neorv32/pull/327#issuecomment-1140450822: when doing single-instruction-stepping the debugger does not halt at _every_ single instruction.

ℹ️ This bug has been introduced with version 1.7.1.9 by PR https://github.com/stnolting/neorv32/pull/327.


stnolting:
Hey @GideonZ,

I need to get this merged to continue with some other hardware modifications. I have tested this PR and I could not find any situations where we have the concurrent trap situation again.

However, if you happen to find any problems with the modifications please let me know. :wink:

Issue #332, opened 2022-06-02
neorv32_twi_start_trans() does not work with prsc>1


andkae:
Hi Stephan,

at the moment i do some tests with the [TWI](https://github.com/stnolting/neorv32/blob/main/rtl/core/neorv32_twi.vhd). My belonging C code looks like follows:

```c
/** Includes **/
/* Standard libs */
#include <string.h>     // memset

/* Processor */
#include <neorv32.h>
#include <neorv32_twi.h>


/**
 *  main
 */
int main()
{
    /** Variables **/

    /* init UART */
    neorv32_uart0_setup(115200, PARITY_NONE, FLOW_CONTROL_NONE);

    /* entry message */
    neorv32_uart0_print("Interfaces I2C Core of NEORV32\n");

    /* setup I2C */
    neorv32_twi_disable();
        // neorv32_twi_setup(uint8_t prsc)
    neorv32_twi_setup(4);
    neorv32_twi_mack_enable();
    neorv32_twi_enable();

    /* write to I2C */
    neorv32_twi_start_trans(0x55);
    neorv32_twi_generate_stop();

    while ( 1 ) { } // inifinite loop
    return 0;
}
```

For asserting the transfer i use the NEORV32 function neorv32_twi_start_trans. For small prescaler values like 1 works this very well.

PRSC=1:
![image](https://user-images.githubusercontent.com/106740887/171612964-fbdc0616-67b4-43b4-a1c8-d5fab0c67349.png)

For PRSC=4 is only the starbit sended and then happends nothing
![image](https://user-images.githubusercontent.com/106740887/171613063-848a4d3d-be7f-4ee7-8a30-8dccb3248d1d.png)

I did some investigation, the issue seems like follows, if a higher PRSC value is used, then is the FSM in the startbit state _arbiter=101_ and the write of ths address - in my case 0x55 - is skipped:
![image](https://user-images.githubusercontent.com/106740887/171613378-3af415d4-2413-440a-acbe-f86e85efe92e.png)

One possible solution could be to poll the busy flag, this would lead to an change in neorv32_twi_start_trans() and the IRQ driven I2C would not work anymore.

What are your thoughts?

BR,
Andreas

stnolting:
Hey Andreas,

you need to generate a start conditions first before you can do actual TWI transfers. So you need to call `neorv32_twi_generate_start` first:

```c
    /* write to I2C */
    neorv32_twi_generate_start(); // generate START conditions before triggering actual transfers
[...]

Issue #330, opened 2022-06-01
SPI Outputs not initializied with Reset


akaeba:
Hi Stephan,

if the Reset is applied to the neorv32 the outputs of the spi are staying uninitilized. The [SPI](https://github.com/stnolting/neorv32/blob/main/rtl/core/neorv32_spi.vhd) uses also only a clock. To avoid unpredictable behaviour needs at least the _CSn_ outputs an reset defined output value. This avoid accidentally activation of the driven SPI component. Better would to assign also CLK and MOSI. In my use case are 180us necessary to assign an valid value.

![spi_1](https://user-images.githubusercontent.com/29815957/171497111-37fe0f88-ab3f-456a-97bc-7534dfe7b4d1.png)

![spi_2](https://user-images.githubusercontent.com/29815957/171497135-c169e265-6c0e-440a-85ce-d611be3e4dfd.png)

If you like i can submit a patch. What do you think?

Greetings,
Andreas

stnolting:
Hey Andreas,

the SPI chip-select lines _are_ initilized, but it takes quite some time since they are configured by a _software reset_ (`crt0` writing zero to all IO device's registers) rater than by an actual hardware reset.

I thought this would not cause any troubles with SPI devices. All outgoing SPI signals might be undefined, but for sure there are NO clock pulses generated on the SPI clock line so any SPI peripheral should not perform any internal operations. We also have the same situation during power-up of the FPGA where all FPGA pins might be in tr-istate mode until the FPGA is configured.

However, I am open for a discussion. Maybe it is about time to add a _real_ hardware reset to all IO/peripheral devices. 🤔

andkae:
For the SPI looks also good, now is the Chipselect de-asserted with the reset:

![image](https://user-images.githubusercontent.com/106740887/171808265-27aedf33-0450-4299-9ad2-e731620d6234.png)

Thanks for providing the fix!

stnolting:
Awesome! The MOSI signal seems to clear correctly now, too

Same question here: you have used the modified version from #334, didn't you?

andkae:
yes i used #334

PR #334, opened 2022-06-02
[rtl] add hardware reset to IO/peripheral devices

stnolting:
This PR (finally?) adds a _dedicated hardware reset_ to all IO/peripheral devices.

After having too much trouble with hardware-vs-simulation mismatches (recently: #330, #332) it is about time to add a real reset to the IO devices. Lessons learned. 🙈 😅 

### Current State (pre-PR)

The current setup relies on a pure software reset scheme, where `crt0` is responsible for clearing/resetting all IO devices by writing zero to all memory-mapped registers. Clearing the control register of a device (or to be more specific: the control register's enable bit) will disable and reset the device putting it in a defined state.

### Proposed State (this PR)

This PR adds a hardware reset to each IO module's interface registers. Module-internal register still do not provide a dedicated hardware reset!

All _interface register_ are set to zero by the hardware reset - including the control register's enable bit, which will cause the rest of the logic (the internal registers that do not have a dedicated reset) to be reset to a defined state.

The `crt0` "hardware reset" code is no longer required and is removed. This makes booting slightly faster.

ℹ️ On Quartus Prime 21.1 (targeting a Cyclone IV FPGA) the modifications from this PR have absolutely no impact on device utilization.

### TODO

* [x] rtl modifications (add reset)
* [x] update documentation
* [x] remove `crt0` IO initialization routine
* [x] check (simulation & on real hardware) that this PR does not break anything
* [x] update pre-built application images (bootloader and blink_led)


rtl/core/neorv32_spi.vhd
```diff
@@ -128,9 +129,13 @@ begin
...
   -- Read/Write Access ----------------------------------------------------------------------
   -- -------------------------------------------------------------------------------------------
-  rw_access: process(clk_i)
+  rw_access: process(rstn_i, clk_i)
   begin
-    if rising_edge(clk_i) then
+    if (rstn_i = '0') then
+      ctrl   <= (others => '0');
+      ack_o  <= '-';
+      data_o <= (others => '-');
```
andkae, line 137: Is there a special Reason, why there is a dont care '-' assigned?

stnolting: Some registers do not require a reset at all because they will be updated with valid data in the following cycle anyway. In this specific case, we do not care about `ack_o` right after reset because there won't be any pending CPU access (for sure). In the very next cycle `ack_o` will become zero until the device is actually accessed by the processor.

In the end, using `'-'` as reset value will make the synthesis tool (tested with Quartus and Vivado) _not_ use the according FF's reset input at all. For FPGAs this is quite irrelevant, but for ASICs this might reduce footprint and power consumption as simpler register cells can be used (those without a reset logic).

andkae: okay, understand :)

PR #493, opened 2023-02-12
[sw/lib/include/neorv32.h]: remove redundant uart typedef

akaeba:
As discussed in #492 is for UART0 and UART1 the same register layout definition used.


stnolting:
Thanks for cleaning up!

stnolting (approved):

PR #494, opened 2023-02-12
Add run.py dump of VHDL-LS library mapping

kraigher:
Hi I am working on a VHDL language server (https://github.com/VHDL-LS/rust_hdl) that I think could be useful to you when developing neorv32.
It is able to analyze the entire neorv32codebase as well as its dependencies VUnit and OSVVM.
It provides type checking, goto definition and find references for the entire codebase and takes just 120 ms to load it.
The configuration file that provides the library mapping can be automatically generated by the VUnit run.py file and is provided by this PR. To run it you need to install the toml python package from pypi.

The easiest way to try is to install the VSCode extension: https://marketplace.visualstudio.com/items?itemName=hbohlin.vhdl-ls
The language server is a 100% static binary that is installed automatically by the extension.
Since it is a language server (https://microsoft.github.io/language-server-protocol/) is will work in emacs and vim as well if you prefer those instead of VSCode.


stnolting:
Hey there.

This looks like a really great tool! I have vscode installed, but I never really tried that as I was not sure about the tool's VHDL capabilities 😅

> The configuration file that provides the library mapping can be automatically generated by the VUnit run.py file and is provided by this PR. To run it you need to install the toml python package from pypi.

Just to make sure I understand this correctly: the modifications to the do.py script will generate the configuration file? So this could be uploaded as GitHub workflow asset for people to "just download and use it"?

The workflow currently fails because of the missing `toml` module. I think it should be no problem to add this to the workflow configuration file, right?

kraigher:
1)
Yes the code I added to the `run.py` will create a `vhdl_ls.toml` file in the repository root on every invocation.
The `vhdl_ls.toml` files is automatically loaded by VHDL-LS if you open the repo root folder in the VSCode extension.

2)
Yes `toml` is a Python dependency that is not installed, the easiest way to solve it is to install it. If you really want to avoid dependencies it is quite easy to create a toml string as well, the format is similar to a .ini file.

kraigher:
> Just to make sure I understand this correctly: the modifications to the do.py script will generate the configuration file? So this could be uploaded as GitHub workflow asset for people to "just download and use it"?

I guess a blocker to having the `vhdl_ls.toml` files as a GitHub artifact to just download is that it will need to standardize where the VUnit files are located. If you install VUnit from pip the files will not end up in the same place on each computer.

kraigher:
I pushed another commit to the PR now that will only generate the `vhdl_ls.toml` file if the `toml` package is install. Thus if will fail gracefully for users and in the CI-pipeline.

stnolting:
Thanks for clearing!

I think I really like this. I need to test this on my system...

Anyway, would you add some words (links) to the user guide how to setup/install this? I think this would be something people (including me!) could benefit from. 😅

kraigher:
> Anyway, would you add some words (links) to the user guide how to setup/install this? I think this would be something people (including me!) could benefit from. 😅

Sure where in the user guide should I put it?

I pushed another commit that writes the `.toml` files in a dependency free way without requiring the `toml` package

stnolting:
> Sure where in the user guide should I put it?

Great, thanks!
I think a completely new section would be nice - but I'll leave that up to you.

> I pushed another commit that writes the .toml files in a dependency free way without requiring the toml package

👍:+1:

kraigher:
@stnolting I pushed a new `VHDL development environment` chapter now. I am unsure about the syntax of adoc but I guess it is generated by CI in the pipeline so I can look at it.

stnolting:
That was fast! :D
Looks good for me!

Issue #495, opened 2023-02-13
GCC version, zicsr extension requirement


alpkoyun:
**Describe the bug**
Hello, the default compilation architecture of the bootloader is rv32i but newer versions of the riscv gcc requires the zicsr extension to handle processor register instructions. Since bootloader has a lot of these instructions, compilations results in an error  below
../../sw/lib/include/neorv32_cpu.h:261: Error: unrecognized opcode `csrc 768,a5', extension `zicsr' required

**To Reproduce**
Compile the bootloader code my "make bootloader" with a new version of the riscv gcc

**Additional context**
I realize that the default architecture is set to rv32i so that it can work on different cores but this causes the problem mentioned above.

stnolting:
You are right! Newer versions of RISC-V GCC require `_zicsr` to be added to the `MARCH` string. Since this is already a mature feature I think we can modify the project's makefile to default to `march=rv32i_zicsr`. I will open a pull request for this.

Thank you very much for the hint! :+1:

PR #496, opened 2023-02-13
[sw] add '_zicsr' to default MARCH configuration

stnolting:
Fixing #495.

@alpkoyun, what do you think about this?


alpkoyun:
> Fixing #495.
> 
> @alpkoyun, what do you think about this?

Seems to work, thanks for the update.

PR #507, opened 2023-02-25
Changed proposal for default ICACHE_NUM_BLOCKS.

emb4fun:
Testing with Dhrystone and CoreMark has shown that 32x64 is the optimal size for the ICACHE if the code is executed from the SDRAM of the DE0-Nano, for example. This gives a total size of 2K for the cache. Greater values than 2K no longer had a big impact on Dhrystone and CoreMark.

### DIFF
--- a/rtl/core/neorv32_top.vhd
+++ b/rtl/core/neorv32_top.vhd
@@ -92,7 +92,7 @@ entity neorv32_top is
 
     -- Internal Instruction Cache (iCACHE) --
     ICACHE_EN                    : boolean := false;  -- implement instruction cache
-    ICACHE_NUM_BLOCKS            : natural := 4;      -- i-cache: number of blocks (min 1), has to be a power of 2
+    ICACHE_NUM_BLOCKS            : natural := 32;     -- i-cache: number of blocks (min 1), has to be a power of 2
     ICACHE_BLOCK_SIZE            : natural := 64;     -- i-cache: block size in bytes (min 4), has to be a power of 2
     ICACHE_ASSOCIATIVITY         : natural := 1;      -- i-cache: associativity / number of sets (1=direct_mapped), has to be a power of 2
 


GideonZ:
Hello Michael,

Thank you for considering different cache configurations. I am curious how
you have done this. Running the benchmark in simulation would be ideal, as
it can be fully scripted for many different configurations.
I am very surprised that a block size of 64 seems to be ideal. I find this
rather large. Where does this benefit of a large block size come from? Is
this because the latency of the SDRAM is hidden more often? I would say
that large blocks can also hurt real bad, if only a part of a block is ever
used.

I would actually love to see an automated testbench for this in the
repository, that can be adapted to actual SDRAM configuration. For
instance, my SDRAM controller can currently only do single bursts (4 beats
of 8 bits = 32 bits), using auto precharge for every request. I need this
because of the guaranteed maximum latency of ~80 ns. Multiple accesses that
are sequential are still at 100% bus utilization, because they use
alternating banks.  The latency of the SDRAM may be larger on the Cyclone
V, but the rate is higher.

Have you also looked into different associativity settings? I'd say that 2
is already a lot better than 1, if I remember correctly from my computer
architecture class 25+ years ago. I understand that higher associativity
settings will also increase the LUT count.

If this testbench is already there, please accept my apology. I have a hard
time keeping track of every detail that is going on in this project.

GideonMessage ID: ***@***.***>

>

emb4fun:
Hello GideonZ,

> I am curious how you have done this.

I have tested "all" combination to check if the result gets better or not. Here I have used the Dhrystone and CoreMark benchmark in TCM memory. The results looks like:

Dhrystone: 0.83
CoreMark : 0.95

After this I have checked the values without cache running out of the SDRAM:

Dhrystone: 0.21
CoreMark : 0.22

After this I have checked different combinations for the cache setting.

Even I have a Crypto benchmark which is running only in SDRAM because of the code size. Here it turned out that a size of 64 is the best one. And a cache associativity of 2.

I will make the results public next time on my project side here:
https://www.emb4fun.de/riscv/de0nref/index.html

Best regards,
Michael

emb4fun:
Some additional note about the ICACHE_ASSOCIATIVITY. It has been shown here that with a value of 2,
the performance of the crypto benchmark could be increased again slightly. But in the case of the Dhrystone
or CoreMark, there were no changes. However, since there is also an additional consumption of resources,
therefore you should stick to the default value of 1 for ICACHE_ASSOCIATIVITY.

stnolting:
The problem with caches is that the "optimal" configuration highly depends on the program being executed and also on the processor-external memory system.

Having a "fits most" configuration would be a nice thing, but we should keep the resource consumption in mind. Obviously, higher associativity can increase efficiency but will also increase the cache's hardware requirements by 100%.

stnolting:
> But if you use the cache, then it shouldn't get any worse than without a cache.

Yes and no 😅
If you only use the internal memories of the processor (latency = 1 clock cycle) and write an application that just jumps from one cache page boundary to the next one then the instruction cache will slow down execution by a factor of N = number of words per cache block.

> If we want to keep the 4x64 as the default then we can close this PR.

This was just some "random" value I have chosen. And actually, such a configuration does not make a lot of sense - at least I have never seen a CPU with 256 bytes of instruction cache :wink:

So, yes, we should provide a more realistic configuration here.

emb4fun:
There is no universal solution. It always depends on the respective application. That's why you have to experiment with the settings for the cache yourself. That's why we can't really make a suggestion here.

For this reason I have now closed the PR here.

PR #533, opened 2023-03-05
⚠️ rework UART modules

stnolting:
This PR is a complete rework of the processor UART modules (_"back to the basics"_).

* remove support for RTS/CTS hardware flow-control
* change and clean-up of control and data register
* remove support for partiy bit; frame format is fixed to 8N1
* clean-up UART VHDL code
* rework interrupt configutation (new style 😎)
* improve inference/mapping of FIFO blockRAM

ℹ️ If you require any of the removed features:
* add a more complex UART via the Wishbone module
* or use the "old" NEORV32 UART (e.g. from a previous release) and add that as custom module to the processor's custom functions subsystem


andkae:
One alternativ could be the heavily in X86 systems used [D16550](https://opencores.org/projects/uart16550) uart.

stnolting:
@andkae That looks like a real nice UART - it even provides Wishbone connectivity and should integrate seamlessly. Thanks for the hint!

PR #542, opened 2023-03-08
[sw/lib] move register and bit definitions to according module header files

emb4fun:
The idea behind this is to also define the information of a device in its header file.

Using the GPIO as an example, I moved the definition of the structure from neorv32.h
to neorv32_gpio.h. Only the define for NEORV32_GPIO_BASE remains in the neorv32.h file.

Later you can then summarize all the BASE addresses again in the neorv32.h file.

What do you think of this proposal?


andkae:
In my opinion looks really good!. The Hardware adresses are in the neorv32.h header and the rest of the sw layer is in the driver itself. 

Question how to deal with Bit definition in other modules?

stnolting:
> Question how to deal with Bit definition in other modules?

What do you mean exactly?
The register bits are defined using simple C enumerations. So we could move them to the according *.h file as well.

andkae:
> What do you mean exactly?

Yes exactaly the enums, but for sure they can also move to the *.h files.

stnolting:
@emb4fun looks good so far! Thank you very much!

The list of base addresses might cause some issues with Doxygen. But that is not critical. I can fix that after merging this.

In your Pr we have three new files: for the system information memory (SYSINFO), the debug module (DM) and the bus monitor (BUSKEEPER). Having a new file for SYSINFO is just fine. Same for the BUSKEEPER. But what about the DM?

The debug module cannot be accessed by "normal" software running on the core. All the register and bit definition were just listed for "completeness". So I am not sure what we should do with this file?

By the way, I will modify your PR and add the new files to the software sections file list. :wink:

emb4fun:
The DM file was listed for completeness too. Because all structure definitions of a "device" was moved to its own file.

stnolting:
The question is do we really want to keep this file. 🤔

Anyway, I'll have a look at your modifications. 👍

stnolting (approved):

stnolting:
Seems ready. @emb4fun what do you think?

emb4fun:
Yes it is ready. @stnolting please merge.

PR #551, opened 2023-03-16
Added support for USER_LIBS in the command SW makefile.

vhdlnerd:
The user can now specify custom libraries for the linker to include via the new `USER_LIBS` make variable.
Also updated the docs to describe `USER_LIBS`.


stnolting:
Looks good! Thank you very much!

stnolting (approved):

PR #552, opened 2023-03-16
[sw] bootloader: send wake up command to flash before trying to speak with it

agamez:
The ice40 family of FPGAs, send a sleep command to the flash once it's finished with self-programming in order to save some power [1], unless instructed not to when generating the bitfile (icepack -s or equivalent setting on Radiant)

We don't want the bootloader to depend on the bitfile generation settings and thus we wake up the flash, but since we want a bootloader as small as possible, we don't care to put it back to sleep again (because we don't even know if it was sleeping or not) and let the user application decide what to do with it.

[1] https://www.latticesemi.com/~/media/LatticeSemi/Documents/ApplicationNotes/IK/FPGA-TN-02001-30-iCE40-Programming-Configuration.pdf?document_id=46502 , page 20


stnolting:
I just updated the bootloader's documentation section and I rebuild the bootloader image. If this is fine for you we can merge this!

stnolting (approved):

agamez:
Oh, I had forgotten about the documentation! And it's ironic because one of the reasons I decided to dive into neorv32 is the fantastic documentation. I really do have a lot to learn from you and this project.

Everything looks fine for me too!

stnolting (approved):

PR #561, opened 2023-03-25
Remove warnings

emb4fun:
Remove some GCC warnings


stnolting:
Could you provide some more information? What is this `pragma GCC diagnostic ignored "-Wtype-limits"` about?

emb4fun:
I like to try to have 0 warnings. I got the following warning:

comparison is always true due to limited range of data type [-Wtype-limits]

This is generated by the following line:

`if ((id >= RTE_TRAP_I_MISALIGNED) && (id <= RTE_TRAP_FIRQ_15)) {`

**id** is here a uint8_t. This means that any value greater than is equal RTE_TRAP_I_MISALIGNED, because RTE_TRAP_I_MISALIGNED is 0. Actually, you would have to change the code as follows to no longer get the warning:

`if (id <= RTE_TRAP_FIRQ_15) {`

But I didn't want to make the change to keep the understanding of the code. That's why I turned off the warning for the small part.

stnolting:
Ah, now I understand.

I would actually _do_ change the code base here. Having those pragmas makes the code base harder to understand. Furthermore, these pragmas might not be supported by other compilers like LLVM.

So here is my proposal: add a "last" identifier to the trap ID enum:
```c
  RTE_TRAP_FIRQ_14      = 27, /**< Fast interrupt channel 14 */
  RTE_TRAP_FIRQ_15      = 28, /**< Fast interrupt channel 15 */

  RTE_TRAP_enum_last = RTE_TRAP_FIRQ_15
};
```

and then use that for the range check:

```c
  // id valid?
  if (id > RTE_TRAP_enum_last) {
```

Furthermore, I would change the type of the `id` argument from `uint8_t` to plain `int`. This seems to be more straightforward here.

Arun789788 (changes requested):
sw/lib/source/neorv32_rte.c

stnolting:
Hey @Arun789788

> sw/lib/source/neorv32_rte.c

What do you mean with that?

stnolting (approved):

stnolting:
Thank you very much for fixing!

PR #566, opened 2023-04-01
[PMP] add support for NA4 and NAPOT modes

stnolting:
* :sparkles: add support for `NA4` (naturally aligned power-of-two region, 4 byte wide) and `NAPOT` (naturally aligned power-of-two region, larger than 4 bytes) to the processor's physical memory protection system - PMP is now full spec-compliant
* :warning: change behavior of `neorv32_cpu_pmp_configure_region()` function


stnolting:
I am locking for a real-world application to further test the PMP system. A test suite for RISCOF is on its way (https://github.com/riscv-non-isa/riscv-arch-test/pull/284) but this is still work-in-progress. So if anyone has an idea or even a complete test setup, it would be really great if you could share it 👍

PR #574, opened 2023-04-11
Updated path for FreeRTOS example - FreeRTOS-Plus-TCP

matty0005:
I know that perhaps a lot of additional work is needed to get networking working on this processor, so this prehaps isn't the most useful fix. 

Nevertheless, FreeRTOS have put all the source files (for only the TCP stuff) under `$(FREERTOS_HOME)/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/source/` and not `$(FREERTOS_HOME)/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/` as it currently is in the makefile in the FreeRTOS example. 

This PR just updates those routes. 

Users would also need to include a `FreeRTOSIPConfig.h` file as well. I have left that out for now, but I can add a note about that in the readme for the example. I can also provide a default one too - just wanted to see everyone's thoughts first.

### DIFF
--- a/sw/example/demo_freeRTOS/makefile
+++ b/sw/example/demo_freeRTOS/makefile
@@ -78,18 +78,18 @@ endif
 # FreeRTOS-Plus-TCP
 # -----------------
 ifneq (,$(findstring FREERTOS_PLUS_TCP,$(USER_FLAGS)))
-APP_SRC += $(FREERTOS_HOME)/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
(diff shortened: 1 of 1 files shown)


emb4fun:
Hello matty,

do you still have a MAC for the PHY or what kind of ethernet hw do you want to use?

Regards,
Michael

stnolting:
> This PR just updates those routes.

Thanks for updating!

> Users would also need to include a FreeRTOSIPConfig.h file as well. I have left that out for now, but I can add a note about that in the readme for the example. I can also provide a default one too - just wanted to see everyone's thoughts first.

That would be great! :wink:

matty0005:
Great, I'll create another PR later once I have things working for the `FreeRTOSIPConfig.h` and can add some notes where appropriate.

PR #576, opened 2023-04-12
RFC: [rtl] Replace peripheral's direct access to XXX_base_c constants with a generic

agamez:
Hi!

This commit is not intended to be merged as-is because it doesn't really accomplish anything, but I thought I would write it anyway so the concept is more clear than just the explanation below.

This won't change generated netlist and doesn't truly allow to select a different base address that is already defined as a constant in neorv32_package because address decoding is made using the whole 32 bits words, so changing just the base address won't make any module recognize access to its own registers.

Next step should be processing only the lower bits of the address to decide on which register to act upon. Once this is done, it should be much easier adding different devices of the same kind to the SoC. There are already two UARTs that would be best candidates to test this concept on.

The idea would be to replace, for instance:

```
constant uart0_ctrl_addr_c    : std_ulogic_vector(31 downto 0) := x"ffffffa0";
if (addr = uart_id_ctrl_addr_c) then -- control register
```
with something like this

```
constant uart_addr_high : natural := 7;
constant uart_ctrl_addr_c    : std_ulogic_vector(uart_addr_high downto 0) := x"a0";
if (addr(uart_addr_high downto 0) = uart_id_ctrl_addr_c) then -- control register
```

so this would remove all conditionals on neorv32_uart.vhd and serve as an example of how one can add more devices of the same kind with a little bit less of trouble.


stnolting:
Hey @agamez!

I really like this idea!

In general, the processor setup was not designed to allow adding further modules to the processor itself. This is what the external memory interface should be used for.

But: the CFS (custom functions subsystem) could be used to replicate a set of already present peripherals (like UART, SPI, TWI, ...). And for this use-case it would be really nice to have a generic configuration option for the module's base address (and the address space size).

From a hardware point of view this requires two additional generics per module:
* base address (seems like you have already implemented this)
* address space size: basically this would be the module's `*_size_c` constant from the package file
* we should also remove all register definitions (like `constant uart0_ctrl_addr_c := ...`) from the package and from the according module and replace them with hard-coded offsets

agamez:
> Hey @agamez!
> 
> I really like this idea!

Great, glad to hear it!

> In general, the processor setup was not designed to allow adding further modules to the processor itself. This is what the external memory interface should be used for.
> 
> But: the CFS (custom functions subsystem) could be used to replicate a set of already present peripherals (like UART, SPI, TWI, ...). And for this use-case it would be really nice to have a generic configuration option for the module's base address (and the address space size).

Yes, the nice thing is that this would allow reusing the already existing devices. Each module would even be completely independent of neorv32 and would be fit for other SoCs, and viceversa.

> 
> From a hardware point of view this requires two additional generics per module:
> 
>     * base address (seems like you have already implemented this)
> 
>     * address space size: basically this would be the module's `*_size_c` constant from the package file

Does the module need this information at all? It already knows how many registers it has and its size, so I think this would be redundant.

>     * we should also remove all register definitions (like `constant uart0_ctrl_addr_c := ...`) from the package and from the according module and replace them with hard-coded offsets

Something like this:

```
diff --git a/rtl/core/neorv32_package.vhd b/rtl/core/neorv32_package.vhd
index 05ebe672..3113b4dd 100644
--- a/rtl/core/neorv32_package.vhd
+++ b/rtl/core/neorv32_package.vhd
@@ -225,8 +225,9 @@ package neorv32_package is
   -- Serial Data Interface (SDI) --
   constant sdi_base_c           : std_ulogic_vector(31 downto 0) := x"ffffff00"; -- base address
   constant sdi_size_c           : natural := 2*4; -- module's address space size in bytes
-  constant sdi_ctrl_addr_c      : std_ulogic_vector(31 downto 0) := x"ffffff00";
-  constant sdi_rtx_addr_c       : std_ulogic_vector(31 downto 0) := x"ffffff04";
+  constant sdi_addr_high_c      : natural := integer(log2(real(cfs_size_c))) - 1;
+  constant sdi_ctrl_addr_c      : std_ulogic_vector(sdi_ctrl_addr_c downto 0) := x"00";
+  constant sdi_rtx_addr_c       : std_ulogic_vector(sdi_ctrl_addr_c downto 0) := x"04";
 
```

stnolting:
> Yes, the nice thing is that this would allow reusing the already existing devices. Each module would even be completely independent of neorv32 and would be fit for other SoCs, and viceversa.

👍

> Does the module need this information at all? It already knows how many registers it has and its size, so I think this would be redundant.

Basically you are right. The modules "know" how much address space they occupy. I just kept all the address space sizes in one place (the package file) to have a general overview - and to prevent address space collisions/overlapping.

> Something like this:

Maybe the actual base address would be sufficient. The individual register addresses are module-relative, so no need to keep them in the package file. Same for the address space size. But as mentioned earlier, then we need to double check there is no collision and no overlap. :wink:

agamez:
> > Yes, the nice thing is that this would allow reusing the already existing devices. Each module would even be completely independent of neorv32 and would be fit for other SoCs, and viceversa.
> 
> +1
> 
> > Does the module need this information at all? It already knows how many registers it has and its size, so I think this would be redundant.
> 
> Basically you are right. The modules "know" how much address space they occupy. I just kept all the address space sizes in one place (the package file) to have a general overview - and to prevent address space collisions/overlapping.
> 
> > Something like this:
> 
> Maybe the actual base address would be sufficient. The individual register addresses are module-relative, so no need to keep them in the package file. Same for the address space size. But as mentioned earlier, then we need to double check there is no collision and no overlap. wink

Does it even really matter that much the real size that each module occupies? Most of them use a few bytes (8,16) and only cfs uses 256 bytes. We have 4G of address space and I doubt anyone is going to connect a 4Gb ram to a SoC like the neorv32. Maybe we can just preassign 1K to each peripheral and be done with it? We would be using less than 30K for peripherals and it would allow to change any of them to add a couple more registers without having to modify all the addresses of all other modules. If, say, you want to add a second SPI device, you could just reduce its address space to a half and place both of them in the previous space of just the one without having to touch anything else.

This, by the way, is the same approach that Xilinx uses, which is reserving up to 4K of space for every module.

agamez:
I've added a new commit for SPI peripheral to show what would the changes entail to all the other modules.

There remains the question of whether we want to keep XXX_size_c constants on the package, if we want to move them to the peripheral, or if we want to just assign 1K to each one regardless of the real use it does inside.

If you like the direction this is taking, I will perform the same changes on the other peripherals and then we could test it all together and merge it.

One test is failing, it is because I use this VHDL'08 idiom:
`constant spi_ctrl_offset_c      : std_ulogic_vector(lo_abb_c-1 downto 0) := 3x"0";`

I can just replace it with something that doesn't use the Nx"0" format if you want.

stnolting:
> There remains the question of whether we want to keep XXX_size_c constants on the package, if we want to move them to the peripheral, or if we want to just assign 1K to each one regardless of the real use it does inside.

Hm, good point. But this would require a rework of the entire address space. There are some modules that have special address requirements. For example, the debug module has to be mapped to an address that has a maximum offset of +/-4kB around address 0x00000000.

agamez:
> > There remains the question of whether we want to keep XXX_size_c constants on the package, if we want to move them to the peripheral, or if we want to just assign 1K to each one regardless of the real use it does inside.
> 
> Hm, good point. But this would require a rework of the entire address space. There are some modules that have special address requirements. For example, the debug module has to be mapped to an address that has a maximum offset of +/-4kB to address 0x00000000.
> 
> Right now, all internal peripherals are mapped to the very last 512 Bytes of the address space. The "coarse" module selection (i.e. access address inside this 512 bytes block) is done globally for all modules so the individual modules can use a simpler address decoding. I know this will only save a hand full of LUTs, but that was the initial design idea. wink

For now, I've just converted all the modules to this new pattern but leaving the XXX_size_c constant in the package description, just so it's easier to verify that there's no overlap between modules. We can think about it later if you want

> > One test is failing, it is because I use this VHDL'08 idiom:
> 
> This is another topic we should discuss at some time.

Indeed. For now, I've added some commits to remove VHDL'08 specifics. I've had to convert some 'switch' into 'if chains' because of some errors regarding locality (see https://github.com/ghdl/ghdl/issues/2368#issuecomment-1506713094 and https://github.com/ghdl/ghdl/issues/2254). Some of the switch cases were nice to have, but some others were making code harder to understand because it was not using the register constants names, but instead hardcoding again the bits, so in the end I think it's still a win, even if it's uglier.

stnolting:
I did not have the time yet to see through your modification - sorry for that 🙈

Anyway, having a base address generic for each module is a great idea! But that alone would not allow to instantiate additional modules - at least not outside of the core itself. The problem is the global address pre-decoding for the peripheral that I have mentioned above:

> Right now, all internal peripherals are mapped to the very last 512 Bytes of the address space. The "coarse" module selection (i.e. access address inside this 512 bytes block) is done globally for all modules so the individual modules can use a simpler address decoding. I know this will only save a hand full of LUTs, but that was the initial design idea. 😉

So maybe we should add a second generic to each module: a mask that defines which bits of the access address should be checked. Maybe something like this:

```vhdl
ADDR_BASE : std_ulogic_vector(31 downto 0);
ADDR_MASK : std_ulogic_vector(31 downto 0);

...

acc_en <= '1' when ((addr_i and ADDR_MASK) = (ADDR_BASE and ADDR_MASK)) else '0';
```

Of course we also would need to internally mask-out the actual register offset bits. But as you have already mentioned, each module "knows" how much bits are needed to distinguish between all those register.

agamez:
> I did not have the time yet to see through your modification - sorry for that see_no_evil

Don't worry, I neither have the time I'd like to spend on this, though it is quite fun!
 
> Anyway, having a base address generic for each module is a great idea! But that alone would not allow to instantiate additional modules - at least not outside of the core itself. The problem is the global address pre-decoding for the peripheral that I have mentioned above:
> 
> > Right now, all internal peripherals are mapped to the very last 512 Bytes of the address space. The "coarse" module selection (i.e. access address inside this 512 bytes block) is done globally for all modules so the individual modules can use a simpler address decoding. I know this will only save a hand full of LUTs, but that was the initial design idea. wink

Right, this is this line of code:

`io_acc  <= '1' when (p_bus.addr(31 downto index_size_f(io_size_c)) = io_base_c(31 downto index_size_f(io_size_c))) else '0';`

That is a simple nice way of reducing logic on the peripherals. But I believe there's something missing there, and I think it should be neorv32_top responsibility to provide the modules with a proper address. If we are preselecting an address space with this bit of logic, then the peripherals could just receive `p_bus.addr(index_size_f(io_size_c) downto 0)`.

So what we're looking at instead from this point of view is that maybe modules should completely disregard the whole address and just decode its offset: let top module fully manage address decoding. In fact, now that I think of it, I think this is the usual coding for other IP out there.

I'm just pushing a new commit that exemplifies this concept, but it won't probably synthesize at all, it's just for you to see if you like the idea before I go all in.

stnolting:
> That is a simple nice way of reducing logic on the peripherals. But I believe there's something missing there, and I think it should be neorv32_top responsibility to provide the modules with a proper address

The full address is already supplied by the top. We just do a simple pre-decoding here to check if the IO region is accessed at all. In the default NEORV32 version, each module then just checks the address bits within this IO region to check if it is actually accessed at all.

> If we are preselecting an address space with this bit of logic, then the peripherals could just receive p_bus.addr(index_size_f(io_size_c) downto 0).

If we provide less then the full 32-bit address then this would make re-instantiating the same module again pretty impossible as you would loose address granularity/flexibility.

> So what we're looking at instead from this point of view is that maybe modules should completely disregard the whole address and just decode its offset: let top module fully manage address decoding. In fact, now that I think of it, I think this is the usual coding for other IP out there.

Right, most SoC design use a centralized address decoding. I decided against this because:
* we can do some pre-decoding as mentioned above
* a centralized decoding logic would need to be aware of the actual address space width of each individual module (e.g. a UART has only two memory-mapped registers, the CFS has 64)
* the de-centralized approach seemed to be more clear as every module is fed with the full address bus - just checking the address bits that are relevant for it

agamez:
> > If we are preselecting an address space with this bit of logic, then the peripherals could just receive p_bus.addr(index_size_f(io_size_c) downto 0).
> 
> If we provide less then the full 32-bit address then this would make re-instantiating the same module again pretty impossible as you would loose address granularity/flexibility.

Well, it would just require for address decoding to be external. The decoder inside the module would only decode the offset itself (which has to do anyway). But it is true that it would require a full decoder if anyone wants to implement a second module using the external wishbone bus, instead of adding it to the core.

> > So what we're looking at instead from this point of view is that maybe modules should completely disregard the whole address and just decode its offset: let top module fully manage address decoding. In fact, now that I think of it, I think this is the usual coding for other IP out there.
> 
> Right, most SoC design use a centralized address decoding. I decided against this because:
> 
>     * we can do some pre-decoding as mentioned above
> 
>     * a centralized decoding logic would need to be aware of the actual address space width of each individual module (e.g. a UART has only two memory-mapped registers, the CFS has 64)
> 

I think the total number of LUTs would not be very different, as instead of doing

`acc_en <= '1' when ((addr_i and ADDR_MASK) = (ADDR_BASE and ADDR_MASK)) else '0';`

on every module, we would have that on the outside. And given that we intend to provide ADDR_MASK via a generic, there still needs to be declared the address space outside of the peripheral, in the top or in the top package.

>     * the de-centralized approach seemed to be more clear as every module is fed with the full address bus - just checking the address bits that are relevant for it

Well, for me it's being pretty confusing analyzing the logic inside the module that determines whether it's being accessed or not. I guess it's just a matter of what you're used to. AXI interconnects do the full decoding and the AXI slaves only have to decode the offset for its own registers. As I've said a lot before, that is my previous experience so I tend to look at it all with Xilinx-colored glasses. Xilinx glasses are also known to scratch with just a speck of dust, so there's that too :sweat_smile:

stnolting:
Hey @agamez!

> Well, for me it's being pretty confusing analyzing the logic inside the module that determines whether it's being accessed or not

The concept being used by the NEORV32 is quite unique I think 😅 Basically, the read-back path of the bus access is split across the module's output registers. Instead of using a single (hug!) mux that selects the output of one specific module, the address decoding is split and only the module being actually accessed outputs some data (otherwise all modules just output all-zero). Thus, a single huge OR gate is sufficient to wire-up all module's output ports.

I like this concept, even if it is a bit harder to understand as it allows to split the memory systems critical path into "access phase" and "read-back phase".

Right now I am doing a rework of the processor-internal bus system to use custom types instead of distinct interface signals:

```vhdl
  -- Internal Bus Interface: Request --------------------------------------------------------
  type bus_req_t is record
    addr : std_ulogic_vector(31 downto 0); -- access address
    data : std_ulogic_vector(31 downto 0); -- write data
    ben  : std_ulogic_vector(03 downto 0); -- byte enable
    we   : std_ulogic; -- write request (single-shot)
    re   : std_ulogic; -- read request (single-shot)
    src  : std_ulogic; -- access source (1=instruction fetch, 0=data access)
    priv : std_ulogic; -- set if privileged (machine-mode) access
  end record;

  -- Internal Bus Interface: Response -------------------------------------------------------
  type bus_rsp_t is record
    data : std_ulogic_vector(31 downto 0); -- read data
    ack  : std_ulogic; -- access acknowledge (single-shot)
    err  : std_ulogic; -- access error (single-shot)
  end record;
```

The intention here is to make the code more readable and also to reduce manual "wiring effort" :wink:

Unfortunately, this might introduce lots of merge conflicts with your PR branch... (sorry for that! 🙈)

However, I would like to keep up your idea of generalizing the access mechanisms!

agamez:
Hi @stnolting 

> Unfortunately, this might introduce lots of merge conflicts with your PR branch... (sorry for that! :see_no_evil:)

Lol, don't worry, you need to break some eggs if you want an omelette! I always worked on this on the assumption that it would be requiring some additional work before merging. I've been a bit busy and I have a few days of extra work ahead, so it'll take some time, though.

If you want to keep the mix of pre-decoding + decentralized full decoding inside the modules, then the address mask idea passed through a generic seems definitely the way to go. On the other hand, how do you feel about replacing all these:

```
  acc_en <= '1' when (bus_req_i.addr(hi_abb_c downto lo_abb_c) = cfs_base_c(hi_abb_c downto lo_abb_c)) else '0';
  addr   <= cfs_base_c(31 downto lo_abb_c) & bus_req_i.addr(lo_abb_c-1 downto 2) & "00"; -- word aligned
  wren   <= acc_en and bus_req_i.we; -- only full-word write accesses are supported
  rden   <= acc_en and bus_req_i.re; -- read accesses always return a full 32-bit word
```

with a new record type like this that can be returned by a function?
```
type acc_t is record
  addr : std_ulogic_vector(31 downto 0);
  wren  : std_ulogic; -- access acknowledge (single-shot)
  rden : std_ulogic; -- access error (single-shot)
end record;

function bus_acc_en(addr : std_ulogic_vector(31 downto 0); addr_mask : std_ulogic_vector(31 downto 0); addr_base : std_ulogic_vector(31 downto 0); wren : std_logic; rden : std_logic) return acc_t;
```

This should ease changing this logic in the future if you decide upon another related change.

stnolting:
>  I've been a bit busy and I have a few days of extra work ahead, so it'll take some time, though.

Same situation over here, so no worries 😉

> On the other hand, how do you feel about replacing all these:

I like that, but maybe it would be a little bit oversized. The module-internal `addr` signal is something we could remove entirely and just use the lowest bits of the address input signal to distinguish between the different module registers. Thus, only `rden` and `wren` are left.

> If you want to keep the mix of pre-decoding + decentralized full decoding inside the modules, then the address mask idea passed through a generic seems definitely the way to go.

I think I would keep this mix. If we move the decoding entirely to the modules than each modules would have to check the entire 32-bit address signal. By moving the "is IO access" check out of the modules and into into a single, global logic we can save something like ~10 LUTs - per module. No much, but it sums up 😅 

If we add the address mask generic, then how do we specify the number of address bits the module has to check (in an elegant way)? Another generic? Or does the module itself "know" how many bits to check (which it does)?

biosbob:
hi @agamez...  i just met with @stnolting last week, where talked in length about #619....

all of the discussion in this thread is very relevant to my "compact address space" proposal, in which i would like to pack all of the memory-mapped peripheral addresses to the highest possible address....  being able to specify the base address of each peripheral i choose to include helps eliminate holes in this portion of the address space; i'm likewise in favor of each module allocating only what it needs for its registers rather than giving it (say) 1K of space.....

i still believe the `*_size_c` constants are required; but the `*_base_c` constants would now be replaced with some sort of per-peripheral GENERIC....  one still has to be careful that these per-peripheral base addresses don't overlap -- but that's the case today anyway....

specific knowledge of the per-peripheral register offsets should remain inside the peripheral module itself....  and even with just 5-8 peripherals selected in my design, saving ~10 LUTs per module is still nice....  i'm also exploring a much smaller peripheral bus, which would reduce the number of (address) wires actually fed into each peripheral....

@stnolting -- i assume this is the feature request you are working to release ASAP????

finally, there is another (experimental) PR #595 that should make adding new peripherals into a "packed" address space less error prone....  by having a single `.svd` file, we can not only generate a master SW `.h` file but also a single HW `.vhd` file containing all of the required `*_size_c` ***AND*** `*_base_c` constants....  much of this discussion above is really a symptom of not having a **SINGLE** source of truth where an ordered set of system peripherals are described in detail....

but this would be a fairly major change that should NOT stand in the way of compressing the peripheral address space -- even if maintenance is somewhat more awkward in the near-term (eg, keeping SW and HW files in sync)....

agamez:
Hi @biosbob !

Indeed, I saw that discussion some time ago and fount it really interesting, although a bit above my experience.

We've discussed also a lot of things on this thread, but I believe we definitely can take it step by step. I still think this changed I talk about [a few comments above here](https://github.com/stnolting/neorv32/pull/576#issuecomment-1532606704) -but including the address decoding logic, in contrast to what @stnolting suggested- could be useful regardless of where we go from there, because that would allow us to change just a function in order to rewrite the address decoding logic each module needs to do.

If this decoding then goes away from the function and must be added back again to the module, so be it, but passing the generics from the io module down to the function would allow us for easier future changes.

What do you guys think?

stnolting:
I thought about this a lot and I think @agamez is right - a central address decoding unit would make a lot of things easier:
* re-use of the existing peripheral/IO/memory modules
* customizing the address space

I will try to setup a first proposal so we can discuss it here. If you have any thoughts/ideas - just go for it! :wink:

stnolting:
Superseded by #648

PR #585, opened 2023-04-15
Freertos xirq example

matty0005:
As mentioned https://github.com/stnolting/neorv32/discussions/347#discussioncomment-5621552 I have created another freertos example for using xirq's.

I have put it in a seperate folder so that it might be less confusing than having several examples in the one folder - I can merge it if needed though.


stnolting:
This looks great! Thank you very much! :+1:

Let's merge #586 first and then update this branch.

stnolting:
Seems to be ready to merge 👍

matty0005:
I have just fixed up the readme file so that it aligns with the new header guards. 

I also changed the `FREERTOS_HOME` path for consistency as it mentioned version 10.4.1 but the example output mentions 10.4.4 - might want to do the same for the other freertos example.

matty0005:
> Which definition (where?) of FREERTOS_HOME do you mean?

Oops, looks like I was looking at a slightly older verision - was fixed in 65820e0

But this was the line I was referring to (this is the updated path).
https://github.com/stnolting/neorv32/blob/92e4d6fff9e3f26ef48216510df4f4cf4fd020c5/sw/example/demo_freeRTOS_xirq/makefile#L29

Should be all ready now 👍

stnolting:
Thank you very much!

PR #592, opened 2023-04-20
Added a test case to show the 1.0 + -1.0 instruction time-out under f…

mikaelsky:
A specific corner test exist in the floating point add/sub function which results in the post add/sub normalizer running for sreg.exp cycles, which for the number 1.0f is enough to trigger the time-out in the execution engine.
An example has been added to the float_corner_test that performs: 1.0f + -1.0f which results in 0.0f but also times out.
A fix has been added to the normalizer in which the S_PERPARE_NORM state has an additional check whether mantissa_i is all 0's. If so the resulting add/sub is 0.0 and we set ctrl.cnt to 0 to ensure the exp is 0 and jump straight to S_FINALIZE. Which should now correctly generate a +/- 0.0f result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ mimpid = 0x01080200 => Version 01.08.02.00 => v1.8.2
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:-------------------:|:-------:|:--------|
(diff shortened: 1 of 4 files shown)


stnolting:
This is really great!
Thank so you very much for finding (and fixing!) this bug :+1:

stnolting (approved):

PR #595, opened 2023-04-22
generating headers from neorv32.svd

biosbob:
for some background, see #590


biosbob:
there are definitely some "wrong" values still sourced in `neorv32.svd`....  until an exhaustive regression test is run, the generated headers should *NOT* be used in general....

right now, i'm looking for general approval of the approach -- after which i'll take this PR a few steps further....

stnolting:
Hey there!

This looks really promising! Unfortunately, my Python capabilities are quite limited, so it'll take a while for me to understand your script :wink:

I see a general problem here. The current C and header files use Doxygen comments for generating the online software framework documentation at https://stnolting.github.io/neorv32/sw/files.html. How would we generate those comments from the SVD file?

biosbob:
my python skills are also quite limited; i only picked python because i found other `.py` files in the project....  my preferred scripting language is node.js, if that helps....  let me know what you'd prefer....

as for the doxygen comments, it's a straightforward exercise to generate these as well....  the `.svd` file already has `<description>` elements, which can be used as the "source" for the doxygen comments; they're already quite close....

biosbob:
the individual header files would continue to exist in this approach -- they would still need to define the functional APIs implemented in the corresponding C files....

but since these API defintions are **NOT** sourced from the `.svd` file, they would still need to be maintained manually in the headers themselves....

while it is possible to mix "automatically-generated" and "manually-maintained" code in a single file, it's not a best-practice....  the generated `neorv32_svd.h` is a "lower-level" register interface to the raw hardware, while the `neorv32_<peripheral>.h` files are a higher-level functional interface....

stnolting:
> my python skills are also quite limited; i only picked python because i found other .py files in the project.... my preferred scripting language is node.js, if that helps.... let me know what you'd prefer....

I Think I am pretty old-school here. I like bash scripts and good old makefiles 😅 But Python is also just fine.

> as for the doxygen comments, it's a straightforward exercise to generate these as well.... the .svd file already has <description> elements, which can be used as the "source" for the doxygen comments; they're already quite close....

That would be great!

Btw, have you tested the auto-generated header file for completeness? I think some defines might be missing (e.g. pre-defined bit patters). You could also define `NEORV32_STD_HEADER` in the CI pipeline's run script: 

https://github.com/stnolting/neorv32/blob/d0cd94a5c3503d7d86bc200ef4c693d49b6b32ed/do.py#L30

stnolting:
Hey @biosbob!

This PR has been pending for way too long... sorry for that 🙈

How about refreshing this? I think it would be really handy if we could provide a simple script that turns all the header files into a single SVD file - **automatically**.

If the impact on the header files is low (which it seems to be) and if we keep the manually-drafted SVD file for now I would be totally fine with this. ;)

PR #614, opened 2023-05-17
[SW] Expand xirq interrupt function by a parameter.

emb4fun:
Now the interrupt function is expanded by a parameter:

```
void xirq_handler_ch0(void *param) {
  (void)param;
  neorv32_uart0_printf("XIRQ interrupt from channel %i\n", 0);
}

neorv32_xirq_install(0, xirq_handler_ch0, NULL);
```

### DIFF
--- a/sw/example/demo_xirq/main.c
+++ b/sw/example/demo_xirq/main.c
@@ -55,7 +55,8 @@
  * XIRQ handler channel 0.
  * @warning This function has to be of type "void xyz(void)" and must not use any interrupt attributes!
  **************************************************************************/
-void xirq_handler_ch0(void) {
+void xirq_handler_ch0(void *param) {
+  (void)param;
   neorv32_uart0_printf("XIRQ interrupt from channel %i\n", 0);
 }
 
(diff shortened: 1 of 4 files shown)


stnolting:
What's the intended purpose of those additional parameters? Could you provide a (typical?) use case?

If I understand your proposal the right way, then the XIRQ software library add two additional 32*4-byte look up tables for storing a function pointer plus some parameter for each XIRQ channel handler. What's the benefit here?

emb4fun:
The purpose is to use the same interrupt handler for more that one
channels. E.g. if you have connected 2 Ethernet modules, one on channel 0
and the other on channel 1, you can use the same interrupt function with
different Ethernet interface context:

```
neorv32_xirq_install(0, NicInterrupt, netif1);
neorv32_xirq_install(1, NicInterrupt, netif2);
```

The old way can be used too, but hier it looks like:

```
neorv32_xirq_install(0, NicInterrupt1);
neorv32_xirq_install(1, NicInterrupt2);

void NicInterrupt1 (void)
{
   NicInterrupt(netif1);
}

void NicInterrupt2 (void)
{
   NicInterrupt(netif2);
}
```

I do not know what version use less resources. It could be possible
that it makes not sense for this PR in the moment.

emb4fun:
It looks like this doesn't make sense at this point, so I'm closing my pull request.

stnolting:
Feel free to open a new PR if you'd like to continue on this :wink:

PR #639, opened 2023-06-26
demo_blink_led_asm bugfix

vivi202:
This PR fix a bug that make demo_blink_led_asm not work as intended.
# Example of the wrong behavior
bug
This behavior is caused by the delay subroutine which sometimes doesn't work as it should.
The instruction `add  t1, t1, t0` can cause an overflow when mcycle is sufficiently large.
These overflows then leads to the delay_loop being skipped multiple times.
# solution 
I have solved this bug by using the MTIME module instead of mcycle, ensuring that every time the delay subroutine is called, the MTIME.TIME_LOW register is resetted back to zero, thereby avoiding overflows.


vivi202:
I agree with you, MTIME should not be altered, so i went back to the `mcycle` variant.

stnolting:
Looks good to me! :+1:
Were you able to test your new version? Is the overflow issue resolved?

vivi202:
Yes i have tested it, and the issue is now resolved, i also finded a typo in the datasheet i can fix it here or i need to open another PR?

stnolting:
> Yes i have tested it, and the issue is now resolved

Great, thank you!

> i also finded a typo in the datasheet i can fix it here or i need to open another PR?

If that is just a typo then just add it to this PR. 😉

vivi202:
>Great, thank you! 

you'r welcome, thank you so much for making this awesome opensource project!

stnolting:
Thank you very much :wink:

Ready to merge?

vivi202:
> Ready to merge?

Yes.

PR #648, opened 2023-07-15
⚠️ Rework SoC bus system and memory map

stnolting:
## Part 1: Version `1.8.6.3`

* remove files `neorv32_bus_keeper.vhd` and `neorv32_busswitch.vhd` - new file `neorv32_intercon.vhd` now includes all modules for the internal bus system
* add centralized gateway that redirects core accesses to the new 6 main address space regions
  * instruction address space (IMEM)
  * data address space (DMEM)
  * memory-mapped XIP address space
  * bootloader ROM address space
  * IO address space
  * the "void" (unmapped addresses; redirected to the external memory interface if implemented)
* fixed XIP memory page to `0xE` (address range `0xE0000000` to `0xEfffffff`)
* removed IO space PMAs (physical memory attributes); instruction fetches and non-32bit-word accesses to the IO space are now allowed (but not recommended as the behavior might be undefined)

**⚠️ The modifications from this PR are not backwards compatible!**


stnolting:
## Part 2: Version `1.8.6.4`

* remove local addressing decoding from IO/peripheral devices
* add central IO address/access decoding unit (IO switch)
* add uncached i-cache accesses
* rework entire address map

**⚠️ The modifications from this PR are not backward compatible!**

stnolting:
## Part 3: Version `1.8.6.5`

* re-add PMAs (physical memory attributes)

This is what the updated address space looks like:
address_space

**⚠️ The modifications from this PR are not backward compatible!**

biosbob:
i've started playing with this PR, specifically see its impact on my own IceBreaker LUT baseline of 4361....

my first step was to update my own fork to the latest upstream main....  once accomodating the changes, my IceBreaker LUT baseline became 4356 (which is effectively unchanged)....

i then switched this PR, which required me to update the `*mem_ice40up_spram.vhd` along the lines of the `*mem_default.vhd` files....

my IceBreaker now became 4306, which is definitely a measurable improvement!!!!

once i fold this PR into my own compressed address branch (where the latest LUT was ~3760) i'm expecting some further reductions....

all moving in the right direction....

stnolting:
@biosbob

> my IceBreaker now became 4306, which is definitely a measurable improvement!!!!

Great to hear!!! What do you think about the address mapping? Do you think the address map is configurable enough to support your compressed address space proposal?

-------------------------------

This is the new SoC bus architecture:
neorv32_bus

PR #651, opened 2023-07-21
✨ Add support for RISC-V A ISA extension (atomic memory access)

stnolting:
This PR provides a first version of the core that supports the RISC-V `A` ISA extension for atomic memory accesses. The basic idea as well as many of the hardware aspects were triggered by @NikLeberg in https://github.com/stnolting/neorv32/discussions/573 (👍).

ℹ️ This is a quite simple/straightforward approach of implementing LR/SC operations. The handling of those operations is done by the memory/bus system instead of the CPU core.

* Only the _load-store_ (`lr.w`) and _store-conditional_ (`sc.w`) instructions are supported. The remaining read-modify-write instructions (like `amoswap`) are **not** supported and will raise an illegal instruction exception when executed. However, these unsupported instructions can be emulated using LR/SC combinations. An emulation library for the missing AMO operations and an according test program will be provided.
* The `A` extension is enabled via the new `CPU_EXTENSION_RISCV_A` processor generic.
* The `A` instructions are tightly coupled to the memory/bus system of the processor. Besides the CPU-internal logic for executing LR/SC instructions a new bus module ("Reservation Set Control") is added that keeps track of the LR/SC _reservation set_. (see image below).
* Right now only a single reservation set is supported. The _granularity_ (= "size") of the reservation set is 4 bytes (1 word) by default, but can be customized via a new processor generic (`AMO_RVS_GRANULARITY`).

neorv32_bus

## For The Records (port-merge comments)

🔍 How do atomic memory accesses / LR-SC combinations work on a hardware level?

* The RISC-V `A` ISA extension: T.B.A.
* The reservation set controller: T.B.A.
* The LR/SC bus interface protocol: T.B.A.

💾 How to emulate AMO operations using LR/SC combination?

* The AMO emulation library: T.B.B.
* An AMO test/example program: T.B.A.


stnolting:
FTR

A extension implemented, documented and tested ✔️
Further links and notes added to initial PR description ✔️

PR #664, opened 2023-08-04
⚠️ Major code edits / cleanups

stnolting:
* remove co-processor timeout logic from ALU (use the watchdog timer to ensure the CPU does not get stuck instead)
* rename CPU data bus unit: `neorv32_cpu_bus.vhd -> neorv32_cpu_lsu.vhd` ("load/store unit")
* move entire PMP logic (including CSRs) into a new design file `neorv32_cpu_pmp.vhd`
* move FPU CSRs from central control unit to FPU module
* constrain maximum number of HPMs to 13 (`[m]hpmcounter3[h]` to `[m]hpmcounter15[h]`)
* minor RTL edits and cleanups

ℹ️ Moving the entire PMP/FPU CSR logic into optional (`if ... generate`) design units ensures that **all** of the logic is trimmed when the according module is not enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:-------------------:|:-------:|:--------|
+| 04.08.2023 | 1.8.7.5 | :warning: major code edits/cleanups and file renaming; #664 |
 | 29.07.2023 | 1.8.7.4 | RTL cleanup and optimizations (less synthesis warnings, less resource requirements); [#660](https://github.com/stnolting/neorv32/pull/660) |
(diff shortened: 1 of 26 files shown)


NikLeberg:
Hi @stnolting since you are already changing and updating stuff.. some small error just detected: in `neorv32_cpu_amo.c` the doxygen comments for `@param[in] wdata` of all `amomin*` functions is wrong. It says `MAX-ed` instead of `MIN-ed`. Seems like a copy-paste mistake. 😉

I could open a pr but I think that is overkill. How would you like such rather small errors to be fixed? Shall I collect them in a special draft pr where I could collect multiple smaller correction commits? Or am I allowed to directly push such small fixes to main? 🤔

stnolting:
@NikLeberg

> some small error just detected: in neorv32_cpu_amo.c the doxygen comments for @param[in] wdata of all amomin* functions is wrong. It says MAX-ed instead of MIN-ed. Seems like a copy-paste mistake. 😉

Thanks for finding!

> How would you like such rather small errors to be fixed? Shall I collect them in a special draft pr where I could collect multiple smaller correction commits?

Sounds good to me.

> Or am I allowed to directly push such small fixes to main?

I think you cannot push to `main` directly (even as a collaborator). 🤔 I need to check the branch protection rules again...

PR #677, opened 2023-08-25
[OCD] add option to select DM legacy mode

stnolting:
Triggered by @stokdam in https://github.com/stnolting/neorv32/pull/463#issuecomment-1690115103

This PR adds a new generic to the processor's top entity that allows to "downgrade" the on-chip debugger's debug module (DM) back to version 0.13.

* `DM_LEGACY_MODE = false`: DM is compatible to debug spec. version 1.0
* `DM_LEGACY_MODE = true`: DM is compatible to debug spec. version 0.13 (**legacy mode**)

This PR also provides some clean-ups and optimizations for the on-chip debugger modules (DTM & DM).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:-------------------:|:-------:|:--------|
+| 25.08.2023 | 1.8.8.4 | add new generic to downgrade on-chip debugger's debug module back to spec. version 0.13 (`DM_LEGACY_MODE` generic); [#677](https://github.com/stnolting/neorv32/pull/677) |
 | 23.08.2023 | 1.8.8.3 | :test_tube: add experimental `Smcntrpmf` ISA extension (counter privilege mode filtering; spec. is frozen but not yet ratified); remove unused `menvcfg` CSRs; [#676](https://github.com/stnolting/neorv32/pull/676) |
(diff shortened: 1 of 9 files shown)


stnolting:
@stokdam

I think the "legacy" mode is working fine now - at least on my test setup. It would be great if you could test that with your setup, too. 😉

NikLeberg:
Just letting you know: this also works for me. Tested legacy _v0.13_ as well as _v1.0_ and successfully could debug the core.

```
$ openocd --version
> Open On-Chip Debugger 0.11.0-R22.4-gc5bf4f3-dirty (2023-04-25-10:00)
$ riscv32-unknown-elf-gdb --version
> GNU gdb (GDB) 12.1
```

Nice cleanups!

stokdam:
Also works with my setup. However, the error "Error: unable to halt hart 0" I think was due to missing IMEM in configuration. 1.7.9 forced me to enable it (I have not connected any system bus yet), while latest version does not. Running this PR with the same configuration (no IMEM) gives the same error, which disappeared after enabling it.

stnolting:
Thanks for the feedback!

@NikLeberg

> Once this merges I can provide an updated version of the Intel specific DTM in the neorv32-setups repo.

That would be very cool! 😉

@stokdam

> Also works with my setup (v 0.13). However, the error "Error: unable to halt hart 0" I think was due to missing IMEM in configuration. 1.7.9 forced me to enable it (I have not connected any system bus yet), while latest version does not. Running this PR with the same configuration (no IMEM) gives the same error, which disappeared after enabling IMEM.

So the debugger does not work if there is absolutely no memory available (no internal memories and no external ones)? At least the bootloader ROM should be there...
Anyway, good to hear that the setup is working now.

So I think this can be merged, right?

PR #684, opened 2023-09-11
minor rtl edits

stnolting:
- constrain configuration generics (via `range`)
- minor VHDL code cleanups
- update AXI wrapper; add missing configuration options and signals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:-------------------:|:-------:|:--------|
+| 11.09.2023 | 1.8.9.1 | minor rtl edits and updates; [#684](https://github.com/stnolting/neorv32/pull/684) |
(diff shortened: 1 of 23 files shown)


NikLeberg:
Since you are already updating the dm: I have noticed during my multi-hart extension that the member `progbuf` in the `dci_t` record is not used anywhere. Seems to be a leftover of a refactoring.

```
index 6622cc16..e738a624 100644
--- a/rtl/core/neorv32_debug_dm.vhd
+++ b/rtl/core/neorv32_debug_dm.vhd
@@ -215,7 +215,6 @@ architecture neorv32_debug_dm_rtl of neorv32_debug_dm is
     execute_req   : std_ulogic; -- DM wants CPU to execute program buffer when set
     execute_ack   : std_ulogic; -- CPU starts executing program buffer when set (single-shot)
     exception_ack : std_ulogic; -- CPU has detected an exception (single-shot)
-    progbuf       : std_ulogic_vector(255 downto 0); -- program buffer, 4 32-bit entries
     data_we       : std_ulogic; -- write abstract data
     data_reg      : std_ulogic_vector(31 downto 0); -- memory-mapped data exchange register
   end record;
```

stnolting:
You are right! Thanks for noticing!

PR #698, opened 2023-10-06
[SW] Illegal instruction removed

emb4fun:
The CoreMark is used in Machine-Mode, therefore the access to CSR_MCOUNTEREN can
be removed. The access to CSR_MCOUNTEREN is for User-Mode.


stnolting (approved):

stnolting:
Thanks for fixing!

PR #701, opened 2023-10-12
Include explanation of how to send raw chars in `cutecom`

lianakoleva:
Good to include as `cutecom` is the recommended Linux terminal program

### DIFF
--- a/ug/index.html
+++ b/ug/index.html
@@ -1475,8 +1475,9 @@ following parameters:</p>
 <ol class="arabic" start="4">
 <li>
 <p>Also make sure that single chars are send from your computer <em>without</em> any consecutive "new line" or "carriage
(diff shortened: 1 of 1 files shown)


stnolting:
Good point! Thanks for the update! 👍

But please update the documentation in the according **markdown source file** in the `main` branch:
https://github.com/stnolting/neorv32/blob/f4438f06f87d83670201c89ca7cf3f5a2a61667e/docs/userguide/executable_upload.adoc?plain=1#L40-L42

The html file in the `gh-pages` are auto-generated from the source files by the GitHub actions. These files are overridden everytime the documentation is updated, so your changes there will be lost after the next run.

lianakoleva:
Got it, my bad! I created a new PR (#702) as I needed to use a different source branch.

PR #717, opened 2023-10-30
Fix bug in neorv32_slink_available() function

Unike267:
# Fix a bug in *neorv32_slink.c* library

In specific in *neorv32_slink_available()* function.

### Context:

I added a [module](https://github.com/Unike267/Practices) to NEORV32 via Stream Link interface (AXI4-Stream) and I used in *.c* program *neorv32_slink_available()* function to check if SLINK module was available. The point is that I put the *IO_SLINK_EN => true* in the top of the design and when I checked the results via UART the program crashed because *neorv32_slink_available()* return -1 when the SLINK module was available. 

error

I checked the *neorv32_slink.c* and **I found this bug**. 

When I remade the *application_image* compiled with the fixed-up library  in local everything was good. I got the following results with CuteCom terminal:

/cc @umarcor

### DIFF
--- a/sw/lib/source/neorv32_slink.c
+++ b/sw/lib/source/neorv32_slink.c
@@ -48,7 +48,7 @@
  **************************************************************************/
 int neorv32_slink_available(void) {
 
-  if (NEORV32_SYSINFO->SOC & (1 << SYSINFO_SOC_IO_SDI)) {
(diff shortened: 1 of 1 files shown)


stnolting:
Hey @Unike267!

Looks like a classical copy-and-paste error... 🙈
Anyway, thanks for finding and fixing this bug! :+1:

PR #722, opened 2023-11-07
Fix-up the litex wrapper

Unike267:
# Fix-up Litex wrapper

+ Board: ARTY A7 35T

### Context:

I'm trying to implement the Neorv32 with Litex and with the deffault wrapper **there is a failure with the SDRAM initialization**. View enjoy-digital/litex#1802.

SDRAM_ERROR

I tried changing the wishbone protocol from *pipelined* to *classic*, and that seems to fix the issue.
It might have some effect in performance, but at least it works.
The following log shows that memory is correctly initialized.

SDRAM_OK

Looking forward, I believe there might be some conflict in the pipelined wishbone implementations of NEORV32 and Litex. @stnolting did you try the pipelined wishbone of NEORV32 with third-party wishbone peripherals?

/cc @enjoy-digital @umarcor


enjoy-digital:
Thanks @Unike267, I was just going to also look at it. Good to know you figured out the issue. It would probably be interesting to merge this PR and then see if we can enable burst with NeoRV32 and LiteX integration.

PR #727, opened 2023-11-16
Fix comment mistake

Unike267:
Hi @stnolting!

There is a little mistake in this comment. According to definition of [custom instruction R4 type](https://github.com/stnolting/neorv32/blob/main/sw/lib/include/neorv32_intrinsics.h#L224-L249) the bit-field is funct3 not funct7. 

Regarts!


stnolting (approved):

stnolting:
Thanks for fixing! 👍

Issue #728, opened 2023-11-16
Clarification on Zfinx behavior for subnormal operands and results needed


jstraus59:
The [NEORV32 documentation Zfinx extension documentation](https://stnolting.github.io/neorv32/#_zfinx_isa_extension)  states:

`Subnormal numbers (exponent = 0) are flushed to zero setting them to +/- 0 before entering the FPU’s processing core. If a computational instruction (like fmul.s) generates a subnormal result, the result is also flushed to zero during normalization.`

Would it be possible to be add more detail on subnormal flushing, specifically:
- How are the fflag bits set when subnormal inputs are seen, or when subnormal results are flushed to zero. A precise specification for each instruction class (Computational, Conversion, Compare and Clasify) would be helpful.
- For input flushing, it simply states *flushed...before entering the FPU’s processing core*, but exactly which instruction classes are implemented in the FPU core, and thus input flushing applies to them, is not clear. Obviously, this applies to the Computational instruction class, but an explicit specification of whether this also applies to the Conversion, Compare and Clasify instruction classes would be helpful.

Thanks!

Jim Straus
Imperas

mikaelsky:
@jstraus59 from my reading of the code the first step in the FPU is to determine the "class" of a given float to extract +/- inf, NAN etc. before passing them further into the FPU. During this intake process if an number with exp 0 is seen the mantissa just gets set to 0, resulting in any subnormal being converted to a 0. E.g a subnormal A + subnormal B gets converted to 0.0 + 0.0.
As this a pre-processing step it happens disregarding instruction class. I'll do a bit more digging but thats the gist of it.

Currently the fflag bits are .. not functional in general which include subnormals. This is on my agenda to fix, so assume that fflags will be functional going forward.

jstraus59:
@mikaelsky Terrific - thanks for the info. I will assume that inputs to the Conversion and Computation instructions have their subnormal inputs flushed to 0.

I still need a couple details clarified, please (sorry but I have never done RTL design - just modeling of processor architectures - so it would be difficult for me to figure this out directly from the RTL myself):

1) I am wondering still about the FCLASS instruction, which simply reports the type of floating-point value that is in a register. One of the things it detects is whether the value is a subnormal, so it would not really make sense for subnormal inputs to be flushed for the FCLASS instruction, but if FCLASS is implemented within the FPU it sounds like they would. Could you please confirm?

[...]

PR #734, opened 2023-11-21
🐛 fix bug in instruction-misaligned exception handling

stnolting:
Bug identified by @mikaelsky in https://github.com/stnolting/neorv32/issues/728#issuecomment-1819638160.

The current processor version captures unaligned instruction addresses in the pipeline's front-end. Hence, any misaligned instruction will still trigger an instruction fetch and jump-and-link instruction will still update the link register. This PR aims to fix this:

* Branch instructions (unconditional branches or "taken" conditional branches) do not trigger an instruction fetch if the address is misaligned.
* Jump-and-link instruction do not write to the link register if the destination address is misaligned.
* Misaligned instruction address exception are raised by the pipeline's back end (execution) and not by the pipeline's front-end (instruction fetch).

I am not quite sure about the handling of `xRET` instructions when `xEPC` contains a misaligned address. The RISC-V priv. spec states:

> Machine Exception Program Counter (mepc)
> [...]
> If an implementation allows IALIGN to be either 16 or 32 (by changing CSR misa, for example),
> then, whenever IALIGN=32, bit mepc[1] is masked on reads so that it appears to be 0. This
> masking occurs also for the implicit read by the MRET instruction. Though masked, mepc[1]
> remains writable when IALIGN=32.

So I think `xEPC[1]` should be hardwired to zero if IALIGN=32 (i.e. the `C` extension is NOT enabled). Hence, `xRET` cannot raise a misaligned instruction exception. However, I cannot find anything regarding `dret`/`dpc` but I assume that we should expect the same behavior there. @mikaelsky can you confirm this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date (*dd.mm.yyyy*) | Version | Comment |
 |:-------------------:|:-------:|:--------|
+| 21.11.2023 | 1.9.1.4 | :bug: fix bug in handling of "misaligned instruction exception"; [#734](https://github.com/stnolting/neorv32/pull/734) |
 | 20.11.2023 | 1.9.1.3 | :bug: fix wiring of FPU exception flags; [#733](https://github.com/stnolting/neorv32/pull/733) |
 | 18.11.2023 | 1.9.1.2 | add XIP clock divider to fine-tune SPI frequency; [#731](https://github.com/stnolting/neorv32/pull/731) |
 | 18.11.2023 | 1.9.1.1 | (re-)add SPI high-speed mode, :bug: fix bug in SPI shift register - introduced in v1.9.0.9; [#730](https://github.com/stnolting/neorv32/pull/730) |
--- a/do.py
+++ b/do.py
@@ -26,7 +26,7 @@ def task_BuildAndInstallSoftwareFrameworkTests():
             "make -C sw/bootloader clean_all info bootloader",
             # Compile and install test application, redirect UART0 TX to text.io simulation output via <UARTx_SIM_MODE> user flags
             "echo 'Compiling and installing CPU/Processor test application'",
-            "make -C sw/example/processor_check clean_all USER_FLAGS+=-DUART0_SIM_MODE USER_FLAGS+=-DUART1_SIM_MODE USER_FLAGS+=-flto EFFORT=-Os MARCH=rv32imac_zicsr_zifencei info all",
+            "make -C sw/example/processor_check clean_all USER_FLAGS+=-DUART0_SIM_MODE USER_FLAGS+=-DUART1_SIM_MODE USER_FLAGS+=-flto EFFORT=-Os MARCH=rv32ima_zicsr_zifencei info all",
         ],
         "doc": "Build all sw/example/*; install bootloader and processor check",
     }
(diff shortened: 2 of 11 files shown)


mikaelsky:
@stnolting so reading the spec there really isn't an "xRET" instruction, from what I can gather it seems its just short-hand for JAL.
"RISC-V does not have a dedicated instruction for return. The ret is a pseudo-instruction that is expanded to jalr zero, 0(ra), which sets the program counter to ra + 0 and saves the previous program counter’s value plus four to register zero, which is hardwired to zero. This implies that the return value must be copied from the stack into ra before returning." 
source: https://infosecwriteups.com/return-oriented-programming-on-risc-v-part-1-dd9817b52d2b

Because of this fixing JAL to handle mis-aligned targets should also fix "xRET". So we shouldn't need to tie xEPC[1] to 0.
For MRET/SRET its an open question what happens as one could imagine xEPC having been modified. Its not clear from the spec vol II. Let me check it and I'll respond.

An additional "fun" detail with JALR is: Spec 2.2 chapter 2.5.1 unconditional branches.
"The indirect jump instruction JALR (jump and link register) uses the I-type encoding. The target
address is obtained by adding the sign-extended 12-bit I-immediate to the register rs1, then setting the
least-significant bit of the result to zero."
This basically makes JALR behave like JAL, as JAL has the LSB implicitly set to 0.

This explains this note for JAL/JALR, as the PC LSB is forced to be 0 disregarding its content.
"The JAL and JALR instructions will generate an instruction-address-misaligned exception if the target
address is not aligned to a four-byte boundary.
Note: Instruction-address-misaligned exceptions are not possible on machines that support
extensions with 16-bit aligned instructions, such as the compressed instruction-set
extension, C."

After thanksgiving I'll take a stab at the v1.9.1.4 release in a local branch and see if it passes our compliance check. Our checker is much more punishing than the riscv-arch test suite.

stnolting:
> so reading the spec there really isn't an "xRET" instruction, from what I can gather it seems its just short-hand for JAL.
"RISC-V does not have a dedicated instruction for return. 

Right, I was just referring to the "special" `mret` and `dret` instructions here.

> So we shouldn't need to tie xEPC[1] to 0.
For MRET/SRET its an open question what happens as one could imagine xEPC having been modified. Its not clear from the spec vol II. Let me check it and I'll respond.

Well, this is what I have found in the specs.:

### Priv. spec regarding `mret` / `mepc`

> If an implementation allows IALIGN to be either 16 or 32 (by changing CSR misa, for example),
then, whenever IALIGN=32, bit mepc[1] is masked on reads so that it appears to be 0. This
masking occurs also for the implicit read by the MRET instruction. Though masked, mepc[1]
remains writable when IALIGN=32.

### Debug spec regarding `dret` / `dpc`

> The writability of dpc follows the same rules as mepc as defined in the Privileged Spec. In particular, dpc must be able to hold all valid virtual addresses and the writability of the low bits depends on IALIGN.

So I think it should be fine to hardwire bit 1 to zero if IALIGN = 32 (i.e. `C` is disabled).

> An additional "fun" detail with JALR is: Spec 2.2 chapter 2.5.1 unconditional branches. [...]

I need to check that again. I read somewhere that a lot of RISC-V core have some issues with a proper implementation of this...

> After thanksgiving I'll take a stab at the v1.9.1.4 release in a local branch and see if it passes our compliance check. Our checker is much more punishing than the riscv-arch test suite.

Thank you very much and happy Thanksgiving! :wink:

mikaelsky:
More a note on the branch misalign fix added in v1.9.1.4.
I've finally gotten to merge up to 1.9.3 and worked through forward fixing the RVVI trace port to be functional with core version 1.9.3. 
Initial results show that the misaligned branch fix is functional vs the imperas reference model - and thus the riscv spec - on a per instruction compare full processor state compare basis.

PR #762, opened 2024-01-10
[rtl/core] add again mtime_o to top entity

mcoroyer:
This PR adds again `mtime_o` to the top entity to retain the ability to retrieve the wall clock time (mtime) from the HDL gateware. (#761)


stnolting:
Thanks for your pull request!
I'll have a look at this (maybe editing some stuff if that's OK) and I'll also take of the versioning and documentation stuff.

stnolting (approved):

PR #766, opened 2024-01-17
Added tweaks to the FIFO to prevent assertion errors.

mikaelsky:
The FIFO generates two types of assertion errors:
- X generation from the async read section
- NULL warnings from the sync read and write sections when fifo_depth_c is set to 1.

The subtle fixes will prevent X generation for async read by forcing a mux on the read output during reset low periods.
The NULL fix is simple detecting fifo_depth_c as being 1 and setting the fifo_mem index accordingly. The read/write point math ends up as a 0 downto 0 index which results in a NULL assertion warning in xcelium.

### DIFF
--- a/rtl/core/neorv32_fifo.vhd
+++ b/rtl/core/neorv32_fifo.vhd
@@ -157,7 +157,13 @@ begin
     begin
       if rising_edge(clk_i) then
         if (we = '1') then
+        -- Added fifo_depth check to the read out path to preven a NULL assertion for
+        -- fifo_depth_c of 1.
+        if (fifo_depth_c > 1) then 
           fifo_mem(to_integer(unsigned(w_pnt(w_pnt'left-1 downto 0)))) <= wdata_i;
+        else
+          fifo_mem(0) <= wdata_i;
+        end if;
         end if;
       end if;
     end process fifo_write;
@@ -168,7 +174,9 @@ begin
   -- -------------------------------------------------------------------------------------------
   fifo_read_async: -- asynchronous read
   if not FIFO_RSYNC generate
-    rdata_o <= fifo_mem(to_integer(unsigned(r_pnt(r_pnt'left-1 downto 0))));
+    -- Added a reset driven bypass to ensure that rdata_o is defined during reset
+    -- This prevents X generation while fifo_mem and r_pnt are still being initialized.
+    rdata_o <= fifo_mem(to_integer(unsigned(r_pnt(r_pnt'left-1 downto 0)))) when (rstn_i = '1') else (others => '0');
     -- status --
     free_o  <= free;
     avail_o <= avail;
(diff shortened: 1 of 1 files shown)


stnolting:
Thanks for your modifications! I'll have a look at this.

rtl/core/neorv32_fifo.vhd
```diff
@@ -157,7 +157,13 @@ begin
     begin
       if rising_edge(clk_i) then
         if (we = '1') then
+        -- Added fifo_depth check to the read out path to preven a NULL assertion for
+        -- fifo_depth_c of 1.
+        if (fifo_depth_c > 1) then
```
stnolting, line 162: Looks good! But shouldn't we add this to the FULL_RESET write process above as well?

rtl/core/neorv32_fifo.vhd
```diff
@@ -168,7 +174,9 @@ begin
   -- -------------------------------------------------------------------------------------------
   fifo_read_async: -- asynchronous read
   if not FIFO_RSYNC generate
-    rdata_o <= fifo_mem(to_integer(unsigned(r_pnt(r_pnt'left-1 downto 0))));
+    -- Added a reset driven bypass to ensure that rdata_o is defined during reset
+    -- This prevents X generation while fifo_mem and r_pnt are still being initialized.
+    rdata_o <= fifo_mem(to_integer(unsigned(r_pnt(r_pnt'left-1 downto 0)))) when (rstn_i = '1') else (others => '0');
```
stnolting, line 179: I think we should `fifo_depth_c` as well:
```vhdl
rdata_o <= fifo_mem(to_integer(unsigned(r_pnt(r_pnt'left-1 downto 0)))) when (fifo_depth_c > 1) else fifo_mem(0);
```

stnolting: I do not like using `rstn_i` as direct input here 😅 The reset should be a global net that is only driving the dedicated reset inputs of FFs. Furthermore, the reset is asynchronous, so we might get trouble with timing analysis here.

The when-else statement adds another layer of logic without "being relevant for the function" so I would like to avoid this to keep the hardware footprint (and the critical path) at a minimum. As far as I understand the X-issue here comes from `fifo_mem` being uninitialized at the beginning, right?

rtl/core/neorv32_fifo.vhd
mikaelsky: Doh. Good catch, yes we should. This would be a copy paste error on my side as I was forward writing from an older version of the code base.
Want me to fix it?

rtl/core/neorv32_fifo.vhd
mikaelsky: Agreed.

rtl/core/neorv32_fifo.vhd
stnolting: That would be great! :wink:

rtl/core/neorv32_fifo.vhd
stnolting: But how can we fix the X-issue at this point?

On Intel FPGAs the entire "asynchronous" FIFO read logic gets absorbed into the primitive memory bits. I am not sure how this is done (need to investigate on that). Maybe the logic to set the output to zero on reset could also be absorbed as most BRAM have output registers with a dedicated reset signal. I need to dig into the technology data sheets...

Anyway, with `FULL_RESET => true,` this X-issue should be gone, right?

stnolting:
I really like the proposals you made in this PR. Should I add those to one of my next "minor edits" PRs or are you working on a revised PR? :wink:

mikaelsky:
Just added a new PR, couldn't reopen this one as to much time had passed :(
The slight delay is more me have some 70 failing regression tests for the core I also need to debug, granted 60 or so the zfinx compliance tests failing because of fflags.

stnolting:
> Just added a new PR [...]

But not in _this_ repository, right? Maybe in a fork?! 🤔

mikaelsky:
Indeed in my fork. PR  #778 submitted.

PR #778, opened 2024-01-28
Updated FIFO NULL assertion fix

mikaelsky:
fifo.vhd has been updated to reflect feedback. Changes has been verified to not cause issues with assertions.
Added a tweak to the write path to make read and write symmetric allowing some flops to potentially be saved for FIFOs of size 1.


stnolting:
Looks good!
I'll just update the comments and the version tracking :wink:

stnolting (approved):

Issue #785, opened 2024-02-01
FPU fflags no being asserted correctly


mikaelsky:
**Describe the bug**
This is a known issue that I finally have gotten around to. The debugging is far from complete, so would suggest I add onto this this issue with examples and fixes before putting in a PR. Alternative I can toss a PR for each, but might be a lot of back-n-forth.

bug #1: When an f.add is being issued and one of the operands is +/- 0.0 and the other operands exponent is large enough to trigger the exception path the FPU will erroneously set the NX flag. This bug was caused by an earlier bug fix from me that prevent the addsub from running forever for large exponent differences.
```
Info (IDV) Instruction executed prior to mismatch '0x304(inst_18+10): 018df8d3 fadd.s  x17,x27,x24'
Error (IDV) CSR register value mismatch (HartId:0, PC:0x00000304 inst_18+10):
Error (IDV) Mismatch 0> CSR 001 (fflags)
Error (IDV)   . dut:0x00000001 NV:0 DZ:0 OF:0 UF:0 NX:1
Error (IDV)   . ref:0x00000000 NV:0 DZ:0 OF:0 UF:0 NX:0
Error (IDV) Mismatch 1> CSR 003 (fcsr)
Error (IDV)   . dut:0x00000001 NV:0 DZ:0 OF:0 UF:0 NX:1 FRM:0(RNE)
Error (IDV)   . ref:0x00000000 NV:0 DZ:0 OF:0 UF:0 NX:0 FRM:0(RNE) (not updated)
Error (IDV) tb.riscv_tb.idv_trace2api.state_compare @ 37380.00 ns: MISMATCH
```
---------

**To Reproduce**
This is the instruction test from our converted F -> Zfinx riscof-arch test. 
Note gcc's asm out doesn't reproduce zfinx disassembly correctly, so the above Imperas DV output for the real assembly command.
Note when you see "sky_" in front of a riscof-arch test function name is calling an updated test function that replaces F# registers with X# registers in a manner that is compliant with riscof-arch.

```
000002f4 <inst_18>:
// rs1==f27, rs2==f24, rd==f17,fs1 == 0 and fe1 == 0x00 and fm1 == 0x000000 and fs2 == 1 and fe2 == 0x7f and fm2 == 0x000000 and  fcsr == 0x0 and rm_val == 7   
/* opcode: fadd.s ; op1:f27; op2:f24; dest:f17; op1val:0x0; op2val:0xbf800000; 
[...]

PR #788, opened 2024-02-02
Fix for issue #785: FPU fflags no being asserted correctly

mikaelsky:
Patched up my fixes to the FPU to the current mainline from my local 1.9.3.0 checkout.

There are still challenges but these fixes makes the following riscof-arch F -> Zfinx tests pass:
rv32i_m_Zfinx_fadd_b1
rv32i_m_Zfinx_fadd_b10
rv32i_m_Zfinx_fadd_b11
rv32i_m_Zfinx_fadd_b12
rv32i_m_Zfinx_fadd_b13
rv32i_m_Zfinx_fadd_b2
rv32i_m_Zfinx_fadd_b3
rv32i_m_Zfinx_fadd_b4
rv32i_m_Zfinx_fadd_b5
rv32i_m_Zfinx_fadd_b7
rv32i_m_Zfinx_fadd_b8
rv32i_m_Zfinx_fclass_b1
rv32i_m_Zfinx_fcvt.s.w_b25
rv32i_m_Zfinx_fcvt.s.w_b26
rv32i_m_Zfinx_fcvt.s.wu_b25
rv32i_m_Zfinx_fcvt.s.wu_b26
rv32i_m_Zfinx_fcvt.w.s_b1
rv32i_m_Zfinx_fcvt.w.s_b22
rv32i_m_Zfinx_fcvt.w.s_b23
rv32i_m_Zfinx_fcvt.w.s_b24
rv32i_m_Zfinx_fcvt.w.s_b27
rv32i_m_Zfinx_fcvt.w.s_b28
rv32i_m_Zfinx_fcvt.w.s_b29
rv32i_m_Zfinx_fcvt.wu.s_b1
rv32i_m_Zfinx_fcvt.wu.s_b22
rv32i_m_Zfinx_fcvt.wu.s_b23
rv32i_m_Zfinx_fcvt.wu.s_b24
rv32i_m_Zfinx_fcvt.wu.s_b27
rv32i_m_Zfinx_fcvt.wu.s_b28
rv32i_m_Zfinx_fcvt.wu.s_b29
rv32i_m_Zfinx_fsub_b1
rv32i_m_Zfinx_fsub_b11
rv32i_m_Zfinx_fsub_b12
rv32i_m_Zfinx_fsub_b13
rv32i_m_Zfinx_fsub_b2
rv32i_m_Zfinx_fsub_b3
rv32i_m_Zfinx_fsub_b4
rv32i_m_Zfinx_fsub_b5
rv32i_m_Zfinx_fsub_b7
rv32i_m_Zfinx_fsub_b8

Some test have.. challenges.. with the F->Zfinx conversion as riscof-arch-test utilize sub-normals and the default Sail model supports sub-normals. All the test pass against a tuned RISCV architectural model where sub-normals are replaced with 0.0 as happens with the core.

### DIFF
--- a/rtl/core/neorv32_cpu_cp_fpu.vhd
+++ b/rtl/core/neorv32_cpu_cp_fpu.vhd
@@ -923,27 +923,82 @@ begin
         if (fpu_operands.rs1(31) = fpu_operands.rs2(31)) then -- identical signs
           addsub.res_sign <= fpu_operands.rs1(31);
         else -- different signs
-          if (addsub.exp_comp(1) = '1') then -- exp are equal (also check relation of mantissas)
-            addsub.res_sign <= fpu_operands.rs1(31) xor (not addsub.man_comp);
-          else
-            addsub.res_sign <= fpu_operands.rs1(31) xor addsub.exp_comp(0);
+          -- if the result is not 0.0 set the sign normally
+          if ((to_integer(unsigned(addsub.add_stage))) /= 0 ) then
+            if (addsub.exp_comp(1) = '1') then -- exp are equal (also check relation of mantissas)
+              addsub.res_sign <= fpu_operands.rs1(31) xor (not addsub.man_comp);
+            else
+              addsub.res_sign <= fpu_operands.rs1(31) xor addsub.exp_comp(0);
+            end if;
+          else 
+            --  roundTowardNegative; under that attribute, the sign of an exact zero sum (or difference) shall be −0
+            if (fpu_operands.frm = "010") then -- round down (towards -infinity)
+              addsub.res_sign <= '1'; -- set the sign to 0 to generate a +0.0 result
+            else
+              addsub.res_sign <= '0'; -- set the sign to 0 to generate a +0.0 result
+            end if;
           end if;
         end if;
       else -- sub
-        if (fpu_operands.rs1(31) = fpu_operands.rs2(31)) then -- identical signs
-          if (addsub.exp_comp(1) = '1') then -- exp are equal (also check relation of mantissas)
-            addsub.res_sign <= fpu_operands.rs1(31) xor (not addsub.man_comp);
-          else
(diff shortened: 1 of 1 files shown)


stnolting:
Thank you so much for your patch and the detailed description in #785!

stnolting (approved):

mikaelsky:
Its been on list to get to these for a Long time, finally get some time to get deep dive into the FPU :)
I do have another patch coming with a lot of different fixes all over the place. For this one I fixed fclass.s to correctly report out subnormals, but there are side effects so have added a generic "FPU_SUBNORMAL_SUPPORT" that defaults to false.
I opted for this path as you already put in a lot of subnormal checks which will need to by bypassed when we flush subnormals. Instead of removing the code I opted to wrap it in an if().

stnolting:
Again, thank you so much for all your work!!

> Instead of removing the code I opted to wrap it in an if().

The question is, do we really want to support subnormal numbers one day? I'm not sure about this. 🤔

mikaelsky:
Its a good question. In principle the RISCV float spec requires sub-normal support. For now the core FPU extension doesn't fully support subnormals, even though I see some conditional checks here and there that seem correct.

Ideally we could keep as a future action to add support for subnormals (hence the generic) with the caveat that it will increase area. This is basically kicking the can down the road to some future point.

Right now the FPU patches are mostly aimed at getting to correct functionality btw. I'm sure some of them could benefit from some refactoring as well. Especially the exception paths. If we decide to just not support subnormals outside fclass and fsgn style operations then we can always just refactor out the conditional path.

Currently hammering away at the multiplier core exception paths.. there are quite a few :)

stnolting:
> Ideally we could keep as a future action to add support for subnormals (hence the generic) with the caveat that it will increase area. This is basically kicking the can down the road to some future point.

Sounds reasonable. :+1:

Issue #816, opened 2024-02-20
Simulation hangs


davidgussler:
**Describe the bug**
The hello_world example program simulation hangs & times out after 10ms.

**To Reproduce**
1. Compile and install latest ghdl
2. Compile and install latest risc-v gcc toolchain
3. cd neorv32/sw/example/hello_world
4. make USER_FLAGS+=-DUART0_SIM_MODE MARCH=rv32imc_zicsr clean_all sim  # Note: User guide section 13.4.1 does NOT specify that "_zicsr" extension needs to be used for the compilation, but I had to add it for the hello_world to compile. I'm unsure if this is an issue on my end or a problem with the documentation.

Console output:
$ make USER_FLAGS+=-DUART0_SIM_MODE MARCH=rv32imc_zicsr clean_all sim
../../../sw/lib/source/neorv32_uart.c: In function 'neorv32_uart_setup':
../../../sw/lib/source/neorv32_uart.c:116:2: warning: #warning UART0_SIM_MODE (primary UART) enabled! Sending all UART0.TX data to text.io simulation output instead of real UART0 transmitter. Use this for simulations only! [-Wcpp]
  116 | #warning UART0_SIM_MODE (primary UART) enabled! Sending all UART0.TX data to text.io simulation output instead of real UART0 transmitter. Use this for simulations only!
      |  ^~~~~~~
Memory utilization:
   text	   data	    bss	    dec	    hex	filename
   3760	      0	    116	   3876	    f24	main.elf
Compiling ../../../sw/image_gen/image_gen
Installing application image to ../../../rtl/core/neorv32_application_image.vhd
Simulating neorv32_application_image.vhd...
Tip: Compile application with USER_FLAGS+=-DUART[0/1]_SIM_MODE to auto-enable UART[0/1]'s simulation mode (redirect UART output to simulator console).
analyze ../../rtl/core/neorv32_package.vhd
analyze ../../rtl/core/neorv32_application_image.vhd
analyze uart_rx.simple.vhd
analyze neorv32_tb.simple.vhd
analyze ../../rtl/core/neorv32_clockgate.vhd
analyze ../../rtl/core/neorv32_fifo.vhd
analyze ../../rtl/core/neorv32_cpu_decompressor.vhd
analyze ../../rtl/core/neorv32_cpu_control.vhd
analyze ../../rtl/core/neorv32_cpu_regfile.vhd
analyze ../../rtl/core/neorv32_cpu_cp_shifter.vhd
[...]

PR #817, opened 2024-02-20
[sim] add simulation check to sw makefiles as target 'sim-check'

umarcor:
Coming from #816, this PR moves the cat-grep assert from the simulation shell script to the Makefile of the corresponding software program. In order to do so, target `sim-check` is added, which depends on `sim`.
This approach allows us to have per-program checks. In order to showcase the feature, I added a check to `hello_world` as well.

See a successful execution https://github.com/umarcor/neorv32/actions/runs/7977125739 and failures when I changed the strings in the software on purpose: https://github.com/umarcor/neorv32/actions/runs/7977200433.


stnolting:
That's a really nice solution! :+1:

stnolting (approved):

PR #819, opened 2024-02-20
[docs/userguide/simulating_the_processor] add admonition and recommend MARCH=rv32im to build hello_world

umarcor:
As discussed in #816, this PR brings a couple of enhancements to the Simulation section of the User Guide.

On the one hand, an admonition is added for users to know that using instructions corresponding to unsupported/disabled extensions will make the processor hung and the simulation will timeout.

On the other hand, the recommended MARCH to build hello_world is changed not to use extension C, since that's disabled in the simple testbench.

### DIFF
--- a/docs/userguide/simulating_the_processor.adoc
+++ b/docs/userguide/simulating_the_processor.adoc
@@ -154,7 +154,7 @@ Blinking LED demo program
 
 To do a quick test of the NEORV32 make sure to have https://github.com/ghdl/ghdl[GHDL] and a
 https://github.com/stnolting/riscv-gcc-prebuilt[RISC-V gcc toolchain] installed.
-Navigate to the project's `sw/example/hello_world` folder and run `make USER_FLAGS+=-DUART0_SIM_MODE MARCH=rv32im clean_all sim`:
+Navigate to the project's `sw/example/hello_world` folder and run `make USER_FLAGS+=-DUART0_SIM_MODE clean_all sim`:
 
 [TIP]
 The simulator will output some _sanity check_ notes (and warnings or even errors if something is ill-configured)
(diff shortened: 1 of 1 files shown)


docs/userguide/simulating_the_processor.adoc
```diff
@@ -149,15 +154,15 @@ Blinking LED demo program
 
 To do a quick test of the NEORV32 make sure to have https://github.com/ghdl/ghdl[GHDL] and a
 https://github.com/stnolting/riscv-gcc-prebuilt[RISC-V gcc toolchain] installed.
-Navigate to the project's `sw/example/hello_world` folder and run `make USER_FLAGS+=-DUART0_SIM_MODE MARCH=rv32imc clean_all sim`:
+Navigate to the project's `sw/example/hello_world` folder and run `make USER_FLAGS+=-DUART0_SIM_MODE MARCH=rv32im clean_all sim`:
```
stnolting, line 157: I would just remove `MARCH=...` from the command so that the (inexperienced) user always uses the default (=minimal) `MARCH` configuration from the makefile.

docs/userguide/simulating_the_processor.adoc
```diff
@@ -30,6 +30,11 @@ A plain-VHDL (no third-party libraries) testbench (`sim/simple/neorv32_tb.simple
 testing the processor.
 This testbench features a 100MHz clock and enables all optional peripheral and CPU extensions except for the `E`.
 
+[IMPORTANT]
+In the simple testbench several optional extensions are disabled, such as C or E.
+If software is compiled using instructions corresponding to disabled extensions, the whole processor will hang in an eternal exception loop and, therefore, the simulation will timeout.
+The `MARCH` must be a subset of the extensions enabled in the testbench.
+
```
stnolting, line 37: :+1:

docs/userguide/simulating_the_processor.adoc
umarcor: Removed!

PR #820, opened 2024-02-20
[ci] split SoftwareFrameworkTests from simple testbench simulation

umarcor:
This PR is based on #817, so I'll keep it as a draft until that one is merged.

Instead of building the processor check software together with other "toolchain testing software compilation" tasks, it is done at the same time the simulation is to be executed, in a separated job.
Moreover, since building the software and running the simulation was done using the same container image but in different steps (thus, spawning two containers), I changed it to call make target `sim-check` (from #817) directly.

By the way, I removed 'Container' from job names/ids, since currently all the steps are executed on containers. I believe that's just a leftover from some years ago, when the simple testbench was executed on GitHub's Ubuntu environment.


umarcor:
This is now rebased, after merging #817.

stnolting (approved):

Issue #753, opened 2023-12-18
neorv32_fifo vivado implementation issue


doud3157:
**Describe the bug**
With vivado 2023.1, the implementation of neorv32_fifo is incorrect. It uses CLB registers instead of BRAM.

**To Reproduce**
Synthesize the neorv32_fifo in standalone with a depth of 1024 bytes (for example)

**Expected behavior**
The neorv32_fifo should use BRAM and not CLB registers when DEPTH > 1.

**Screenshots**
NA

**Environment:**
 - Windows 11
 - vivado 2023.1

**Hardware:**
 - Zynq ultrascale+

**Additional context**
The problem is due to the usage of the record fifo_t. The parameter data (type fifo_data_t) should be define outside of the type fifo_t to help vivado. The processes (sync_read and async_read) should be modify with the new definition of data.

stnolting:
Hey @doud3157. Thanks for finding this!

I know Vivado has some issues with multi-dimensional arrays sometimes (and a reocord could be seen as suchy), but I think the problem is somewhere else. This "dual-access" to the FIFO RAM might be an issue:

```vhdl
if (fifo_depth_c = 1) then
  rdata_o <= fifo.data(0);
else
  rdata_o <= fifo.data(to_integer(unsigned(fifo.r_pnt(fifo.r_pnt'left-1 downto 0))));
end if;
```

Anyway, I will do some tests and update the FIFO module.

robhancocksed:
It looks like this might still be an issue, at least with respect to the SLINK module where it generates a 33 bit wide FIFO. If I set the FIFO depths to 1024 words for RX and 512 for TX, for example, it seems to not be inferring block RAM but 33-bit registers, which uses a large amount of logic resources.

stnolting:
I've just tested that using Vivado - and you are right, Vivado is trying to build the SLINK's FIFOs from FF primitives. It seems like Vivado is not happy with these size checks:

https://github.com/stnolting/neorv32/blob/4155b728f2b4cfcd4dfc2bc1dbf8f0b5cc7e68d6/rtl/core/neorv32_fifo.vhd#L190-L196

I think that somehow Vivado is not capable of identifying these checks as "runtime static". Maybe it is better to use `if-generate` here? 🤔

@mikaelsky what do you think (I think we've added those checks in #778)?

umarcor:
[...]

PR #827, opened 2024-02-23
[fifo] fix (Vivado) synthesis issue

stnolting:
Vivado was not able to infer block RAM / LUT-RAM for the FIFO memory.

@robhancocksed, @mikaelsky, @umarcor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Link |
 |:----:|:-------:|:--------|:----:|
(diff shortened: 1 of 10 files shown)


robhancocksed:
I can confirm that Vivado is now inferring the CPU prefetch buffer, SLINK and UART FIFOs as RAM rather than registers (though in this case distributed RAM rather than block RAM, since they are set to be quite small).

stnolting:
> I can confirm that Vivado is now inferring the CPU prefetch buffer, SLINK and UART FIFOs as RAM rather than registers (though in this case distributed RAM rather than block RAM, since they are set to be quite small).

Great to hear!

Thanks for the feedback!

umarcor:
I can also confirm that I can now get expected results (https://github.com/stnolting/neorv32/issues/753#issuecomment-1960169120) with the main branch. I merged https://github.com/stnolting/neorv32-setups/pull/150 for neorv32-setups to pick recent changes and then implementation suceeded without further changes.

stnolting:
Thanks @umarcor! :+1:

Issue #825, opened 2024-02-22
Vivado warnings/errors for SLINK AXI stream interface clocks


robhancocksed:
**Describe the bug**
When using the AXI-Lite top-level interface, Vivado 2023.2 produces critical warnings like this for the AXI Stream interfaces for SLINK:

[BD 41-967] AXI interface pin /processor_subsystem/neorv32_SystemTop_ax_0/s1_axis is not associated to any clock pin. It may not work correctly.

This can also lead to block design validation errors when connecting the interfaces to other components such as the Xilinx AXI DMA, since Vivado cannot detect the correct clock frequency for the AXI stream interface (it seems to default to 100 MHz) and detects a mismatch with the other component. It is possible to work around this by manually setting the CLK_DOMAIN and/or FREQ_HZ parameter on the AXI stream interface on the NEORV32 core in the block design. However, it seems that these overridden settings do not always "stick" when exporting the block design back into tcl code for version control, so this is a nuisance to manage.

It looks like adding some dummy clock inputs to the top-level AXI wrapper (rtl/system_integration/neorv32_SystemTop_axi4lite.vhd) avoids this problem, for example:

```
    -- ------------------------------------------------------------
    -- AXI4-Stream-Compatible Interface --
    -- ------------------------------------------------------------
    -- Source --
    s0_axis_tdata  : out std_logic_vector(31 downto 0);
    s0_axis_tvalid : out std_logic;
    s0_axis_tlast  : out std_logic;
    s0_axis_tready : in  std_logic;
    s0_axis_aclk   : in  std_logic;
    -- Sink --
    s1_axis_tdata  : in  std_logic_vector(31 downto 0);
    s1_axis_tvalid : in  std_logic;
    s1_axis_tlast  : in  std_logic;
    s1_axis_tready : out std_logic;
    s1_axis_aclk   : in  std_logic;

```

Then if the same clock input used for m_axi_aclk is connected to these new clock ports, Vivado is able to figure out the clock domain for the AXI stream ports properly.

I guess the only downside of doing this is it might create the impression that the core supports using different AXI stream clocks from the main core clock, when it really doesn't. I suppose if you did try to use a mismatched clock it would probably end up with a timing failure, unless you happened to use clocks that were related/divisible which could just cause duplicate data words or something.

stnolting:
Good point. Seems like I have never used the SLINK in a Vivado block design... 🙈

> `s0_axis_aclk   : in  std_logic;`

[...]

PR #831, opened 2024-02-23
Added dummy clocks for SLINK streams in AXI4-Lite wrapper

robhancocksed:
Fixes Vivado clock domain warnings/errors when connecting SLINK interfaces to other AXI4-Stream components, as described in https://github.com/stnolting/neorv32/issues/825


stnolting:
Thank you very much!

stnolting (approved):

Issue #55, opened 2021-06-03
Distribution/installation of the GCC toolchain


umarcor:
Coming from #54 

> https://github.com/stnolting/neorv32/pull/54#issuecomment-853920947
> By the way, what do you think about having the pre-build GCC toolchain as submodule inside this repository? Maybe in sw/gcc?
> Tha scripts (makefiles) could use that as default so new-user could do a simple recursive git clone and they would get everything as a bundle 🤔

> https://github.com/stnolting/neorv32/pull/54#issuecomment-853938959
> That's something I wanted to bring in a different issue 🤣. So, why are you using your pre-built GCC toolchain? Do you need to customise it so that it understands NEORV32 as the target architecture? Or is it because of convenience? That is, could you use an "upstream" GCC? May NEORV32 be upstreamed?
> 
> This is a tool which your project depends on. It should not be a submodule inside the repository. Furthermore, most users should not install it through a custom script in this repo. Similarly to GHDL, Yosys, nextpnr, VUnit, etc. those should be installed independently. We want this project (and potentially any open source project) to be usable very regardless of the preferred environment of the user. Some will install system packages, others will use containers, MSYS2, Conda, Homebrew... They don't want to install those tools each time, but (very) infrequently. In that context, you are currently providing a single build for a very specific subset of environments.
> 
> Please, don't take me wrong. This is absolutely normal. Most open source EDA tooling projects are not properly packaged yet. Some (a few) are available on some (a few) popular distributions. There are efforts such as HDL's bazel and conda repos, or YosysHQ's oss-cad-toolchain. Yet, each of them is an ecosystem for the domain the maintainers/supporters are interested in.
> 
> Should you need to distribute a custom GCC because you need it customised, I would recommend that we add the generation of a container image and an MSYS2 package to you other repo. Then, consume those as any user will do. In fact, this is one of the main purposes of CI: to reproduce the experience you expect users to have on different platforms.

> https://github.com/stnolting/neorv32/pull/54#issuecomment-853950156
> Then let's discuss that in a new issue. 😄

> https://github.com/stnolting/neorv32/pull/54#issuecomment-853950156
> > This is a tool which your project depends on. It should not be a submodule inside the repository.
>
> Yeah right. You convinced me. 😉

stnolting:
[...]

PR #832, opened 2024-02-25
[sw] remove unused variable RISCV_TOOLCHAIN

umarcor:
While trying to build and install the bootloader on MSYS2 following my notes from #55, I found the following typo: `s/RISCV_TOLLCHAIN/RISCV_TOOLCHAIN/`. So I fixed it.

However, then I found out that it seems not to be used anymore. Now `RISCV_PREFIX` is used: `RISCV_PREFIX=riscv64-unknown-elf- MARCH=rv32i make -C sw/bootloader/ install`. Therefore, I removed it. Yet, I don't know if that's correct.

@stnolting wdyt?

### DIFF
--- a/sw/common/common.mk
+++ b/sw/common/common.mk
@@ -358,7 +358,7 @@ info:
 	@echo "------------------------------------------------------"
 	@echo "-- Toolchain"
 	@echo "------------------------------------------------------"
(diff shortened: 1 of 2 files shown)


stnolting:
Looks like a leftover from the early days...

stnolting (approved):

PR #840, opened 2024-03-06
Performance tests

mikaelsky:
Added a benchmark test bench that is being used for the performance tests. The benchmark testbench has almost everything in the core disabled. It has 128kb of IRAM to ensure we have enough space for the test suites and U mode disabled.

Added a test finish process allowing the  main.c to signal the testbench to stop running.

Updated common.mk to support using something other than the simple testbench using makefile variables.

Added a run_performance.sh script that runs through all the tests with the _all setting. This will basically plot all the performance numbers measured.

Fixed a few bugs in the I performance test causing test failures.

TODO: Fix the mret test as its still failing. Added support for C and B extensions.


mikaelsky:
Note: Software pull request fails as its going to the performance_test area and trying to build, which right now doesn't work.

stnolting:
FTR

Update of this PR in #844

PR #854, opened 2024-03-18
Connected SPI bus to on-board Flash and SPI peripheral

lovelesh-mis:
- Added Connection to on-board FLASH and SPI bus. 
- Resolved ERR_FLSH error
- Storing to Flash and booting from Flash working


stnolting:
Oh, seems like the SPI signals got lost during some earlier update of this wrapper... 🙈 Thanks for fixing this!

PR #862, opened 2024-03-24
Support tool-specific standard flags in makefile

### LINKED ISSUE #856: [feature request] Makefile: Additional CXX_USER_FLAGS only for C++ Compilation
**Is your feature request related to a problem? Please describe.**

When compilling C++ code, I sometimes need flags which are not valid for C code, for example `-fno-rtti -fno-use-cxa-atexit -fno-threadsafe-statics`. I can just pass them to `USER_FLAGS`, but then they also get passed when compiling C files. in that case, GCC complains with warnings:
`cc1: warning: command-line option '-fno-rtti' is valid for C++/D/ObjC++ but not for C`

**Describe the solution you'd like**
Split `CC_OPTS` into `C_OPTS` and `CXX_OPTS`. Introduce `CXX_USER_FLAGS` and only use those for `CXX_OPTS` and in cpp compilation.

Though I guess there are other options to implement this as well. In the end, anything that allows specifying flags in the user Makefile that only get passed to the C++ compiler should work.

jpf91:
This introduces the following new makefile variables: `CFLAGS` for C, `CXXFLAGS` for C++, `ASFLAGS` for Assembler and `LDFLAGS` for the linker. Names follow  conventions from  https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

These variables can then be set either in the user application makefile (before including `common.mk`) or on the command line when calling make.

Fixes #856

### DIFF
--- a/docs/datasheet/software.adoc
+++ b/docs/datasheet/software.adoc
@@ -224,6 +224,14 @@ MABI  ?= ilp32
 USER_FLAGS ?=
 # User libraries (will be included by linker)
 USER_LIBS ?=
+# Language specific compiler flags: C
+CFLAGS ?=
+# C++
+CXXFLAGS ?=
+# Assembly
+ASFLAGS ?=
+# Flags passed only to the linker
+LDFLAGS ?=
 # Relative or absolute path to the NEORV32 home folder
 NEORV32_HOME ?= ../../..
 # GDB arguments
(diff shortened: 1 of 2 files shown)


stnolting:
This looks great! Thanks for implementing this!

Btw, I do not think we should set a new "version" here. This is just a minor change that is fully backwards compatible (:+1:).

It would also be nice if perhaps two lines could be added to the [makefile's documentation](https://stnolting.github.io/neorv32/#_makefile_configuration) 😅.

jpf91:
> This looks great! Thanks for implementing this!

Thanks :-)

> Anyway, I do not think we need to set a new "version" here.

:+1: I was a bit confused as all changelog entries have a new version. I guess I shouldn't add a changelog entry then? Or is it ok to have an entry but keep the old version?

> It would also be nice if perhaps two lines could be added to the [makefile's documentation](https://stnolting.github.io/neorv32/#_makefile_configuration) 😅.

Good point, I totally forgot about that :innocent:  I've pushed a new revision which adds the documentation and removes the version change :smile:

stnolting:
> I guess I shouldn't add a changelog entry then? Or is it ok to have an entry but keep the old version?

Well, that's a good point... The [changelog](https://github.com/stnolting/neorv32/blob/main/CHANGELOG.md) describes the "version increments" as follows: "The identifier is incremented for each core hardware change and also for major software and/or general project changes."

But this doesn't really clarify how a "major change" is actually defined... 😅

In this case, it is only a minor change that is also fully backwards compatible. I therefore think that it is okay not to assign a new version number for this. 😉

stnolting (approved):

jpf91:
Thanks!

PR #865, opened 2024-03-26
:warning: rework TWI module

stnolting:
## ✨ Add Configurable TWI FIFO

An optional FIFO is implemented that can be used to "program" long sequences of TWI operations (START conditions, repeated START conditions, STOP conditions, data transmissions) without further interaction of the CPU. The FIFO is configured by a new top generic:

```vhdl
IO_TWI_FIFO : natural range 1 to 2**15 := 1;
```

## :warning: Rework TWI Core

* remove support for clock-stretching
* remove bus state sensing
* rework low-level interface / HAL
* interrupt will fire when the FIFO is empty and no bus operation is pending

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Link |
 |:----:|:-------:|:--------|:----:|
+| 26.03.2024 | 1.9.7.6 | :warning: rework TWI module; add configurable command/data FIFO | [#865](https://github.com/stnolting/neorv32/pull/865) |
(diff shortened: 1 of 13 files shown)


NikLeberg:
Hi @stnolting, I just skimmed the changes and saw that clock stretching detection was removed. Is it gone for good? So very simple TWI devices that require clock stretching will no longer work?

stnolting:
After reading [this](https://learn.adafruit.com/working-with-i2c-devices/clock-stretching) great guide provided by AdaFruit, I had the feeling that clock stretching is not a good idea 🤔

Furthermore, I do not know of any off-the-shelve components that actually use clock stretching.

However, maybe I was too hasty. Do you think this is a handy feature that we should add again?

NikLeberg:
I don't think that clock stretching is all that bad. But of course slaves that do it in weird ways could be a pain.

The BNO055 which also AdaFruit mentions is actually a sensor that utilises/requires clock stretching. I used it together with an ESP32 without problems. And I think to remember seeing actual clock stretching on my digital logic analyzer.

I havent used the BNO055 (a really excellent sensor I have to say) with NEORV32. So I don't know if the previously present clock stretching detection would have worked. But if it does not cause too much hastle I'd recommend or actualy prefer the functionality brought back.

PR #867, opened 2024-04-01
add back TWI clock stretching option

stnolting:
requested by @NikLeberg in https://github.com/stnolting/neorv32/pull/865#issuecomment-2029060807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Link |
 |:----:|:-------:|:--------|:----:|
+| 01.04.2024 | 1.9.7.7 | add back TWI clock stretching option | [#867](https://github.com/stnolting/neorv32/pull/867) |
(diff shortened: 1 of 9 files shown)


stnolting:
@NikLeberg 

Do you have a peripheral device available that actually makes use of clock stretching? The simple port expanders and EEPROMs on my desk are also happy without this function.

NikLeberg:
Hi @stnolting, thanks for swiftly bringing this back!

At home I do have the mentioned BNO055 and could test it... but I'm on pretty long backpacker holidays 😇. So I don't have access currently to it for a few months.

Maybe I can come up with a way to simulate and test it.

stnolting:
> but I'm on pretty long backpacker holidays 😇

Now that sounds nice! Which continent are you exploring?

> Maybe I can come up with a way to simulate and test it.

Let's just merge this for now going back to the previous clock stretching mechanism.

NikLeberg:
> Now that sounds nice! Which continent are you exploring?

We did India, currently Nepal and soon south-east asia with Thailand, Vietnam and more. Lets see. Gonna be a few more months. 🏝🤗

stnolting:
Asia is nice. Definitely one of my 7 favorite continents! 😄
Have fun!

Issue #879, opened 2024-04-19
JTAG halt request gets missed if it occures during an illegal instruction


mikaelsky:
**Describe the bug**
While debugging the OCD interface using a Segger J-Link we've noticed that if we issue a halt request and the core is executing an illegal instructions then the halt request gets discarded.

**To Reproduce**
The simplest way to reproduce it, which is how we found it, is to create an 0'ed ROM image which will cause the core to go into an infinite illegal instruction loop. This as mtvec as default points to the ROM whose first location is all 0s which triggers the illegal instruction trap.
When the core is in this state you can see the trap.cause shift to 0x63 (halt request seen) but it gets overwritten by the illegal instruction trap 0x02 as that has a higher priority.

I've attached a screen grab from a Xilinx FPGA image with ILAs showing the trap_ctrl.cause, the cpu_halt_req_o from the DM module and the PC. The PC is looping between the first 2 ROM instructions (all 0's) and you can the cause register flipping between 0x63 and 0x02 with 0x02 winning as we keep looping around the first ROM address.

**Expected behavior**
This I'm a bit unsure of as I'm not 100% into the RISCV specification regarding debug behavior. I'm assuming that halt request would take the highest priority to ensure that no matter the state of the core you can always halt it and get into debug mode.

**Screenshots**
![image](https://github.com/stnolting/neorv32/assets/119539842/3b0f5afb-1c86-47b4-a797-463ba90fbe08)

**Environment:**
 - OS: RHEL 7
 - GCC Version RISCV gcc 13.1
 - Libraries used: newlib

**Hardware:**
 - Hardware version (1.9.3) bug patched

**Additional context**
I have a local fix to the issue which is to re-order the trap_ctrl.cause priority to make halt request the highest priority. With the fix we can successfully connect to the neorv32 core with Segger J-Link.

```
      -- if we are supposed to halt, always halt
      if (trap_ctrl.irq_buf(irq_db_halt_c)  = '1') then trap_ctrl.cause <= trap_db_halt_c;  -- external halt request (async)
--      if    (trap_ctrl.exc_buf(exc_ialign_c)   = '1') then trap_ctrl.cause <= trap_ima_c;  -- instruction address misaligned
```

[...]

PR #882, opened 2024-04-23
fix external debug-halt vs. exception concurrency

stnolting:
Fixing #879.

The CPU can now be halted right after reset **before** the first instruction is executed. However, the on-chip debugger does not provide the "halt on reset" functionality yet (but that can be emulated by the debugger).


mikaelsky (approved):
Looks good to me. Will see if I can patch up my 1.9.3 core version with this change to do a real-time debug on an FPGA.

stnolting:
I've tested this by replacing the IMEM by a constant `x"00000000"` (illegal instruction) and I was able to halt the core at any time (even right after reset 🎉).

So, hopefully, this fixed the "bug".

Thanks again for your help! :+1:

Issue #890, opened 2024-05-02
[bug] uart issue after #883 upd


Polaris-III:
**Describe the bug**
After #883 i discover that neorv32 determines any printed character as "?".
Looks like #883 uart_rx upd broke something

**To Reproduce**
Check "sreg" bus of "rx_engine_t" record in "neorv32_uart.vhd" file.
before #883 - sreg[9:0]
after #883 - sreg[8:0]  <- where this issue happened

**Expected behavior**
Correct character determination

**Screenshots**
I push "e" button to abort autoboot sequence
Screen 1 - incorrect symbol.
![Screenshot from 2024-05-02 15-16-15](https://github.com/stnolting/neorv32/assets/42891977/16492230-d881-4630-870f-26d42382161f)

Screen 2 - correct after bus fix
![Screenshot from 2024-05-02 15-22-33](https://github.com/stnolting/neorv32/assets/42891977/a84b3146-aaac-4c69-890a-518719125a2d)

**Environment:**
 - OS Ubuntu 
 - Xilinx xc7k325 fpga
 - ft232rl as uart2usb bridge
 - stock bootloader image

**Hardware:**
 - Hardware version (1.9.8.9)

**Additional context**
I changed sreg back to "9 downto 0" and reimplement design - works fine

Unike267:
Hi everyone!

I've reproduced this bug in Arty A7 35t.

As shown in the following result via CuteCom:

![issue_890](https://github.com/stnolting/neorv32/assets/36209778/c585b1a0-d0dc-4fc7-add8-db41d449b131)

I hope it will be solved soon,

Cheers!

Unike267:
Hi,

I've fixed this in local changing this line:

https://github.com/stnolting/neorv32/blob/e2fa5c7579444b1142de635ae0cf2fd82dd30ab5/rtl/core/neorv32_uart.vhd#L289

From `rx_engine.sreg(8 downto 1);` to `rx_engine.sreg(7 downto 0);`

Cheers! 😃

Polaris-III:
Hi @Unike267, I'll check your solution tomorrow, thank you

stnolting:
Seems like I have messed up the UART receiver 🙈

@Polaris-III thanks for finding this bug!

@Unike267 your fix looks great! Would you open a PR? :wink:

PR #891, opened 2024-05-02
Fix UART receiver

Unike267:
According to #890
[bug] uart issue after #883 upd

The change is in neorv32_uart.vhd file
On line 289:
From rx_engine.sreg(8 downto 1) to rx_engine.sreg(7 downto 0)


stnolting (approved):

stnolting:
Thanks for fixing! :+1:

Issue #878, opened 2024-04-18
ASIC Processor boot conccept


ucycg:
I'm currently exploring the possibility to create an ASIC from a specific NeoRV32 configuration. As I'm still a student without much experience in digital design I still have a lot of things to figure out on this endeavour. 

One thing I currently came across, as it is fundamental for the ASIC processor to be actually usable is the boot concept. I was always thinking I would use the bootloader to upload binaries from an External SPI Flash via the auto boot SPI process. But today I realized that the bootloader itself is contained in the neorv32_boot_rom.vhd which isn't exactly synthesizable either. 

So now my second next idea would be to use the XIP module with another External SPI Flash as a replacement ROM which contains the bootloader code instead. Is this a reasonable approach? 

When I use the XIP module is it still possible to also execute instructions from internal IMEM as well?

As for the processor internal IMEM/DMEM I have the plan to replace the architecture of these moduels with a technology dependent version which is based on RAM.

ucycg:
Yeah so I managed to synthesize a netlist in our tech process and simulated the startup of the bootloader successfully which tells me that the boot_rom is successfully synthesized in the GTL netlist. I would still like to find out how much area is occupied by that memory, which i couldnt figure out until now sadly.
![Bildschirmfoto vom 2024-04-19 15-42-50](https://github.com/stnolting/neorv32/assets/101277242/d64cdec3-aa6f-4c0d-8943-8caa30bc92a2)

stnolting:
That looks great! :+1:
Keep us updated if you like :wink:

stnolting:
Sounds like a cool project!

If you "just" want to support read accesses (i.e. no flash write) you only need to implement the READ command, so the logic should be quite simple. Hit me up if I can help you in any way.

jtplaarj:
Picorv32 has such a simulation module por XIP: https://github.com/YosysHQ/picorv32

I have successfully used it with neorv32.

stnolting:
You mean this one? https://github.com/YosysHQ/picorv32/blob/main/picosoc/spiflash.v

I thinks it looks quite promising.

jtplaarj:
Yes. I have done small changes, it looks like neorv32 implementation of XIP and that of picorv32 are not exactly the same. the spi simulator from picorv32 waits for one command to initialize the flash, while neorv32 does not send it. Because of that, I have changed line 63 and initialize `reg powered_up = 1` to avoid the issue.

stnolting:
[...]

PR #903, opened 2024-05-16
Update neorv32_sdi.vhd - Minor typo correction

ucycg:
RX FIFO misses the value setting for FULL_RESET generic completly. I suppose then it falls back to the default value, since that one is false too in the neorv32_fifo.vhd this commit isn't leading to any different behavior in RTL, but should be added just for consisentcy issues


stnolting:
Looks great!

Btw how can I make these nice links of graphical representation of code from a file like you did in https://github.com/stnolting/neorv32/issues/878 inside a comment instead of just linking it?

grafik

stnolting (approved):

PR #996, opened 2024-08-25
RTL reworks, cleanups and optimizations

stnolting:
* Vivado IP module: major rework; use explicit `std_logic[_vector] <-> std_ulogic[_vector]` conversion; remove VHDL2008 requirement (not supported by the IP packager?)
* CPU: remove output register of multiplier unit to improve mapping to DSP block (`FAST_MUL_EN` option only)
* CPU: :warning: remove user-level HPMs
* CPU: rework `mcountern` CSR; add explicit `CY` and `IR` bits
* CPU: rework "next PC" logic - use ALU adder to compute the next instruction address (removes one carry chain and shortens the critical path of the base CPU core)
* DOCs: updates, cleanups and fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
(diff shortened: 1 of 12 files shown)


Sam-Vervaeck (commented):
I had a look through all the changes, did not see anything weird.

stnolting:
@Sam-Vervaeck Thanks for reviewing! :)

I'm still doing some (synthesis) tests, but everything seems to be fine so far.

Sam-Vervaeck:
@stnolting This is still in the user guide but I think this pull request removed the need for this disclaimer in section 12?
image

stnolting:
You are right! I forgot to remove that... Thanks for pointing that out!

PR #1006, opened 2024-09-07
minor fixes in OCD

NikLeberg:
This proposes some minor fixes (no impact expected) of the OCD.
 - fix address map in  linker script (not really important, but just for completeness)
 - fix renaming of `dm_base_c` to `base_io_dm_c`

As I don't think the changes will have any impact, I left out the `CHANGELOG.md` update. Let me know if you'd like to have one nonetheless.


stnolting (approved):

stnolting:
This looks great - thank you very much!

> fix address map in linker script (not really important, but just for completeness)

Luckily the code only uses indirect jumps, otherwise it would have been a real bug. 🙈

> As I don't think the changes will have any impact, I left out the CHANGELOG.md update. Let me know if you'd like to have one nonetheless.

:+1:

Issue #1024, opened 2024-09-19
Check code quality in a continuous way


racodond:
Hi,

I analyzed the VHDL code from your NEORV32 project with Linty: https://oss.linty-services.com/dashboard?id=neorv32&codeScope=overall

Do not get scared by the number of issues :-)
It's just to provide you with data to browse. I activated all the rules, some are mutually opposed, others are not properly configured with your requirements (such as naming conventions).

There might be a potential bug behind that one though: https://oss.linty-services.com/project/issues?issues=0bea7495-7740-4097-91c3-baac7057b3f4&open=0bea7495-7740-4097-91c3-baac7057b3f4&id=neorv32

We offer Linty for free for open-source projects. I'd be glad to organize a call if you're interested in checking code quality in a continuous way through GitHub actions.
We also offer our VS Code extension for free: https://hdl-designer.linty-services.com/

I'm looking forward to hearing from you.

Thank you

Have a good day

David Racodon
CTO at Linty Services

racodond:
> Wow, this looks really amazing! Linting has been on my to-do list for quite some time... 😅

You'll see that Linty does a lot more than linting -:)
CDC, RDC, etc.
See https://oss.linty-services.com/coding_rules?languages=hdl and https://oss.linty-services.com/coding_rules?languages=vhdl

> That sounds really great and I appreciate your offer. Can you provide a link to the according GH action? Or maybe a link to some documentation?

You're welcome!

Here's the documentation: https://doc.linty-services.com
To scan your code: https://doc.linty-services.com/scan.html (see GitHub section)

If you provide me with a name and an email address, I'll create an account for you on our OSS cloud platform (https://oss.linty-services.com/) so that you'll be able to update your quality profile (set of rules) and run Linty analyses.

stnolting:
> You'll see that Linty does a lot more than linting -:)
CDC, RDC, etc.

That sounds really great! :+1:

> Here's the documentation: https://doc.linty-services.com/
To scan your code: https://doc.linty-services.com/scan.html (see GitHub section)

Thanks - I'll have a look at those.

> If you provide me with a name and an email address, I'll create an account for you on our OSS cloud platform (https://oss.linty-services.com/) so that you'll be able to update your quality profile (set of rules) and run Linty analyses.

Can I just create an OSS account over there?

[...]

PR #1028, opened 2024-09-20
:bug: fix minor bug in FPU MUL instruction

stnolting:
Fix invalid instruction logic (if any operand is sNAN).

Bug identified by @racodond in #1024.

@mikaelsky, maybe this is also interesting for you :wink:


mikaelsky:
Interesting.. wonder why it was caught by the compliance coverage test. Will add it to my internal test list.

PR #1030, opened 2024-09-22
[rtl] remove redundant `prog_buf`

NikLeberg:
Just found this signal that is completely redundant. I kinda remember us having found that and removing it a few months ago... did it creep back it? 🤔

Unrelated: I'm playing around with implementing system bus access for the DM. Is that something you would accept a PR for?

### DIFF
--- a/rtl/core/neorv32_debug_dm.vhd
+++ b/rtl/core/neorv32_debug_dm.vhd
@@ -133,10 +133,6 @@ architecture neorv32_debug_dm_rtl of neorv32_debug_dm is
   end record;
   signal dm_ctrl : dm_ctrl_t;
 
-  -- program buffer access --
-  type prog_buf_t is array (0 to 3) of std_ulogic_vector(31 downto 0);
-  signal prog_buf : prog_buf_t;
-
   -- **********************************************************
   -- CPU Bus and Debug Interfaces
   -- **********************************************************


stnolting:
> Just found this signal that is completely redundant.

You're right!

> I kinda remember us having found that and removing it a few months ago... did it creep back it?

I can't remember... 😅

> Unrelated: I'm playing around with implementing system bus access for the DM. Is that something you would accept a PR for?

It has been some time since I have dealt with the debug-spec... Correct me if I'm wrong, but the system bus access allows the debugger to directly access the entire address space without going through the CPU, right?

To be honest, I do not see any benefit here (except for speed improvements). Furthermore, this would require another host interface for the bus system increasing area consumption. 🤔

stnolting (approved):

NikLeberg:
> To be honest, I do not see any benefit here (except for speed improvements). Furthermore, this would require another host interface for the bus system increasing area consumption. 🤔

You were right. Got it working and the test for my `cosim_jtag` project merely runs 10 seconds faster. Still a 10% improvement, but not worth the area increase.

stnolting:
> Got it working and the test for my cosim_jtag project merely runs 10 seconds faster. Still a 10% improvement, but not worth the area increase.

That's more than I have expected! 😅 

Did you do some synthesis experiments? Can you say something about logic utilization and critical path / f_max impact?

NikLeberg:
I pushed the changes to a branch on [NikLeberg/neorv32](https://github.com/NikLeberg/neorv32/commit/c43bf0bb877cca61bc2f3b8c5cf340700afa8904). It is very much WIP, unoptimized and ~potentially~ _likely_ containing bugs.

|  | without SB | with SB |  |
| --- | --- | --- | --- |
| Logic Cells (only DM) | 387 | 638 | +251 |
| Registers (only DM) | 243 | 363 | +120 |
| f_max [MHz] | 70.79 | 69.29 | -0.15 |

My f_max is generally very slow, as I'm synthesizing for an old Cyclone IVE where 80 MHz would be the theoretic maximum if you optimize it down to the logic element. I usually run it at 50 MHz.

stnolting:
That is just the modified DM, right? So without the additional bus mux?

Does _f_max_ show the critical path of the DM or of the entire processor?

NikLeberg:
Aah, right, forgot to include the muxes logic cell count... 

But _f_max_ is of the entire processor and propably caused by the additionally required bus mux.

PR #1031, opened 2024-09-23
[rtl] fix generate spelling

NikLeberg:
Just a little something.

BTW: I'm assuming the generate(s) with `true` value are for better visibility and partitioning when viewing the hierarchy in simulators etc. Is there a specific reason why you didnt use VHDL standard `block`s? E.g.:

```vhd
-- **************************************************************************************************************************
-- Clock and Reset Generators
-- **************************************************************************************************************************
generators: block
begin

-- Reset Sequencer ------------------------------------------------------------------------
neorv32_sys_reset_inst: entity neorv32.neorv32_sys_reset
port map (
  clk_i      => clk_i,
  rstn_ext_i => rstn_i,
  rstn_wdt_i => rstn_wdt,
  rstn_dbg_i => dci_ndmrstn,
  rstn_ext_o => rstn_ext,
  rstn_sys_o => rstn_sys
);

-- <...>

end block generators;
```
BTW: Is there a specific reason you use


stnolting:
> I'm assuming the generate(s) with true value are for better visibility and partitioning when viewing the hierarchy in simulators etc.

Right.

> Is there a specific reason why you didnt use VHDL standard blocks? E.g.:

That's a very good question... I also realized that I never ever use the `with select` construct... Maybe I just don't like the way they look 🤣 I don't know.

Anyway, is the `block` does not require VHDL2008, right?

Furthermore, I think that the `block` statement can only be used to "encapsulate" concurrent statement, correct? So processes won't work there?!

stnolting (approved):

PR #1056, opened 2024-10-12
:warning: Remove OCD's DM legacy mode

stnolting:
* debug module is compatible to spec. version 1.0 that has been ratified over a year ago and is widely supports (like upstream opcenOCD)
* remove `OCD_DM_LEGACY_MODE` generic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
-| 11.10.2024 | 1.10.5.5 | :sparkles: :lock: add optional support for on-chip debugger authentication; :warning: rename OCD-related top generics | [#](https://github.com/stnolting/neorv32/pull/1053) |
+| 12.10.2024 | 1.10.5.6 | :warning: remove legacy support for on-chip debugger DM version v0.13; now only supporting DM v1.0 (removing `OCD_DM_LEGACY_MODE` generic) | [#1056](https://github.com/stnolting/neorv32/pull/1056) |
+| 11.10.2024 | 1.10.5.5 | :sparkles: :lock: add support for optional on-chip debugger authentication; :warning: rename OCD-related top generics | [#1053](https://github.com/stnolting/neorv32/pull/1053) |
 | 06.10.2024 | 1.10.5.4 | :warning: rework PWM module | [#1049](https://github.com/stnolting/neorv32/pull/1049) |
 | 05.10.2024 | 1.10.5.3 | upgrade neoTRNG to version 3.2 | [#1048](https://github.com/stnolting/neorv32/pull/1048) |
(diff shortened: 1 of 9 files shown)


stokdam:
Hi! Unfortunately Microchip has not updated SoftConsole version yey.  They're still stuck at version 2022.2. There is no easy way to obtain source code of their modified openocd (shipped with SoftConsole) which includes support for their proprietary JTAG controller FlashPro. This PR drops support for every Microchip FPGA.

stnolting:
Hey @stokdam!

> Unfortunately Microchip has not updated SoftConsole version yet. They're still stuck at version 2022.2.

Oh really? Wow, they should really catch up development. 🙈

I'm not familiar with SoftConsole, but under the hood it seems to be plain Eclipse, right? Maybe you could just configure a different openOCD version that supports newer RISC-V debug specs (just select a different openOCD installation - no source code modifications).

At least this is an approach I have used for some other "proprietary" RISC-V development environments that are also based on Eclipse (I\*tel P\*thfinder).

> This PR drops support for every Microchip FPGA.

Basically, it is not the FPGA itself, right? It is just the internal JTAG access port. An external JTAG bridge would still work with Microchip FPGAs, wouldn't it?

I understand that this is annoying. I also don't like it when equipment manufacturers don't update their toolchain (or don't include instructions on how to do it yourself).

At the same time, however, it is also difficult to be backwards compatible and to maintain this in newer versions. So I am not really sure how to proceed. 🤔

The DM interface of NEORV32 did not change - so replacing the upstream one by the legacy one (that still provides compatibility with Microchip tools) should be possible. But I can also understand that this is quite fiddly and annoying in the long run.

stokdam:
> I'm not familiar with SoftConsole, but under the hood it seems to be plain Eclipse, right?
Yes, it is Eclipse, but actually I don't need anything else than their openocd.
> Maybe you could just configure a different openOCD version that supports newer RISC-V debug specs
The problem is support of their proprietary JTAG programmer FlashPro, which is embedded in the modified openocd they ship with SoftConsole. Moreover, it implements the tunneling protocol needed to use the same pin for FPGA programming as user JTAG (main TMS signal is not available for user, they provide a bridge that decodes encapsulated JTAG "packets" just from TDI pin and regenarates all JTAG control signals https://ww1.microchip.com/downloads/aemDocuments/documents/FPGA/ProductDocuments/UserGuides/ip_cores/directcores/CoreJTAGDebug_UG.pdf).
immagine

What I can do for now is to connect neorv32 JTAG signals directly to some user pins and use a different JTAG adapter.

stnolting:
> The problem is support to their proprietary JTAG programmer FlashPro, which is embedded in the modified openocd they ship with SoftConsole.

Ah okay, I understand the problem. :(

> What I can do for now is to connect neorv32 JTAG signals directly to some user pins and use a different JTAG adapter.

Sure, that should work. But I understand that this is not the best solutions (requiring an additional adapter + USB port).

I would recommend to use the last DM module that still supports the legacy mode and just use that instead of the one from the main branch.

Alternatively, we could try to figure out which specific DM features are required by your openOCD version. If it is just the version identifier (and maybe so blank register addresses) then we could try to come up with a DM that follows the up-to-date spec but is still backwards compatible to your setup.

PR #1057, opened 2024-10-14
[sw/lib] restore previous registers in neorv32_xirq_get_num function to avoid side effects

donlon:
Hi,
The `neorv32_xirq_get_num` function didn't restore previous `CSR_MIE` and `NEORV32_XIRQ->EIE` register and could cause unexpected side effects. So these registers are now saved before writing test bits and then restored finally.


stnolting:
Good point - thanks for the update!

stnolting (approved):

PR #1058, opened 2024-10-14
[vivado ip] hide entire AXI-Stream interface in block diagram if disabled

donlon:
Currently empty ports for AXI-Stream interfaces (`s1_axis` and `s0_axis`) are still shown even if they're not enabled (shown in the picture). Setting `enablement_dependency` property for the bus interfaces instead can hide them entirely.

Tested in Vivado 2022.2

neorv-axis

### DIFF
--- a/rtl/system_integration/neorv32_vivado_ip.tcl
+++ b/rtl/system_integration/neorv32_vivado_ip.tcl
@@ -66,8 +66,8 @@ set_property description "The NEORV32 RISC-V Processor" [ipx::current_core]
 # **************************************************************
 # Interfaces: Configuration Dependencies
 # **************************************************************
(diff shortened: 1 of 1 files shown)


stnolting:
Hey @donlon!

> Setting enablement_dependency property for the bus interfaces instead can hide them entirely.

That's great! I was not aware of this property.

So we could also use this approach to make the AXI4-MM optional, right?

stnolting (approved):

donlon:
> So we could also use this approach to make the AXI4-MM optional, right?

Yeah it does. We can use the same approach to make it more neat. (And I just found out that the external bus is optional even though I have used the core for a while.)

I'm also trying to reorganize the layout of the IP GUI and maybe will open another PR for them later :)

PR #1060, opened 2024-10-16
[vivado ip] reorganize IP GUI layout

donlon:
Hi!

So I used the approach in #1058 to make m_axi bus optional, and made some parameters editable depending on other parameters. (e.g. XBUS_CACHE_EN is editable only when XBUS_EN is set)

I also reorganized the IP GUI by splitting parameters into pages and groups, which maybe look more neat and is easier to use.


donlon:
The attributes for JTAG ports doesn't compile in GHDL, not sure if it's a bug of it.🤔

donlon:
Sure! And I'm closing this PR to make it convenient.

PR #1064, opened 2024-10-18
[vivado ip] reorganize Vivado IP GUI

donlon:
Parameters inside the IP GUI are split into pages and groups, and there editability are depending on other corresponding parameters (e.g. XBUS_CACHE_EN is editable only when XBUS_EN is set). Hopefully it looks more neat and is easier to use now :)


stnolting:
The new GUI layout looks really really great!

Individual tabs and even conditional dependencies of individual configuration options (e.g. number of XBUS cache blocks is greyed out if this specific module is not enabled). :+1:

grafik

Thank you very much for the rework!

stnolting (approved):

PR #1067, opened 2024-10-18
[vivado ip] make m_axi (XBUS) interface optional

donlon:
Hey thank you for merging my recent pulls! And here's the last split of #1060.

As mentioned in https://github.com/stnolting/neorv32/pull/1060#issuecomment-2420540728 did you encounter any error from the tool? It works fine for me on my computer with Vivado 2022.2.

### DIFF
--- a/rtl/system_integration/neorv32_vivado_ip.tcl
+++ b/rtl/system_integration/neorv32_vivado_ip.tcl
@@ -151,7 +151,7 @@ proc setup_ip_gui {} {
     { OCD_AUTHENTICATION    {OCD Authentication}    {Implement Debug Authentication module}                   {$OCD_EN} {$OCD_EN ? $OCD_AUTHENTICATION : false}}
   }
 
-  set group [add_group $page {External Bus Interface (XBUS)}]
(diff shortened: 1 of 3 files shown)


stnolting:
When I disable the XBUS interface the according input ports are set to all-zero. However, in my (auto-generated) VHDL top wrappers this is done in a "Verilog style" (using a literal) instead of using VHDL's `(others => '0')`:

grafik

This is one of the faulting lines:

```vhdl
m_axi_rresp => X"3",
```

`m_axi_rresp` is only 2-bits wide, but `X"3"` implies a 4-bit width.

Are you using a Verilog top wrapper?

stnolting:
I just removed the default values for AXI ports whose bit-width is not 1 or a power of two and this seems to fix this issue. 🚀

stnolting (approved):

stnolting:
This seems ready from my side. @donlon can you have another look at it?

donlon:
Oops forget to synthesize the generated core...My Vivado also generate the same wrapper that results in the width mismatch errors, and removing these default values does fix, which looks like a bug of Vivado. I think it's good and ready to go now!

stnolting:
Thanks for the feedback!

PR #1079, opened 2024-10-24
create neorv32_sigcount

victorlga:


stnolting:
Was this just a mistake, or did you want to make a proposal?

victorlga:
This was a mistake. I built a feature to make a little improve in software performance specifically for my college capstone project. It counts falling edges of a digital signal, I was doing it via external interrupts before. Now I can just read a memory address that already has the value counted.

stnolting:
> This was a mistake.

No problem!

> I built a feature to make a little improve in software performance specifically for my college capstone project. It counts falling edges of a digital signal, I was doing it via external interrupts before. Now I can just read a memory address that already has the value counted.

I had a quick look at your code - it looks great! Really good work! :+1:

PR #1096, opened 2024-11-19
use vhdl 2008 standard in ghdl simulations

csantosb:
Support for vhdl08 is quite good in current ghdl release.


csantosb:
I'm unable to reproduce the failed checks locally when using ghdl-clang/gcc. The problem arises with its ghdl-mcode variant, which operates differently and has more limitations, as in this case.

stnolting:
I think `--std=08` is just missing in the GHDL run command:

https://github.com/stnolting/neorv32/blob/457f902d117e07a895f0fdfe0000ad1fc978d3b3/sim/ghdl.run.sh#L48

stnolting:
This seems to work! 🚀

csantosb:
This is [correct](https://github.com/ghdl/ghdl/blob/master/doc/using/InvokingGHDL.rst?plain=1#L125).

stnolting:
> This is [correct](https://github.com/ghdl/ghdl/blob/master/doc/using/InvokingGHDL.rst?plain=1#L125). To consider all these differences between backends.

:+1:

> It's even possible to further simplify the main sim script, as there is no need to provide compilation order or location of sources in project. Let ghdl decide.

That's right. But I think it might be a good thing to keep that here (to check the file list is correct and to provide an application example for it).

csantosb:
> That's right. But I think it might be a good thing to keep that here (to check the file list is correct and to provide an application example for it).

Fine with me.

stnolting:
Would you restore the file-list stuff? Or do you think we should leave it out?

csantosb:
To me, the simpler, the better: ghdl is not using this list, anyway (which is produced by ghdl itself). Additionally, most tools today are able to figure out by themselves the correct order.

Finally, I'm interested in [packaging](https://git.sr.ht/~csantosb/guix.channel-electronics/tree/main/item/electronics/packages/fw.scm#L30) neorv32, including testing, which means neorv32-vunit would be part of neorv32, as it was the case in 1.10.6 (similar to [open-logic](https://github.com/open-logic/open-logic), for example).

stnolting:
> To me, the simpler, the better: ghdl is not using this list, anyway (which is produced by ghdl itself). Additionally, most tools today are able to figure out by themselves the correct order.

I think you've convinced me :+1:

Furthermore, we have a simple howto for the file lists in the documentation: https://stnolting.github.io/neorv32/#_file_list_files

stnolting (approved):

PR #1101, opened 2024-11-26
[docs/makefile] Copy figures also for single targets

LukasP46:
Hey all,
as I only want to generate html, I added

- the figures-copy as target 
- dependencies on it for the html targets
- targets for folders and dependencies on it

I would welcome if someone could test if I broke the other targets as I cannot evaluate it currently.


LukasP46:
I found a bug in the doxygen artifact: `neorv32_logo_riscv.png` and `neorv32_processor.png` are missing for the `md_README.html`. I'll fix it by copying all images to the acording folder, although there are some resources that will never be used. The other two methods would be to put the individual files in the makefile - no overhead, but needs to be maintained, or by automatically determining which images are needed - which seems overkill.

stnolting (approved):

PR #1107, opened 2024-11-28
[sw] Use build folder and add example for more complex project structure

LukasP46:
- Use `build` folder for all object files
- Use `sw/example/coremark` as example for using explicit `source` and `include` folder

Those changes should not break existing projects but I would welcome some testing as the object targets now use [VPATH search](https://www.gnu.org/software/make/manual/html_node/General-Search.html) using the `SRC` folders.

All main output files remain in the project's folder.


stnolting:
Thank you very much for the PR!

> I would welcome some testing

Works on my side.

> All main output files remain in the project's folder.

Great!

stnolting (approved):

PR #1108, opened 2024-11-28
[sw/bootloader] add TWI boot option

LukasP46:
Add ability to read program from TWI (also known as I2C) memory like common EEPROMs with one address byte.


stnolting:
> Add ability to read program from TWI

Nice! Thank you very much!

> The bootloader gets to big if all options are enables. 

Oh I know this problem... 🙈
How much more ROM do you need for your modified bootloader?

> How should we handle this? Maybe disable TWI by default but add an additional test where SPI is disabled instead?

I was thinking about rewriting the bootloader in plain assembly... But to be honest, I'm too lazy for that... 😅

So yes, we could disable the TWI option by default and let the user decide if this is a relevant feature.

LukasP46:
> How much more ROM do you need for your modified bootloader?

`Memory utilization:
   text    data     bss     dec     hex filename
   4456       0       8    4464    1170 main.elf`

> I was thinking about rewriting the bootloader in plain assembly... But to be honest, I'm too lazy for that... 😅

~~too lazy~~ not crazy
Surely it can be done, but it sounds really time consuming and rather hard to debug (and contribute to).

I actually use the C ISA extension so the bootloader is always small enough, but I guess I'll change the default and then the user can enable it if needed. I would actually suggest changing the warning as well and recommend recompiling the bootloader for each processor to enable a potentially smaller bootloader, what do you think?

LukasP46:
While writing to the TWI memory via the bootloader is not planned, I will probably upload a Python script to flash via an USB I2C adapter.

LukasP46:
> That's a good solution! But I would recommend to provide the pre-built bootloader image compiled with the minimal ISA options only - so it is compatible to any ISA configuration.

I agree, the pre-built should be compatible with any (but E currently) ISA configuration. But we could encourage the user to build the bootloader to their needs and ISA.

> Which warning?

This one in `bootloader.c`:
```
/**********************************************************************//**
 * Sanity check: Base RV32I ISA only!
 **************************************************************************/
#if defined __riscv_atomic || defined __riscv_a || __riscv_b || __riscv_compressed || defined __riscv_c || defined __riscv_mul || defined __riscv_m
  #warning In order to allow the bootloader to run on *any* CPU configuration it should be compiled using the base rv32i ISA only.
#endif
```

But actually I'm fine with it, because the first part says _to allow_, which actually implies _if you need it_...

> Btw, we have already enabled link-time-optimization and Os as compile-time switches. I am not sure if there is anything more we could do to further shrink the bootloader.

Probably not? [`-Oz`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-Oz) is even more aggressive, but is currently the same size as `-Os`.

> However, maybe it is worth a try to compile the bootloader with the absolute minimal ISA configuration which is rv32e. The E ISA extension only provides 16 x registers. Maybe the overhead of register spilling could be reduced by this?! 

I'll look into it. This would also have the advantage that the bootloader would actually work on all ISA configurations.

> I think we could re-use parts of the "SPI write" code?

Probably yes? Unfortunately I don't have a use case at the moment, but I keep it in mind. Probably not in this PR though. Some other bootloader extensions have a higher priority, you'll see :)

> I just scanned your code briefly, but I think you are using a fixed number of address bytes for accessing the TWI flash, right? Maybe we could also make this configurable just like we did for the SPI accesses.

I agree, I'll add that one!

stnolting:
> I agree, the pre-built should be compatible with any (but E currently) ISA configuration. But we could encourage the user to build the bootloader to their needs and ISA.

:+1:

> This one in bootloader.c:

Oh, right. I forgot about that...

I think we could remove that. It does not capture all possible ISA extensions. So basically it is useless.

> Probably not? [-Oz](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-Oz) is even more aggressive, but is currently the same size as -Os.

Interesting! I was not aware of that option. I'll look it up.

> I'll look into it. This would also have the advantage that the bootloader would actually work on all ISA configurations.

I just tested that and it saves more than 100 bytes (yeay!).

-> #1118

> I agree, I'll add that one!

Cool, thanks!

LukasP46:
Seems to work on the newest version as well :) only tested without clint/autoboot and only on the two-address-byte eeprom but I haven't noticed any relevant commits anyways.

stnolting:
This looks great, thank you very much!
I'll have a close look at the code (unfortunately, I do not have a compaitble I²C flash...).

However, I'm not sure about the EEPROM tool. This seems to be quite application-specific, right? 🤔

sw/bootloader/bootloader.c
```diff
@@ -103,6 +104,34 @@
...
+  #define TWI_CLK_PRSC CLK_PRSC_64
+#endif
+
+/** TWI Clock divider */
+#ifndef TWI_CLK_DIV
+  #define TWI_CLK_DIV 3
+#endif
+
+/** TWI First Slave ID */
+#ifndef TWI_SLAVE_ID
```
stnolting, line 126: I suggest renaming this to `TWI_DEVICE_ID` to match the phrases used by the other parts of the code.

stnolting (changes requested):

LukasP46:
> unfortunately, I do not have a compaitble I²C flash...

I suppose there could be FPGA models as well, but no additional testing would be required on my part.

> However, I'm not sure about the EEPROM tool. It is a cool feature and a handy tool, but this seems to be quite application-specific, right?

It kind of is. It should work with all kinds of I2C chips, but yeah, it's only for a weirdly specific USB adapter. I'll remove it from the branch then.

> I suggest renaming this to TWI_DEVICE_ID to match the phrases used by the other parts of the code.

I agree :)

LukasP46:
Now the renaming should have an end, sorry for alle the different commits :) feel free to squash

docs/userguide/customizing_the_bootloader.adoc
```diff
@@ -39,14 +39,23 @@ minimal base & privileged ISA `rv32e_zicsr_zifencei` only to ensure it can work
 | `SPI_BOOT_BASE_ADDR`    | `0x00400000`        | _any_ 32-bit value | Defines the _base_ address of the executable in external flash
 4+^| XIP configuration
 | `XIP_EN`                | `0` | `0`, `1`      | Set `1` to enable the XIP options
+4+^| TWI configuration
+| `TWI_EN`                | `0` | `0`, `1`      | Set `1` to enable the usage of the SPI module (including load/store executables from/to SPI flash options)
```
stnolting, line 43: The description should be "Set `1` to enable the usage of the **TWI** module (including loade executables from TWI flash option)" I think.

docs/userguide/customizing_the_bootloader.adoc
```diff
@@ -39,14 +39,23 @@ minimal base & privileged ISA `rv32e_zicsr_zifencei` only to ensure it can work
...
+| `TWI_CLK_PRSC`          | `CLK_PRSC_64` | `CLK_PRSC_2` `CLK_PRSC_4` `CLK_PRSC_8` `CLK_PRSC_64` `CLK_PRSC_128` `CLK_PRSC_1024` `CLK_PRSC_2024` `CLK_PRSC_4096`   | TWI clock pre-scaler (dividing main processor clock)
+| `TWI_CLK_DIV`  | `3` | `0` ... `15` | TWI clock divider (dividing twi clock)
+| `TWI_DEVICE_ID`  | `0x50` | `0x00` ... `0x7F` | First TWI device ID to start. Is incremented until the end of the program is reached, when `TWI_ADDR_BYTES` is `1`.
+| `TWI_ADDR_BYTES`  | `1` | `1`, `2` | TWI memory address size in number of bytes. When `TWI_ADDR_BYTES` is `1`, `TWI_DEVICE_ID` the gets incremented as well.
```
stnolting, line 47: That's a cool feature! :+1:

docs/userguide/executable_upload.adoc
```diff
@@ -123,6 +123,9 @@ See section https://stnolting.github.io/neorv32/#_bootloader[Bootloader] of the
 See section <<_programming_an_external_spi_flash_via_the_bootloader>> to learn how to use an external SPI
 flash for nonvolatile program storage.
 
+[TIP]
+The bootloader also supports booting from external TWI memory. Enable it in the bootloader makefile, but be caution, enabling all features might result in a too-big binary.
```
stnolting, line 127: typo: "but be careful"

sw/bootloader/bootloader.c
```diff
@@ -111,7 +140,8 @@
  **************************************************************************/
 enum EXE_STREAM_SOURCE_enum {
   EXE_STREAM_UART  = 0, /**< Get executable via UART */
-  EXE_STREAM_FLASH = 1  /**< Get executable via SPI flash */
+  EXE_STREAM_FLASH = 1, /**< Get executable via SPI flash */
+  EXE_STREAM_TWI   = 2  /**< Get executable via I2c Device */
```
stnolting, line 144: Let's also use the term "TWI" here

sw/bootloader/bootloader.c
```diff
@@ -122,18 +152,20 @@ enum ERROR_CODES_enum {
...
 /**********************************************************************//**
  * Error messages
  **************************************************************************/
-const char error_message[4][5] = {
+const char error_message[5][5] = {
   "EXE",
   "SIZE",
   "CHKS",
-  "FLSH"
+  "FLSH",
```
stnolting, line 167: Maybe we should change this to "SPI" to make the error more clear.... 🤔

stnolting (changes requested):

stnolting:
This looks great! I just have some minor comments. :wink:

By the way... How about also adding an option to store an image to TWI memory (similar to the SPI programming option)?

sw/bootloader/bootloader.c
```diff
@@ -331,7 +376,11 @@ int main(void) {
       }
 
       if (neorv32_clint_time_get() >= timeout_time) { // timeout? start auto boot sequence
-        get_exe(EXE_STREAM_FLASH); // try booting from flash
+        #if (SPI_EN != 0)
+          get_exe(EXE_STREAM_FLASH); // try booting from flash
+        #elif (TWI_EN != 0)
+          get_exe(EXE_STREAM_TWI); // try booting from twi
+        #endif
```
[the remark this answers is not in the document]
LukasP46: That's right for autobooting. 

I thought anyone with SPI flash would prefer it to any I2C memory, but this could be configurable. Just another `#define` with the desired `EXE_STREAM_SOURCE_enum`. And a check, I guess, if the selected option is available. What do you think?

LukasP46:
> This looks great! I just have some minor comments. 😉

Thanks! And also thanks for the suggestions.

> By the way... How about also adding an option to store an image to TWI memory (similar to the SPI programming option)?

I actually already answered this (https://github.com/stnolting/neorv32/pull/1108#issuecomment-2535221706), but no worries.  I don't have a use case right now, but I'm keeping it in mind. Probably not in this PR though.

Scutua (approved):

Scutua (approved):

LukasP46:
There might be problem with CLINT enabled, I need to run some tests...

sw/bootloader/bootloader.c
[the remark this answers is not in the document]
stnolting: > I thought anyone with SPI flash would prefer it to any I2C memory,

Good point. Besides that, you would adapt the configuration of the bootloader to the hardware setup anyway.

stnolting (approved):

LukasP46:
Maybe an additional Test would be nice where the bootloader get's compiled with the new options as there aren't as default?

docs/userguide/customizing_the_bootloader.adoc
Scutua: Thank you very much boss

PR #1112, opened 2024-12-03
Add UART disable tag

LukasP46:
Optional tag to reduce lib footprint if you know that UART is not synthesized. Especially with Link-Time-Optimization this can reduce the size of the binary significantly, especially as UART is used in the RTE.


stnolting:
Nice idea! Thanks for adding this.

stnolting (approved):

PR #1118, opened 2024-12-11
🧪 Shrink bootloader ISA and RAM requirements

stnolting:
The default bootloader is now compiled using the minimal `rv32e` ISA (`rv32_e_zicsr_zifencei` actually). Furthermore, the minimal RAM requirements are lowered to 256 bytes. Thus, the bootloader can also operate on tiny processor configurations. Additionally, these constraints also reduce the total bootloader image size.


docs/userguide/customizing_the_bootloader.adoc
```diff
@@ -12,7 +12,7 @@ bootloader ROM) and the processor has to be re-synthesized.
 
 [NOTE]
```
LukasP46, line 13: Maybe this is a good place to encourage the user to build the bootloader to their needs and ISA? (as mentioned here: https://github.com/stnolting/neorv32/pull/1108#issuecomment-2537021651). Especially if someone has the C extension enabled, it'll get even smaller and for other extensions the bootloader would just run faster.

Don't get me wrong, the default should be the minimal one and we should leave a note saying so!

LukasP46 (changes requested):

LukasP46:
And also I guess this PR is a good place to remove the `bootloader.c` sanity check as mentioned here: https://github.com/stnolting/neorv32/pull/1108#issuecomment-2537021651

Thanks for trying it out!

docs/userguide/customizing_the_bootloader.adoc
stnolting: Good point!

PR #1128, opened 2024-12-26
Update neorv32_cpu_control.vhd

mahdi259:
Hi, 
This change enables 64-bit counter to be implemented.

The former formula zeros hpm_cnt_hi_width_c when HPM_CNT_WIDTH equals to 64 and the actual bit width of implemented counter is 32 in that case. Thanks


stnolting:
Oh, thanks for the hint! I'll check that!

The rather complex formula we were using here was menat to avoid a negative value for the high-counter width. But I cannot remeber why this was an isuue. 🙈

mahdi259:
You're welcome.

stnolting (approved):

PR #1130, opened 2024-12-27
⚠️ Replace MTIME by CLINT

stnolting:
This PR replaces the MTIME machine timer by the RISC-V-compatible **core local interruptor (CLINT)**. From a functional point of view, both devices are backwards-compatible. However, base address and register offsets are changed. Additionally, the CLINT also provides an interface for triggering machine-level software interrupts.

The CLINT supports up to 4095 harts (CPU cores). So maybe it is finally time to think about a multi-core setup (like the one GitHub user @NikLeberg has created). 🤔

:warning: This PR also removes the processor `HART_ID` generic.


NikLeberg:
Wow! This looks nice. :)

I guess I have to rebase my (a bit dormant) SMP system.
If we'd like to think about a multi core system what are your thoughts how to progress? If you like I can sketch an architecture overview of my current implementation so that we could discuss points/entities that would need to be created or modified.

Cheers and happy new year! 🚀

stnolting:
> Wow! This looks nice. :)

Thanks a lot!

> If we'd like to think about a multi core system what are your thoughts how to progress? If you like I can sketch an architecture overview of my current implementation so that we could discuss points/entities that would need to be created or modified.

That's a good question. Let's open an issue or discussion for that. What worries me most is the multi-hart DM and of course the software support. 😅

> Cheers and happy new year! 🚀

A happy new year to you too!

PR #1135, opened 2024-12-30
✨ add smp dual-core support

stnolting:
Add SMP dual-core option.

* [x] add rtl infrastructure (two CPU instances + caches + bus arbiter)
* [x] adjust crt0 boot code
* [x] add HAL boot functions for secondary core
* [x] add dual-core example program
* [x] add dual-core test to CI
* [x] add dual-core openOCD configuration
* [x] test dual-core debugging
* [x] update documentation
* [x] update bootloader (image)

🧪 Note that the dual-core configuration is still experimental.


sw/lib/source/neorv32_rte.c
```diff
@@ -414,6 +414,69 @@ void neorv32_rte_debug_handler(void) {
...
+  // setup launch-configuration struct
+  __neorv32_rte_smp_startup.magic_word  = 0x1337cafeu;
+  __neorv32_rte_smp_startup.stack_lower = (uint32_t)stack_memory;
+  __neorv32_rte_smp_startup.stack_upper = stack_top;
+  __neorv32_rte_smp_startup.entry_point = (uint32_t)entry_point;
+
+  // flush data cache (containing configuration struct) to main memory
+  asm volatile ("fence");
+
+  // use CLINT.MTIMECMP[hart_sel].low_word to pass the address of the configuration struct
```
NikLeberg: Oooh I like it! Its hacky but smart. 😍
Where have you got the idea from?

rtl/core/neorv32_top.vhd
```diff
@@ -593,21 +606,21 @@ begin
...
-        host_req_i => cpu_d_req,
-        host_rsp_o => cpu_d_rsp,
-        bus_req_o  => dcache_req,
-        bus_rsp_i  => dcache_rsp
+        host_req_i => cpu_d_req(i),
+        host_rsp_o => cpu_d_rsp(i),
+        bus_req_o  => dcache_req(i),
+        bus_rsp_i  => dcache_rsp(i)
       );
     end generate;
```
NikLeberg, line 614: As we do not have any form of data coherency between the two cores, I'd suggest disabling the L1 data cache in dual-core mode. In such cases a shared (!) L2 cache like the XBUS cache should work ok enough.

rtl/core/neorv32_top.vhd
```diff
@@ -618,21 +631,52 @@ begin
...
+      PORT_A_READ_ONLY => false,
+      PORT_B_READ_ONLY => false
+    )
+    port map (
+      clk_i    => clk_i,
+      rstn_i   => rstn_sys,
+      a_lock_i => '0',
+      a_req_i  => core_req(core_req'left),
+      a_rsp_o  => core_rsp(core_req'left),
+      b_req_i  => core_req(core_req'right), -- [hack] core_req(1) does not exist if single core
```
NikLeberg, line 662: Is this comment still valid with the `if num_cores_c > 1 generate` condition?

sw/common/crt0.S
```diff
@@ -22,197 +22,224 @@ __crt0_entry:
...
+  lw    x1, 12(x13)                      // ra = __neorv32_rte_smp_startup.entry_point
+  sw    x0, 0(x13)                       // invalidate configuration struct
+  csrw  mstatus, x5                      // re-initialize mstatus
+
+  // acknowledge startup by clearing software interrupt
+  li    x14,  0xfff40000                 // CLINT.MSWI base address
+  sw    zero, 4(x14)                     // CLINT.MSWI[1]
+
+  // start main function
+  j     __crt0_main_entry
```
NikLeberg, line 115: Aren't we in the `mswi` exception here? So shouldn't we use something like below to actually return from the exception before we can call the provided main function in `x1`?
```asm
  la    x14,  __crt0_main_entry
  csrw  mepc, x14
  mret
```

sw/example/demo_dual_core/main.c
```diff
@@ -0,0 +1,202 @@
...
+}
+
+
+/**********************************************************************//**
+ * CLINT machine timer interrupt handler for core0.
+ **************************************************************************/
+void clint_mtime_handler_core0(void) {
+
+  spin_lock();
+  neorv32_uart0_printf("[core0] Primary core 1-second MTIMER interrupt. SMP is so cool!\n");
```
NikLeberg, line 181: I agree! :)

sw/example/demo_dual_core/spinlock.c
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file spinlock.c
+ * @brief Single simple spin-lock based on atomic lr/sc operations.
+ */
+#include <neorv32.h>
+
+/**********************************************************************//**
+ * Private spinlock locked variable. We can only use a single spinlock
+ * as the processor only features a single reservation set.
```
NikLeberg, line 9: If my understanding of this is correct, then we aren't really limited by having only a single reservation set. The hardware doesn't care. We just have to write correct software for it. The only thing that I think might happen is, if both cores try to atomically change anything via `lr/sc`, they may end up in livelock as they reset the reservation for each other.
But if we design the spinlock with exponential (or random) backoff (on lock failure) then I think we can have more than one spinlock, can't we?

NikLeberg (commented):
I left some small remarks. The biggest issue I see is with the L1 data caches. I think we should disable them automatically (and emit a warning) even if the user enabled them.

sw/lib/source/neorv32_rte.c
stnolting: Thank you, but I think that's actualy the only option we have. We cannot use RAM to pass the address of the configuration structure as it would have to be "hardcoded" into the crt0 code - at least for the bootloader. But when compiling the bootloader the address is not known as the bootloader itself does not use the dual-core setup.

So we need an actual register that is accessible by both core and that has a **fixed address**. The CLINT's timer compare register was the only option:
* it is **guaranteed** to be available in dual-core configuration
* using this register does not have any side-effects (or at least they are not critical)

However, I don't like this because it is counter-intuitive. A dedicated "SMP communication module" might be better... 🤔

rtl/core/neorv32_top.vhd
stnolting: That is really a problem and. But I think we should leave that to the user.

rtl/core/neorv32_top.vhd
stnolting: The first version used hard-coded indices:

```vhdl
a_req_i  => core_req(0),
a_rsp_o  => core_rsp(0),
b_req_i  => core_req(1),
b_rsp_o  => core_rsp(1),
```

Obviously, this did not work as index 1 does not exist. The whole generate statement won't be synthesized because `if num_cores_c > 1 generate` is false, but linting of the code still failed.

sw/common/crt0.S
stnolting: That's right, we are "in" a trap handler here. But that should not hurt - we just need to re-initialize `mstatus` before launching `main`, right? 🤔

sw/example/demo_dual_core/main.c
stnolting: 😎👍

sw/example/demo_dual_core/spinlock.c
stnolting: I also racked my brains about that... 😅
I've tried to use two individual spinlock variables based on the same LR/SC primitives and ended up in a livelock.

> But if we design the spinlock with exponential (or random) backoff (on lock failure)

How would you do that? 🤔

stnolting:
@NikLeberg Thank you so much for your feedback!

The current design works - at least for a simple dual-core demo program. However, some things are not really efficient and quite hard to handle for a user. But I think this is fine for a first version.

Later, we should perhaps add a special module to avoid the problems (mentioned above). The RP2040 provides a cool block that is used for inter-cpu communication. It provides some nice features:

* hardware spinlocks
* FIFOs for passing messages from one CPU to another

grafik

Source: [RP2040 data sheet](https://files.seeedstudio.com/wiki/XIAO-RP2040/res/rp2040_datasheet.pdf)

The rp2040 provides interrupts for the communication FIFOs. I think that is not necessary as we have the software interrupts (and we are running out of interrupt channels anyway).

sw/common/crt0.S
NikLeberg: > we just need to re-initialize `mstatus`

Ah I see, you mean the line 
```as,
csrw  mstatus, x5                      // re-initialize mstatus
```

Not sure. You know it best if the CPU is in any weird state because of the exception context. 😅

For me, exiting the exception seems cleaner. 🤔

stnolting: > For me, exiting the exception seems cleaner.

Maybe you're right... 🤔 But we can fix/improve this after merging I think.

PR #1153, opened 2025-01-13
[sw/bootloader] aligned trap handler

LukasP46:
This fixes the bootloader to be compatible with the c isa extension. It took a lot of debugging to figure out why the bl sometimes worked and why it did not ;)


stnolting:
Thanks for fixing!

For the records:
The two lowest bits of the `mtvec` CSR (address of the handler for all traps) are used for configuration. Hence, the trap handler address has to be 4-byte aligned.

PR #1158, opened 2025-01-15
[rtl] TWD idle when NoACK during reading

### LINKED ISSUE #1157: TWD going into idle depends on TX Fifo content
First of all, I am still working out the details and may find a solution along the way.

Situation: I read (one or more) bytes from the Neorv32 as TWD.

It depends on the last read value if the twd engine.state returns to idle. It seems that the MSB of the value has to be '1' to succeed, but I haven't tested all possibilities.

The corresponding C code is really simple:
  neorv32_twd_setup(TWD_DEVICE_ID, 0, false, false, false);
  neorv32_twd_put(0x06);
  neorv32_twd_put(0x81); // won't work if < 0x80

And I just read the two bytes from a host.

I already see that the stop signal isn't noticed anymore. Like I said, I'll look into it myself, but now it's noted here for everyone to read.

LukasP46:
Fixes #1157 and to my understanding should be TWI compliant.

### DIFF
--- a/rtl/core/neorv32_twd.vhd
+++ b/rtl/core/neorv32_twd.vhd
@@ -412,7 +412,10 @@ begin
           engine.cnt  <= (others => '0');
           engine.sreg <= engine.rdata; -- FIFO TX data
           engine.dout <= engine.rdata(7); -- FIFO TX data (first bit)
-          if (ctrl.enable = '0') then -- disabled?
+          if (
+            (ctrl.enable = '0') or  -- disabled?
+            ((engine.cmd = '1') and (smp.scl_rise = '1') and (smp.sda = '1') ) -- or no ACK while read?
+          ) then
             engine.state <= S_IDLE;
           elsif (smp.scl_fall = '1') then
             engine.state <= S_RTX;


LukasP46:
Explanation: this prevents the MSB to pull the SDA line down and thus preventing the stop command to be recognized.

stnolting:
Ah I see. So we should also check for a STOP condition in the ACK time slot? Shouldn't we do this for read **and** write accesses?

LukasP46:
I guess checking for the stop anytime makes sense, but this problem might occur anyway since in my example the stop signal comes afterwards:
image
(Green with the patch, yellow without)

rtl/core/neorv32_twd.vhd
```diff
@@ -412,7 +412,10 @@ begin
           engine.cnt  <= (others => '0');
           engine.sreg <= engine.rdata; -- FIFO TX data
           engine.dout <= engine.rdata(7); -- FIFO TX data (first bit)
-          if (ctrl.enable = '0') then -- disabled?
+          if (
+            (ctrl.enable = '0') or  -- disabled?
+            ((engine.cmd = '1') and (smp.scl_rise = '1') and (smp.sda = '1') ) -- or no ACK while read?
```
stnolting, line 417: Basically, cou are checking for a STOP condition here, right?
Can't we use the `smp.stop` signal?

```vhdl
((engine.cmd = '1') and (smp.stop = '1')) -- STOP while read?
```

stnolting:
Something isn't right with the TWD... 🙈
I'll take a close look at this tonight.

rtl/core/neorv32_twd.vhd
LukasP46: No, as the stop signal is a SDA rise with SCL being one and not the other way around. This is more like reading the 9th bit: is the ack set or not? If not => I'm done (technically waiting for the stop), if yes => I send the next byte.

stnolting:
I think there are several design flaws in the TWD 🙈
https://github.com/stnolting/neorv32/pull/1161 tries to fix them.

LukasP46:
Works with #1161, thanks!

stnolting:
Great to hear and thanks for the feedback!

PR #1162, opened 2025-01-18
[rtl/xbus2axi4lite_bridge] fix back-to-back wishbone transfers

donlon:
Hi!

This PR fixes the converter when wishbone transfer is back-to-back. When the stb signal is continuously high, the ready signal won't be reset and causes arvalid/awvalid pulse of the next transfer disappear (in the following pic). rvalid and bvalid condition are added to ensure the handshake signals are correctly generated.

wb2axi4

### DIFF
--- a/rtl/system_integration/xbus2axi4lite_bridge.vhd
+++ b/rtl/system_integration/xbus2axi4lite_bridge.vhd
@@ -68,9 +68,9 @@ begin
     if (resetn = '0') then
       ready <= (others => '0');
     elsif rising_edge(clk) then
-      ready(0) <= xbus_cyc_i and (ready(0) or std_ulogic(m_axi_arready));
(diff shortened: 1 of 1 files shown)


stnolting:
What a perfect timing - I was just working on something similar 😅

https://github.com/stnolting/neorv32/pull/1163 also adds back-to-back AXI operations (dedicated for atomic read-modify-write operations). Maybe you can check that with your setup (so we can potentionally merge your edits)?

> When the stb signal is continuously high

Do you mean `cyc`? The current version of the AXI brdige does not use `stb` at all.

> rvalid and bvalid condition are added to ensure the handshake signals are correctly generated.

I just had a quick look at your code. Thanks for the PR!

However, I think your changes could inadvertently trigger multiple AXI read accesses to the same address?! 🤔

stnolting:
Hey @donlon.

I think the changes from #1163 / the current AXI4-Lite bridges already fixes this, right? 🤔

PR #1165, opened 2025-01-21
[rtl] use past SDA in twd FSM transition

LukasP46:
Since the SDA has to have a valid value only during high SCL, I had the case where with the fall of SCL the SDA also changed and thus a wrong state transition happened. But the protocol should guarantee a valid input during high SCL, so this should be a valid fix?

The image shows such a transition, which would fail with the old implementation.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 22.01.2025 | 1.10.9.10 | :bug: fix TWD ACK/NACK sampling | [#1165](https://github.com/stnolting/neorv32/pull/1165) |
 | 18.01.2025 | 1.10.9.9 | atomic memory access updates and improvements | [#1163](https://github.com/stnolting/neorv32/pull/1163) |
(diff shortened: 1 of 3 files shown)


stnolting:
You are right. SDA gets tested here on the falling edge of SCL which is not compliant to the spec (I think)...

However, checking `smp.sda_sreg(2)` might not be a good idea as this represents the state of SDA from one (processor clock!) cycle before.

LukasP46:
> However, checking smp.sda_sreg(2) might not be a good idea as this represents the state of SDA from one (processor clock!) cycle before.

Spec wise this would be fine "The data on the SDA line must be stable during the HIGH period of the clock." (3.1.3 [UM10204
I2C-bus specification and user manual](https://www.nxp.com/docs/en/user-guide/UM10204.pdf)) and even one clock period before the fall should be fine.

But it is not so transparent because it depends on the system clock. So your solution is cleaner and only costs one more bit on the ctrl register...

> Feel free to revert my changes if this does not work for your test case.

Works just fine, thanks a lot!

> Could you then please provide a minimal test case to reproduce this ACK "race condition"?

Unfortunately I use parts of a library I cannot share, but the tricky part was that the SDA has a transition to '1' at the very same moment the SCL falls.

stnolting:
> Spec wise this would be fine "The data on the SDA line must be stable during the HIGH period of the clock." (3.1.3 UM10204 I2C-bus specification and user manual) and even one clock period before the fall should be fine.

This should be the case now.

> Works just fine, thanks a lot!

Great to hear! Thank you for testing!

PR #1167, opened 2025-01-23
[rtl] reset SDA and SCL of TWI and TWD to '1'

LukasP46:
Should improve compatibly with other TWI devices on the same bus as the reset doesn't pull-down SCL and SDA anymore.


stnolting:
That's a nice idea! :+1:

Btw, I think we could even remove the SCL driver logic from the TWD module. It does not support clock stretching so it will never ever actually drive the SCL line. Anyway, that might be a cleanup thing for another PR.

stnolting (approved):

PR #1175, opened 2025-02-01
⚠️ remove execute in-place (XIP) module

stnolting:
The XIP module looks like an over-engineered module, but due to its low performance (only SPI, no QSPI) it does not provide any direct added value. The XIP module is therefore removed with this PR. Perhaps it will be moved to a new repository as an independent component.


jpf91:
It's a bit sad to see the XIP module gone :cry: 

XIP is probably not that useful on FPGAs, where you usually have enough memory. But if someone would want to tape out a NEORV32 ASIC, SRAM is quite expensive, so I'd imagine XIP can be useful there. Another reason for XIP would be using very large applications, basically whenever you have more instructions than you would configure ICACHE.

stnolting:
That's true, but the maintenance effort was simply too great. Apart from that, the XIP module only supported SPI (i.e. single-bit transmission). A QSPI or HyperRAM variant would be significantly more powerful. But they would then be even more complex to maintain... 🙈

The processor offers three different bus interfaces (Wishbone, AXI4-Lite, AHB3-Lite). I thought that if someone needs XIP, it would be better to connect an external (open-source) variant. I, for example, use the module from the Vivado IP catalog.

As far as ASICs are concerned, you are of course doubly right.

jpf91:
Hmm I see. One of the best things with NEORV is that it provides a completely integrated package, with SoC peripherals, great documentation and drivers. So having XIP integrated was quite nice.

OTOH if it's too much effort to maintain, it's completely understandable to remove it. If moving it to a new repository is an option, having the code still available would be highly appreciated :-)

stnolting:
> So having XIP integrated was quite nice.

I also like this "all-in-one" concept. 

Maybe I'll manage to completely rework the XIP module. QSPI mode would be great and write accesses would also be cool (using PSRAM as data memory).

> If moving it to a new repository is an option, having the code still available would be highly appreciated :-)

That is the plan. The same XIP modules only with a Wishbone interface. Let's see when I get around to it. 😅

PR #1189, opened 2025-02-18
Fix Bootloader Makefile UART_BAUD

lebruu:
Fix Bootloader Makefile, USER_FLAG for setting the UART Baudrate was wrong


stnolting:
Thanks for fixing!

stnolting (approved):

PR #1196, opened 2025-03-04
rtl: processor_templates: enable Zicntr ISA extension on minimal templates

henrikbrixandersen:
Re-enable the Zicntr ISA extension on the neorv32_ProcessorTop_Minimal and neorv32_ProcessorTop_MinimalBoot templates to restore RISC-V compliance.

This allows for an easier path for evaluating 3rd party software depending on Zicntr (e.g. the Zephyr RTOS).

### DIFF
--- a/rtl/processor_templates/neorv32_ProcessorTop_Minimal.vhd
+++ b/rtl/processor_templates/neorv32_ProcessorTop_Minimal.vhd
@@ -51,6 +51,8 @@ begin
     CLOCK_FREQUENCY              => CLOCK_FREQUENCY,   -- clock frequency of clk_i in Hz
     -- Boot Configuration --
     BOOT_MODE_SELECT             => 2,                 -- boot from pre-initialized interal IMEM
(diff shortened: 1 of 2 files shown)


stnolting:
> This allows for an easier path for evaluating 3rd party software depending on Zicntr (e.g. the Zephyr RTOS).

Good point.

> [...] restore RISC-V compliance

Just for the records: in the most recent version of the ISA spec `Zicntr` is no longer listed as mandatory extension. ;)

stnolting (approved):

henrikbrixandersen:
> Just for the records: in the most recent version of the ISA spec `Zicntr` is no longer listed as mandatory extension. ;)

I was not aware of that, appreciate the heads-up.

stnolting:
`Zicntr` is _recommended_. I at least think that means that it is not mandatory. 😅 

grafik

Source: https://github.com/riscv/riscv-isa-manual/blob/main/src/counters.adoc?plain=1#L10-L28

PR #1197, opened 2025-03-05
Fix: quote readlink to fix windows make check

ecstrema:
`@readlink -f ${SHELL}` means `@readlink -f C:\Program Files\Git\usr\bin\sh.exe` on windows, so we need to escape the path to account for the space


stnolting:
Thanks for fixing!

stnolting (approved):

PR #1201, opened 2025-03-09
fix: increment location counter with heap size

brkydnc:
The heap section **sets** the location counter to `__neorv32_heap_size` instead of incrementing it. 

I discovered this one when `rust-lld` failed with an error while creating a runtime support crate for neorv32 in Rust.


stnolting:
Oh, your are right. Seems like a stupid typo... 🙈
Thanks for finding and fixing this!

stnolting (approved):

Issue #1199, opened 2025-03-05
More than two cores


pepijndevos:
I'm quite interested in making a SoC with more than two cores.

Looking at the code a bit it seems like maybe all that would be needed is to generalize the bus switch for more than two cores?

Maybe inter-core communication would require some thought as well.

stnolting:
Hey @pepijndevos!

> I'm quite interested in making a SoC with more than two cores.

The simplest approach would be to instantiate several NEORV32 tops, connect them to a global memory and daisy-chain the JTAG chain. This would then no longer be an SMP architecture, but rather a NUMA architecture.

> Looking at the code a bit it seems like maybe all that would be needed is to generalize the bus switch for more than two cores?

Basically, you are right. The on-chip-debugger's debug module already supports up to 4 CPU cores (but that could be easily extended):

https://github.com/stnolting/neorv32/blob/8ab1fecf6832164c4c97c8c61c67afd60369ef2d/rtl/core/neorv32_debug_dm.vhd#L21-L23

The core-local interrupt supports even more CPU cores:

https://github.com/stnolting/neorv32/blob/8ab1fecf6832164c4c97c8c61c67afd60369ef2d/rtl/core/neorv32_clint.vhd#L21-L23

> Maybe inter-core communication would require some thought as well.

Right. That would require a more complex fully-connected archiitecture since every core needs to be able to communicate with every other core.

----------

When we introduced dual-core support, I did a few experiments with even more cores.

Two cores (without any CPU caches) surprisingly rarely get in each other's way (in terms of bus congestion). I think this is due to the fact that the CPU is a multi-cycle architecture that does not perform bus accesses in each and every cycle. With four cores, however, there was already a significant congestion in the central bus system.

With bigger/better CPU caches, however, you could certainly compensate for some of this.

stnolting:
I think this can be closed.

PR #1204, opened 2025-03-11
Add NUMA LiteX configuration

pepijndevos:
As discussed in #1199 one way to make a SoC with more than two cores would be to make more of a NUMA architecture with several instances of the top level CPU.

I had that more or less working on 1.9.7 that LiteX was using, but noticed that's actually quite outdated.
Applying the same changes to the latest version is somehow not as functional.

I'll put up the LiteX changes in a second, maybe someone more familiar with these two projects can have a look.

### DIFF
--- a/rtl/core/neorv32_package.vhd
+++ b/rtl/core/neorv32_package.vhd
@@ -788,6 +788,7 @@ package neorv32_package is
     generic (
       -- Processor Clocking --
       CLOCK_FREQUENCY       : natural                        := 0;
+      HART_BASE             : natural                        := 0;
       -- Dual-Core Configuration --
       DUAL_CORE_EN          : boolean                        := false;
       -- Boot Configuration --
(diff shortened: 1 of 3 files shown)


stnolting:
This looks great! Thanks for adding this option.

stnolting (approved):

pepijndevos:
Here is the WIP LiteX code: https://github.com/enjoy-digital/litex/pull/2204

Was able to run multicore but after the update hangs the simulator.

Also haven't done the JTAG chaining that you suggested.

pepijndevos:
That was fast! I assumed you'd want LiteX to work before merging. Hopefully we can figure out why it hangs after the update. Have there been any breaking changes between the two versions?

stnolting:
> That was fast! I assumed you'd want LiteX to work before merging. 

I thought that would be easier if the changes were already upstream. Also, your modifications should all be "backwards compatible".

> Have there been any breaking changes between the two versions?

That's a good question... Which reference version are you referring to?

One of the most important differences is that the base addresses of all peripheral modules have changed. The entire software must therefore be adapted or recompiled.

pepijndevos:
LiteX was on 1.9.7
If the memory regions are wrong that could definitely explain the simulation hang. For all I know it could be booting but fail to write to UART.
Idk how involved you are on the LiteX side but any help welcome ofc. I'll get back to it next week due to some other projects and the baby ;)

PR #1205, opened 2025-03-14
[rtl/sw] Add twd dummy byte, add twd bootloader and smaller fixes

LukasP46:
Moved from #1173.

- Add twd dummy register to allow continuous reading of same byte as host when enabled
- Add hide read option to only ACK on Address when TX Fifo is not empty
- Always NACK during the transfer, if the FIFO is full
- Separate TWD TX and RX Fifo depth generics
- Add twd option to bootloader

Tested only on release 1.10.9


Unike267:
Hi!

Please don't take this the wrong way, but in my opinion, each change should be in a separate commit with a clear message that describes it.

[How write a good commit message](https://github.com/torvalds/subsurface-for-dirk/blob/master/README.md#contributing)

Eighteen files changed, with several modifications in each, all in a single commit, makes for a very messy workaround.

I wanted to review the pull request but didn’t proceed because it was too chaotic.

Sometimes commit messages are almost as important as the code changes themselves.

I want to clarify that I am not belittling your work at all. On the contrary, open source is built through the continuous contributions of the community.

I hope you take my constructive criticism positively. :heart: 

Cheers!

LukasP46:
Hi there!

> Each change should be in a separate commit with a clear message that describes it.

I totally agree, and usually do. Unfortunately, I am currently using an old version of the processor and have to create an extra branch with the changes for the upstream. While this may change in the future, it would currently take some time to put each change into a separate commit. That's what I would do if the changes documented here didn't exist: #1173.

But I have to admit that this is mostly useful for @stnolting and me, and not really for reviewing this standalone. I create a new branch with separate commits if @stnolting agrees that this is too messy.

Unike267:
Thanks so much, Lukas!

I thought about not saying anything in case you got upset :sweat_smile: , but I really wanted to see/understand your improvement.

stnolting:
> Please don't take this the wrong [...]

I think that's a good point. Even if I have to admit that I do exactly the same thing far too often myself... 🙈

Anyway, you're right. Sometimes things are changing here pretty fast and also with brekaing changes. So it's quite helpful if the commit messages give an overview of what has actually changed (and why).

So it would be great if you could split the changes across multiple commits. Individual PRs are not necessary in my opinion.

Issue #640, opened 2023-06-28
feature request - Zb* sub-extensions


biosbob:
the `Zbc` sub-extension (carry-less multiplication) is quite different from the other sub-extensions, which really do "bit manipulation" in the more traditional sense...

not critical, but it might be helpful to enable the `Zb*` sub-extensions individual -- as this also has impact on the overall size of the synthesized core....

a placeholder for the future....

stnolting:
For the records: this is being discussed in https://github.com/stnolting/neorv32/issues/633#issuecomment-1604684303

biosbob:
#633 talks about the `Zc*` sub-extensions; this issue concerns the `Zb*` sub-extensions....

perhaps the data-sheet could not only list the extensions currently supported, but those on (or NOT on) the roadmap....

stnolting:
Oh sorry, seems like I have mixed up something 😅

`B` vs. `Zb*` is something I need to take care of because the `B` extensions basically does not exist (anymore): https://github.com/riscv/riscv-bitmanip/issues/183

So we should replace `CPU_EXTENSION_RISCV_B` by several `CPU_EXTENSION_RISCV_Zb*` generics to fine tune the bitmanip extension. Actually, this fine-grained tuning is already implemented:

https://github.com/stnolting/neorv32/blob/a5bf1c822ff02482029f3aa93e8ca07e91a45a77/rtl/core/neorv32_cpu_cp_bitmanip.vhd#L74-L79

But theses switches are not propagated to the top nor to the control unit to check for legal/illegal instructions.

ecstrema:
I've just tried using zbb, and couldn't have it work:

code:

```c
  uint32_t trailing_zeros;
  asm volatile("ctz %0, %1" : "=r"(trailing_zeros) : "r"(mask));
  neorv32_uart0_printf("%d\n", trailing_zeros);
```

ecstrema:
Enabled extensions: 

```vhdl
      -- RISC-V CPU Extensions --
      RISCV_ISA_C      => true, -- Compressed ISA (16-bit instructions)
      RISCV_ISA_M      => true, -- Integer Multiplication and Division
      RISCV_ISA_U      => true, -- Integer base ISA
      RISCV_ISA_ZICNTR => true, -- Base counters extension
      RISCV_ISA_Zbb    => true, -- Bit manipulation extension (used to count trailing bits)
```

Error printed to the console:

```
[...]

PR #1206, opened 2025-03-14
:bug: Fix `Zbb` shift instructions

stnolting:
The shift operations of the `Zbb` instructions caused a permanent ALU stall (and thus, a timeout leading to an illegal instruction exception) when `CPU_FAST_SHIFT_EN` and `RISCV_ISA_Zbkc` are both disabled.

Bug discovered by @ecstrema in https://github.com/stnolting/neorv32/issues/640#issuecomment-2724946979.


ecstrema (approved):
Makes sense. It says in S_BUSY state while slmul.run is '1'.

ecstrema:
I wonder why it was set to '0' at first then...

PR #1209, opened 2025-03-16
docs: datasheet: cpu: fix name for the "A" ISA extension

henrikbrixandersen:
Fix the name for the "A" (Atomic memory instructions) ISA extension in the instruction set extensions table.


stnolting:
Thanks for fixing!

stnolting (approved):

PR #1212, opened 2025-03-19
build: rebuild exe when header files change

ecstrema:
Because the headers were not declared as dependencies of the elf, make was not detecting changes to them. This adds an optional configuration option for header files (APP_HEADERS), that makes make rebuild the executables when the header files change.


ecstrema:
We might also want to add the line

```makefile
#APP_HEADERS += $(wildcard ./*.h)
```

to the examples.

ecstrema:
A better way to do this might be to use

```makefile
	@$(CC) $(NEO_CFLAGS) -I $(NEORV32_INC_PATH) $(APP_INC) -MMD -MP -MF $(BUILD_DIR)/$*.d -MT $*.c.o $<
```

to generate a dependencies file. This dependencies file tells which c files depends on which .h file, and we can then have make recompile the c file only when the headers it depends on have changed.

However, I think this is currently overkill, and the current way works fine.

ecstrema:
I must also admit that my knowledge of make is not sufficient to do it that way without losing too much time.

ecstrema:
I added a second commit which fixes the second part of the debug output: 

```
File '../../libs/neorv32/sw/image_gen/image_gen' does not exist.
  Must remake target '../../libs/neorv32/sw/image_gen/image_gen'.
Compiling image generator...
  Successfully remade target file '../../libs/neorv32/sw/image_gen/image_gen'.
```

stnolting:
Great idea! Thanks for adding this!

Could please also add `APP_HEADERS` to the makefile's documentation (https://github.com/stnolting/neorv32/blob/main/docs/datasheet/software.adoc#14-application-makefile)?

sw/common/common.mk
```diff
@@ -121,6 +121,9 @@ CC_HOST = gcc -Wall -O -g
 
 # NEORV32 executable image generator
 IMAGE_GEN = $(NEORV32_EXG_PATH)/image_gen
+ifeq ($(OS),Windows_NT)
```
stnolting, line 124: :+1:

stnolting:
Ah right, I did not see that.

Looks good to me! :+1:

stnolting (approved):

Issue #1221, opened 2025-04-01
PWM prescaler value ignored


henrikbrixandersen:
**Describe the bug**
The value of the 3-bit PWM clock prescaler is ignored. PWM frequency only depends on fmain and the 10-bit PWM clock divider value. 

This is the formula for calculating the PWM frequency given in the datasheet: 
> fPWM[Hz] = fmain[Hz] / (2^8 * clock_prescaler * (1 + PWM_CFG_CDIV))

This seems to be the formula matching the implemented functionality:
> fPWM[Hz] = fmain[Hz] / (2^8 * 1 * (1 + PWM_CFG_CDIV))

**To Reproduce**
Running the UP5kdemo on the iCEBreaker FPGA board with Fmain set to 18 MHz, PWM CDIV set to 0 result in a PWM frequency of ~70.3 kHz regardless of the 3-bit PWM prescaler value.

This can be reproduced by setting the relevant PWM channel configuration register to 0x80000080.

**Expected behavior**
The 3-bit PWM prescaler value should take effect.

**Screenshots**

**Environment:**
 - OS: Ubuntu Linux
 - GCC Version: riscv64-zephyr-elf-gcc (Zephyr SDK 0.17.0) 12.2.0
 - Libraries used: Zephyr, picolibc

**Hardware:**
 - Hardware version: 1.11.2.0
 - Implemented CPU extensions, peripherals: `rtl/processor_templates/neorv32_ProcessorTop_UP5KDemo.vhd`
 - Hardware modifications: None

**Additional context**
Found while implementing a Zephyr RTOS PWM driver for the NEORV32.

PR #1222, opened 2025-04-01
:bug: fix PWM prescaler

stnolting:
Fixing #1221 by @henrikbrixandersen


henrikbrixandersen:
@stnolting Thanks! I can confirm the fix is working here.

stnolting:
@Unike267 

Yep. AI is now everywhere... But (un)fortunately Copilot does not understand HDL (yet). 😅

@henrikbrixandersen

Great to hear! Thanks for the feedback!

PR #1225, opened 2025-04-05
Minor rtl edits and optimizations

stnolting:
* a bus access exception will be raised when accessing unavailable CLINT registers
* fix typo in XBUS-to-AHB-Lite bridge rtl code
* rework CPU front-end code
* code cleanups

### DIFF
--- a/rtl/core/neorv32_clint.vhd
+++ b/rtl/core/neorv32_clint.vhd
@@ -96,7 +96,6 @@ architecture neorv32_clint_rtl of neorv32_clint is
   -- misc --
   signal mtime : std_ulogic_vector(63 downto 0);
   signal ack_q : std_ulogic;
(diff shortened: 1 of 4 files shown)


rtl/core/neorv32_clint.vhd
```diff
@@ -198,13 +197,12 @@ begin
     if (rstn_i = '0') then
       ack_q <= '0';
     elsif rising_edge(clk_i) then
-      ack_q <= bus_req_i.stb;
+      ack_q <= mtime_en or or_reduce_f(mtimecmp_en) or or_reduce_f(mswi_en);
```
NikLeberg, line 200: Does this now rely on the timeout feature of the bus to terminate the transaction with `err`?

stnolting: That's right. 😉

PR #1228, opened 2025-04-08
[sw/lib/neoled] fix neoled "irq_mode" not a applied in "neorv32_neoled_setup" and [sw/lib/neoled] updated doxygen comments

SirBramble:
- Fixed "NEOLED_CTRL_IRQ_CONF" bit never being set in the "neorv32_neoled_setup" function.
- Updated corresponding doxygen comments

"irq_mode" parameter was shifted to the "NEOLED_CTRL_EN" position and thus rendered ineffective. This caused the "NEOLED_CTRL_IRQ_CONF" bit to remain 0 at all times, causing the interrupt to always trigger on an empty FIFO.

Tested with current main branch up to commit e1283bc. Interrupt is triggered on both an empty and (less than) half full FIFO.

### DIFF
--- a/sw/lib/include/neorv32_neoled.h
+++ b/sw/lib/include/neorv32_neoled.h
@@ -62,7 +62,7 @@ enum NEORV32_NEOLED_CTRL_enum {
   NEOLED_CTRL_T_ONE_H_3  = 23, /**< NEOLED control register(23) (r/w): pulse-clock ticks per ONE high-time bit 3 */
   NEOLED_CTRL_T_ONE_H_4  = 24, /**< NEOLED control register(24) (r/w): pulse-clock ticks per ONE high-time bit 4 */
 
-  NEOLED_CTRL_IRQ_CONF   = 27, /**< NEOLED control register(27) (r/w): TX FIFO interrupt: 1=IRQ if FIFO is empty, 1=IRQ if FIFO is less than half-full */
+  NEOLED_CTRL_IRQ_CONF   = 27, /**< NEOLED control register(27) (r/w): TX FIFO interrupt: 0=IRQ if FIFO is empty, 1=IRQ if FIFO is less than half-full */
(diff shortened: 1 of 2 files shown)


stnolting:
I've never used the NEOLED interrupt in an actual application.
But the processor check should actually test that?! 🤔

Anyway, thanks for finfing and fixing this bug!

stnolting (approved):

SirBramble:
> But the processor check should actually test that?! 🤔

It does. But only kind of 😅
It only checks for a empty FIFO interrupt trigger, but not for a half full FIFO interrupt trigger.

```c
// ----------------------------------------------------------
// Fast interrupt channel 9 (NEOLED)
// ----------------------------------------------------------
neorv32_cpu_csr_write(CSR_MCAUSE, mcause_never_c);
PRINT_STANDARD("[%i] FIRQ9 (NEOLED) ", cnt_test);

if (NEORV32_SYSINFO->SOC & (1 << SYSINFO_SOC_IO_NEOLED)) {
  cnt_test++;

  // enable fast interrupt
  neorv32_cpu_csr_write(CSR_MIE, 1 << NEOLED_FIRQ_ENABLE);

  // configure NEOLED, IRQ if FIFO  empty  <===========================
  neorv32_neoled_setup(CLK_PRSC_4, 0, 0, 0, 0);

  // send dummy data
  neorv32_neoled_write_nonblocking(0);
  neorv32_neoled_write_nonblocking(0);

  // sleep until interrupt
  asm volatile ("nop");
  asm volatile ("nop");

  neorv32_cpu_csr_write(CSR_MIE, 0);

  if (neorv32_cpu_csr_read(CSR_MCAUSE) == NEOLED_TRAP_CODE) {
    test_ok();
  }
  else {
    test_fail();
  }
}
else {
  PRINT_STANDARD("[n.a.]\n");
}
```

PR #1229, opened 2025-04-11
[sw/bootloader] TWI fix for bootloader

SirBramble:
## Fixes
- fixed bootloader to work with a I2C EEPROM via the TWI

## Description
The two main issues I found were:
- The non initialized TWI peripheral
- Flipped ACK/NACK on the read cycle.

I also added a approx. 5ms delay between write cycles, as the EEPROM needs up to 5ms to store the received data. This resulted in about 7.6ms of actual delay time between write bursts. The delay could be optimised, but the current 7.6ms delay should be stable for most EEPROMs.

I attached captures of the resulting EEPROM read/write opperations. I also exposed the internal drive signals from the TWI peripheral to see is the TWI peripheral was causing problems, or the attached device was interfering.

|Channel|Color|Function|
|--|--|--|
|Channel 12|yellow|SCL drive FPGA|
|Channel 13|green|SDA drive FPGA|
|Channel 14|blue|SCL bus|
|Channel 15|purple|SDA bus|

### Read cycle
image
### Write cycle
image


SirBramble:
- Added a bus released delay before the repeated start, as the setup time with no delay was to short for the EEPROM due to the reduced pulse length from SDA at higher frequencies.

Tested with 800kHz and worked fine.

stnolting:
> The non initialized TWI peripheral.

That is intentional. :wink: The TWI is disabled by default in `config.h` to keep the bootloader small enough so it can fit into a 4kB ROM.

> Flipped ACK/NACK on the read cycle.

You mean the ACK after the processor has received a data byte? The EEFPROM/flash should generate a NACK here, right (i.e. keep SDA high)?

> I also added a approx. 5ms delay between write cycles, as the EEPROM needs up to 5ms to store the received data.

I did not consider this. Do you propose a dynamic delay (based on the processor clock and/or the TWI clock speed configuration) or should we just decrease the TWI clock?

sw/bootloader/config.h
```diff
@@ -11,131 +11,142 @@
...
-#define TWI_FLASH_BASE_ADDR 0x00000000U
-#endif
-
-// TWI flash address bytes (1,2,3,4)
-#ifndef TWI_FLASH_ADDR_BYTES
-#define TWI_FLASH_ADDR_BYTES 2
-#endif
-
-#endif // CONFIG_H
+ #ifndef CONFIG_H
```
stnolting, line 14: Please remove the leading spaces, as they make the diff unclear.

sw/bootloader/makefile
```diff
@@ -11,16 +11,20 @@ EFFORT = -Os
...
 USER_FLAGS += \
--Wl,--defsym,__neorv32_rom_size=4k \
+-Wl,--defsym,__neorv32_rom_size=8k \
 -Wl,--defsym,__neorv32_rom_base=0xFFE00000 \
 -Wl,--defsym,__neorv32_ram_size=256 \
 -DMAKE_BOOTLOADER \
 -flto
 
+RISCV_PREFIX ?= ~/xpack-riscv-none-elf-gcc-14.2.0-3/bin/riscv-none-elf-
+
```
stnolting, line 27: These are custom settings that should not be part of the code base.

sw/bootloader/main.c
```diff
@@ -87,7 +87,15 @@ int main(void) {
   // setup TWI
 #if (TWI_EN != 0)
   if (neorv32_uart0_available()) {
-    neorv32_twi_available();
```
stnolting, line 90: Oh... 🙈
Thanks for fixing this!

sw/bootloader/main.c
```diff
@@ -152,12 +160,16 @@ int main(void) {
 
   // try booting from SPI flash
 #if (SPI_EN != 0)
-  if (load_exe(EXE_STREAM_SPI) == 0) { start_app(); }
+  if(neorv32_spi_available()) {
```
stnolting, line 163: It is not necessary to check here whether SPI is available. The final SPI functions that are called later already have such checks.

sw/bootloader/main.c
```diff
@@ -152,12 +160,16 @@ int main(void) {
...
+    if (load_exe(EXE_STREAM_SPI) == 0) { start_app(); }
+  }
 #endif
 
   // try booting from TWI flash
 #if (TWI_EN != 0)
-  if (load_exe(EXE_STREAM_TWI) == 0) { start_app(); }
+  if(neorv32_twi_available()) {
```
stnolting, line 170: It is not necessary to check here whether TWI is available. The final TWI functions that are called later already have such checks.

sw/bootloader/main.c
```diff
@@ -504,8 +516,9 @@ void save_exe(int dst) {
     }
     addr += 4;
     i += 4;
+    uart_putc('.');
```
stnolting, line 519: This status info is good for debugging, but should not be part of the final version. Furthermore, we should reduce image size to an absolute minimum.

sw/bootloader/hal/source/twi_flash.c
```diff
@@ -11,198 +11,225 @@
...
-    return 1;
-  }
-  else {
-    return 0;
-  }
-#else
-  return 1;
-#endif
-}
+ #include <neorv32.h>
```
stnolting, line 14: Please also remove these leading spaces. With them in place it is hard to see the actuall diff.

stnolting (changes requested):

sw/bootloader/config.h
SirBramble: Ah, so thats what it was. Thanks for the info.

sw/bootloader/main.c
SirBramble: Yeah, thats what I meant with the non initialised TWI😅

sw/bootloader/main.c
SirBramble: Ah, sry, that was a quick fix. The controller crashed when trying to boot off SPI with no SPI peripheral installed.
But the cleaner option would be to just disable it with the compiler flags.

sw/bootloader/main.c
SirBramble: Fair enough. Will remove

SirBramble:
> You mean the ACK after the processor has received a data byte? The EEFPROM/flash should generate a NACK here, right (i.e. keep SDA high)?
Kind of. When reading, the processor should ACK Packets sent from the EEPROM until it's read enough and NACKs the last packet followed by a STOP command. The previous bootloader did that too, but ACK and NACK were flipped. So the first 3 packets were NACKed and the last packet ACKed. But probably just a typo, as the TWI transfer function takes the master ACK (mack) as an argument, but a 1 in the function (aka. mack sent) means the SDA line is pulled to 0.

> I did not consider this. Do you propose a dynamic delay (based on the processor clock and/or the TWI clock speed configuration) or should we just decrease the TWI clock?
I thought about using the TWI NOP commands as a delay mechanism. I used them as a read delay, as the required wait time is somewhere around 100us (could probably be shorter, but works quite well).
However, thinking about it now, a TWI-clock independent delay might be better. As most datasheets provide a worst-case wait time independent of the applied TWI clock. So 5ms, no matter if it's 100KHz or 1MHz. So probably more to do with storage access than the actual TWI interface.

sw/bootloader/makefile
stnolting: We should also keep the config.h in its default state (with the slow TWI as fall-back). I can revert that if that is ok.

sw/bootloader/config.h
```diff
@@ -115,12 +115,17 @@
...
+#define TWI_CLK_PRSC CLK_PRSC_64
 #endif
 
 // TWI clock divider
 #ifndef TWI_CLK_DIV
-#define TWI_CLK_DIV 1
+#define TWI_CLK_DIV 3
+#endif
+
+// TWI allow clock streching
```
stnolting, line 126: That's a nice addon!

sw/bootloader/hal/source/twi_flash.c
```diff
@@ -73,6 +73,12 @@ int twi_flash_read_word(uint32_t addr, uint32_t* rdata) {
   #error "Invalid TWI_FLASH_ADDR_BYTES configuration!"
 #endif
 
+  // delay repeated-start with bus released
+  uint32_t ctrl = NEORV32_TWI->CTRL;
+  NEORV32_TWI->CTRL = 0;         // disable TWI
+  for(int i = 0; i < 5; i++) asm volatile("nop"); // delay
+  NEORV32_TWI->CTRL = ctrl;	    // enable TWI
+
```
stnolting, line 81: Do we really need to disabled the TWI module here?

If we decrease the TWI clock, could we then remove this busy wait?

sw/bootloader/hal/source/twi_flash.c
```diff
@@ -84,26 +90,22 @@ int twi_flash_read_word(uint32_t addr, uint32_t* rdata) {
...
 
-  if (device_nack) {
-    neorv32_uart0_puts("\nTWI_R ERR\n");
-    return 1;
-  }
-  else {
-    neorv32_uart0_puts("\nTWI_R OK\n");
-    return 0;
-  }
+  return 0;
```
stnolting, line 108: Shouldn't we return `device_nack` in case the device does not respond to the address phase?

sw/bootloader/hal/source/twi_flash.c
```diff
@@ -191,11 +200,28 @@ int twi_flash_write_word(uint32_t addr, uint32_t wdata) {
 
   // write four bytes
   data.uint32 = wdata;
-  device_nack += twi_flash_write_byte(addr+0, data.uint8[0]);
-  device_nack += twi_flash_write_byte(addr+1, data.uint8[1]);
-  device_nack += twi_flash_write_byte(addr+2, data.uint8[2]);
-  device_nack += twi_flash_write_byte(addr+3, data.uint8[3]);
-
+  #if(TWI_FLASH_BULK_WRITE_EN != 0)
```
stnolting, line 203: This bulk feature is interesting. Is it required for some I2C devices? Or is this just a speed optimization?

stnolting (changes requested):

sw/bootloader/config.h
SirBramble: Thanks. Thought it might be useful, seeing as the peripheral supports it.
Might even be usable as a replacement for the read/write delays. Although that's just superstition at the moment, as I don't know if any EEPROMs support clock stretching.

sw/bootloader/hal/source/twi_flash.c
SirBramble: Just a speed optimisation. Sending the bytes one at a time with 5ms delay in between took quite long.
It makes use of the page write functionality that quite a few EEPROMs (at least he ones I tested) have.
It only works if you stay within the same 16-byte (or 32-byte) segment/page per transfer, but seeing as we start from 0x00 (or at least usually a 4-byte multiple) and send 4-byte blocks, it shouldn't cause any problems.

sw/bootloader/hal/source/twi_flash.c
SirBramble: Good point. Didn't think of that...

sw/bootloader/hal/source/twi_flash.c
SirBramble: A lower clock would probably work. But then it would set a cap on the TWI clock speed.
At higher clock speeds, the SDA line isn't pulled high for long enough, causing the repeated start to not trigger reliably or at all.

SirBramble:
Just tested the bootloader after the changes on Hardware. TWI read/write works fine.
Used the following flags:
```bash
-DTWI_EN=1 -DTWI_CLK_PRSC=3 -DSPI_EN=0
```
However, even with SPI disabled, the bootloader won't fit in 4k ROM. Just to large with 4328 bytes. I'll have a look if I can squash it down a bit more.

SirBramble:
Managed to press it down to 4288 bytes by getting rid of the delay before the repeated start.
However, for the repeated start condition to work with higher TWI clocks, the SDA line '0' state had to be delayed by one TWI clock cycle.
My guess is that the SDA and SCL lines have to be held high (bus released) for a fixed time before a repeated start can occur. By delaying the SDA low write, the lines are released for longer, omitting the need for a active bus release in code.

Before:
grafik

After:
grafik

SirBramble:
Added an option to disable the UART splash screen. Got it compile down to 4020 bytes. Tested with 800kHz TWI clock.
Flags:
```bash
-DTWI_EN=1 -DTWI_CLK_PRSC=2 -DSPI_EN=0 -DUART_PRINT_SPLASH_EN=0
```

stnolting:
> However, even with SPI disabled, the bootloader won't fit in 4k ROM. Just to large with 4328 bytes. I'll have a look if I can squash it down a bit more.

For a custom setup you can enabled the `C` ISA extensions adding compressed instructions. This will reduce code size by about 30%.

> However, for the repeated start condition to work with higher TWI clocks, the SDA line '0' state had to be delayed by one TWI clock cycle.

A hardware modification?

rtl/core/neorv32_twi.vhd
```diff
@@ -333,7 +333,7 @@ begin
         -- ------------------------------------------------------------
           if (clk_gen.phase(0) = '1') then
             io_con.sda_out <= '1';
-          elsif (clk_gen.phase(1) = '1') then
+          elsif (clk_gen.phase(2) = '1') then
```
stnolting, line 336: Let's keep hardware changes out of this SW-only branch.
Please make another PR after merging this.
Ok for you?

sw/bootloader/config.h
```diff
@@ -42,6 +42,11 @@
 #define UART_HW_HANDSHAKE_EN 0
 #endif
 
+// Print splash screen
+#ifndef UART_PRINT_SPLASH_EN
+#define UART_PRINT_SPLASH_EN 1
+#endif
+
```
stnolting, line 49: That's a nice new option!

sw/bootloader/hal/source/twi_flash.c
stnolting: If it is just a speed optimization then maybe we should leave that out here.
I think it would be a good idea to keep the bootloader code base as small as possible. My goal is to have all options enabled at some time and still fitting within the 4kB ROM... 😅

stnolting (changes requested):

rtl/core/neorv32_twi.vhd
SirBramble: Yeah, that's fine. I'll put it in a separate PR 👍

sw/bootloader/config.h
SirBramble: Thanks. Was looking for a way to save some bytes and removing the splash screen saved about 150 bytes I think. But I'd have to check to be sure.

rtl/core/neorv32_twi.vhd
stnolting: Cool, thanks!

sw/bootloader/hal/source/twi_flash.c
SirBramble: I tried disabling it as a possibility to save some space, but it didn't make a big difference in file size. But I might be wrong. I will check this again tomorrow and return with some numbers.
Otherwise, would leaving it in and disabling the flag by default be an option?

stnolting: I think that's also fine.

sw/bootloader/hal/include/twi_flash.h
```diff
@@ -19,4 +19,20 @@
 int twi_flash_read_word(uint32_t addr, uint32_t* rdata);
 int twi_flash_write_word(uint32_t addr, uint32_t wdata);
 
+/**
+* @brief Keeps TWI Peripheral in IDLE for 'tick_count' TWI clock ticks
+*
+* @param tick_count Amount of TWI NOP ticks to wait
+*
+*/
+inline void __attribute__ ((always_inline)) twi_flash_delay_twi_tick(int tick_count){
```
stnolting, line 28: Btw, why is this an inline function? You are calling it 4 times, so it might be smaller to actually _call_ it... 🤔

SirBramble: TBH, I don't know...😅
I'll put it on the list👍

SirBramble: Ok, changed it to a called function, but the size didn't change.

sw/bootloader/hal/source/twi_flash.c
SirBramble: Ok,I compiled it with and without the flag and got this:
Bulk write enabled: 4020 bytes
Bulk write disabled: 4076 bytes
So aparrently, it actually saves space. No idea how though...
So, should it be disabled by default or left enabled?

stnolting: I think we should disable it by defualt and stick to the minimum operation set. Every I²C flash should at least support the byte write command.

stnolting (approved):

PR #1263, opened 2025-05-10
✨ add cache burst transfers

stnolting:
This PR introduces **read bursts** for the internal bus system and the external bus. Write bursts are not supported yet. In the current setup, only the CPU instruction and data caches can generate burst transfers. If both caches are disabled, no additional logic overhead (for the bursts) is generated.

Furthermore, the new cache design also offers a significant improvement in timing. Even with the dual-core configuration, the critical path is now back in the CPU and not in the bus system.

### External Bus Interface

The external bus interfaces now provides the Wishbone-compliant "cycle type identifer" (`cti`) to specific the transfer type:

* `000` - classic / single access
* `001` - constant address burst (used for atomic read-modify-write operations)
* `010` - incrementing address burst (used by the caches to fetch data from main memory)

### AXI4 Bridge

The AXI bridge / Vivado IP setup now also supports incrementing address AXI bursts.

### First Benchmarking

I am testing the cache bursts with a DDR RAM controller attached to the AXI bus interface. With this setup CoreMark runs about 65% faster when compared to a setup that does not use bursts.

Even if only processor-internal memories are used, the caches with their bursts provide a certain performance gain of approx. 10% (depending on the size of the caches). This gain increases again with the dual-core configuration.


NikLeberg:
That looks amazing @stnolting !
Is always a full cache line burst-read? I.e. `CACHE_BLOCK_SIZE` blocks?
There is no form of back pressure, right? After the first `ack` of `N` transactions has been received, the remaining `N - 1` reads are issued immediately without pause. So slaves are expected that they accept the N transactions?

Excoted to play around with this. :)

Unike267:
BTW

> The external bus interfaces now provides the Wishbone-compliant "cycle type identifer" (cti) to specific the transfer type:

Why three bits for just three cases? Are you planning to implement more options in the future?  :thinking: 

> I am testing the cache bursts with a DDR RAM controller attached to the AXI bus interface. With this setup CoreMark runs about 65% faster when compared to a setup that does not use bursts.

Wow, that's really cool! What DDR wrapper are you using? Is it available on gh, like [LiteDRAM](https://github.com/enjoy-digital/litedram)??

stnolting:
> Is always a full cache line burst-read? I.e. CACHE_BLOCK_SIZE blocks?

@NikLeberg 

That's right. This is also the reason why I$ line isze and D$ line size must be the same size.

> There is no form of back pressure, right? After the first ack of N transactions has been received, the remaining N - 1 reads are issued immediately without pause. So slaves are expected that they accept the N transactions?

That is also correct. If there is only one device that can be addressed by the cache and does not support bursts, then bursts cannot be used at all. That's not nice, but I wanted to do without a "de-burstr" to convert bursts back into single accesses.

> Excoted to play around with this. :)

It's all still experimental, but everything looks good so far. 😉

@Unike267 

> Why three bits for just three cases? Are you planning to implement more options in the future?

The `cti` signal is defined by the Wishbone spec. And I want to try to stay as close to this spec as possible.

Nevertheless, there is an incompatibility: Wishbone signals the end of a burst with `cti=111`, but the XBUS interface does not. However, this should not be a big problem as the burst size is always the same (= the size of a cache line).

> Wow, that's really cool! What DDR wrapper are you using?

Vivado MIG 😉

Unike267:
> The cti signal is defined by the Wishbone spec. And I want to try to stay as close to this spec as possible.

😬 I got no clue about! It must have been added "recently", because I’ve studied the specification.

However, the manual I have printed here in the lab is revision B4 (copyright 2010), so I guess that’s why I didn’t know about it, this signal isn’t mentioned in that version.

Looks like I’ll have to print a new copy lol.

> Vivado MIG 😉

:+1:

Unike267:
> revision B4

hum! I'm searching about and it looks like the latest revision is B4 :thinking: 

I'm searching with `ctrl + f` in the pdf and I am not able to find this signal :cry: 

I've found info about here: https://wishbone-interconnect.readthedocs.io/en/latest/04_registered.html#signal-description

Unike267:
Oh my bad!

The information about this signal is in the section `4.3 Signal Description`. I guess I haven't studied the standard as well as I thought.

stnolting:
These are the specs that I am using: https://wishbone-interconnect.readthedocs.io/en/latest/index.html

Maybe we should remove these outdated opencores-based Wishbone specs from the docs...

Unike267:
> Maybe we should remove these outdated opencores-based Wishbone specs from the docs...

Oh no! According to opencores webpage the B4 version is the last version of the spec.

Source: https://opencores.org/howto/wishbone

Also according to their github docs b4 is the latest version.

Source: https://github.com/fossi-foundation/wishbone/tree/master/documents/spec

Unike267:
BTW, it is funny that the spec was uploaded by Unai and the copy of the spec that I've in the lab was also printed by him (if something goes wrong we already know who to blame) 😂😂

stnolting:
> Oh no! According to opencores webpage the B4 version is the last version of the spec.

OpenOcres has been dead for many years, so this is not surprising.

> https://wishbone-interconnect.readthedocs.io/en/latest/index.html

These are from 2019. But I have no idea who stands behind those specs.

> BTW, it is funny that the spec was uploaded by Unai and the copy of the spec that I've in the lab was also printed by him (if something goes wrong we already know who to blame) 😂😂

Then we have found the culprit... 😅

PR #1292, opened 2025-06-19
Remove enable logic for SoC-wide clock generator

stnolting:
Now the clock generator is always active providing "derived" clocks for the IO modules.


stnolting:
A continuously running 12-bit counter (= the clock generator) is probably negligible compared to a permanently toggling clock tree (as far as dynamic power dissipation is concerned)... 😉

PR #1305, opened 2025-07-06
Direct state change in case of misaligned memory access

mmcheraghi:
In case of misaligned memory access, CPU goes to DISPATCH unit first. Direct state change into EX_TRAP_ENTER saves one cycle.


stnolting:
> In case of misaligned memory access, CPU goes to DISPATCH unit first. Direct state change into EX_TRAP_ENTER saves one cycle.

You got that right!
However, I would refrain from this "shortcut" for two reasons:

1. Exceptions should never occur and if they do, they are very rare. So there is no direct performance gain if we implement a shortcut here.
2. At the moment there is only one retrievable state from `EX_MEM_RSP` (namely `EX_DISPATCH`). If we add `EX_TRAP_ENTER` as another reachable state, we also need another multiplexer (i.e. additional logic).

mmcheraghi:
Ok so I close this pull request.

stnolting:
> Ok so I close this pull request.

That's what I would suggest.

So don't get me wrong, I welcome all contributions - but especially suggestions for improvement! :+1:
 In this case, I just don't think the changes would be too much of a gain.

mmcheraghi:
No problem. I've learned many things from your project.

stnolting:
Cool, I'm glad to hear that! :)

Issue #1297, opened 2025-06-24
More precise instruction timing documentation


CharlesAverill:
# TL;DR

I would like the instruction timing documentation in the datasheet to provide a more fine-grained description than worst-case execution time.
Ideally, I would like a formula parametrized over operands - even better if cache state is included as a parameter as well.

# Why?

I'm the lead on a project involving formally-verified timing bounds for machine code.
See this [recently-published abstract at PLDI's 2025 SRC](https://www.charles.systems/publications/PTM_SRC.pdf) and a [presentation](https://m.youtube.com/live/BRhBv_aYNks?si=MQarpFIYobx_shU0&t=1h47m12s).
In short, we've developed the capability to mathematically prove exactly how long it takes for a program to execute.
I use the NEORV32 as the CPU of choice for my primary examples - [several functions from FreeRTOS-Kernel/tasks.c](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/tasks.c) and [a ChaCha20 implementation](https://github.com/CharlesAverill/time_sensitive_riscv/blob/main/chacha/final_chacha20.S).

A critical part of the formal timing project is that it relies completely on manufacturer/designer documentation for instruction timing.
To prove something like "this cryptographic implementation has no timing side-channel vulnerabilities," worst-case execution time per instruction does not suffice.
For this level of granularity, I need **exact-case execution time** per instruction.

I know this is a big ask, so I'm open to helping wherever I can. 
I think a collaboration here would be mutually beneficial for the NEORV32 project and for my work.

stnolting:
Hey @CharlesAverill,
that sounds like an interesting project and I am of course pleased that you are using NEORV32.

>For this level of granularity, I need exact-case execution time per instruction.
I know this is a big ask, so I'm open to helping wherever I can.
I think a collaboration here would be mutually beneficial for the NEORV32 project and for my work.

I would also find an exact timing model very helpful, e.g. to be able to make an a priori estimate of how long a certain code takes to execute. So I also think that this could be a valuable addition.

I think it wouldn't be too much trouble to update the timing information in the documentation first and then make it more detailed. But as soon as you leave the actual CPU behind and look at the rest of the system, it becomes more complicated. We have caches, several CPU cores and also a DMA, which all together makes the timing model much more complicated.

Do you have any ideas on the best way to get started? Or what would you focus on (initially)?

CharlesAverill:
[...]

PR #1313, opened 2025-07-12
✨ add new module: execution trace buffer (TRACER)

stnolting:
Some time ago I read a blog article about the ARM embedded trace buffer (-> [interrupt.memfault.com/blog/instruction-tracing-mtb-m33](https://interrupt.memfault.com/blog/instruction-tracing-mtb-m33)) and was quite excited about the concept. Hence, this PR adds a very similar module to the processor that allows for execution tracking via "instruction delta tracing".

The tracer module is configured by two new top generics:

```vhdl
IO_TRACER_EN     : boolean := false; -- implement instruction tracer
IO_TRACER_BUFFER : natural range 1 to 2**15 := 1 -- trace buffer depth, has to be a power of two, min 1
```

Once started, the trace buffer collects the instruction execution flow in an internal FIFO. Older elements are automatically discarded so that the buffer only contains the last `IO_TRACER_BUFFER` instruction deltas. Tracing can be stopped manually or automatically: when a certain address is reached, the tracer stops automatically and can optionally trigger an interrupt.

The trace data can be processed directly (stand-alone; no debugger required) by the processor:

```
<< NEORV32 Tracer Demo Program >>

Trace buffer: 64 entries
Starting trace...

Trace log:
[0] SRC: 0x000003b8 -> DST: 0x000002cc <TRACE_START>
[1] SRC: 0x000002d0 -> DST: 0x000004cc <TRAP_ENTRY>
[2] SRC: 0x00000568 -> DST: 0x0000056c
[3] SRC: 0x00000590 -> DST: 0x000001e4

Program completed
```

Alternatively, the data can also be output and displayed directly using a GDB script:

grafik


stnolting:
Things planned for upcoming PRs:

* write full execution trace to file (simulation-only; -> https://github.com/stnolting/neorv32/issues/1297#issuecomment-3016410456)
* add new folder with GDB helper scipts (trace configuration, trace readout, etc.)
* integrate basic trace features in Eclipse
* add a tracer user guide

As always, I would appreciate ideas, thoughts or help.

PR #1333, opened 2025-08-09
minor fixes and optimizations

stnolting:
* Runtime environment (RTE): Adjust the return address (`mepc`) _after_ calling the handler. This allows the trap handler to check/use the original `mepc` value.
* Hardware trigger module (`Sdtrig` ISA extensions): Correct trigger logic. Now the instruction address trigger is triggered precisely even when it is set to an instruction that raises an exception (such as the environment call command `ecall`).
* Single-stepping: Rework single-step trap trigger (turn it into an actual _synchronous_ exception).
* `mtinst` CSR: Do not write zero to the CSR in case of an interrupt. `mtinst` now also shows the last executed instruction word when encountering an interrupt.
* CSR write logic: Code cleanup; use a single `case` statement instead of several `if` statements.
* Code cleanups and logic optimizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 09.08.2025 | 1.11.9.2 | minor fixes and optimizations | [#1333](https://github.com/stnolting/neorv32/pull/1333) |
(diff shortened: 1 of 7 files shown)


BEforlin:
Hi @stnolting, I'm not sure where, but one of these commits seems to have broken "Table 90. NEORV32 Trap Listing" on the datasheet "docs/datasheet/cpu.adoc". It is not displaying on the HTML version.

stnolting:
That's probably just a typo in the table headers. Thanks for pointing that out!

stnolting:
@BEforlin fixed

PR #1341, opened 2025-08-16
[Vivado IP] Remove relative pathes from the IP-packaging script

### LINKED ISSUE #1340: [User Guide] System integration - running TCL script through GUI (Tools -> Run TCL script)
In the `7. Packaging the Processor as Vivado IP Block` section of the user guide, it is mentioned:

```
In GUI mode select either "Tools → Run TCL Script" to directly execute the script or open the TCL shell ("Window → Tcl Console") to manually invoke the script.
```

In my case (using Vivado 2020.2 on Ubuntu), running the script like that results in the following error:
Image

Running the script from the Tcl console worked well.

stnolting:
Fixes #1340.

The IP packagin script can now be called from any location.

@michalmonday Could you check this

### DIFF
--- a/docs/userguide/packaging_vivado.adoc
+++ b/docs/userguide/packaging_vivado.adoc
@@ -23,14 +23,28 @@ with tool tips to make it easier to understand the different configuration optio
 .NEORV32 IP Customization GUI
 image::vivado_ip_gui.png[]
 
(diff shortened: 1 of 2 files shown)


michalmonday:
Just tested and it worked well, thank you.

stnolting:
Thanks for the feedback!

PR #1344, opened 2025-08-17
[rtl] replace individual IMEM & DMEM modules by a generic memory component

stnolting:
* remove `neorv32_imem.vhd` and `neorv32_dmem.vhd`
* replace by instances of a generic memory module `neorv32_mem.vhd`


Unike267:
We should change the generics in this image:

Captura de pantalla de 2025-08-18 10-16-05

From `MEM_INT_*_SIZE` to `*_SIZE`

BTW, what program do you use to create images?? 

Cheers!

stnolting:
> From MEM_INT_*_SIZE to *_SIZE

Oh right, I'll update that. Thanks for the hint!

> BTW, what program do you use to create images??

Just PowerPoint 😅
I would like to switch to a free alternative, but I haven't found a nice tool yet.

Unike267:
> Just PowerPoint 😅

Nice!

>  I would like to switch to a free alternative, but I haven't found a nice tool yet.

I usually use latex (tikz) to draw my figures, but it takes me the whole day to draw a simple image so I don't recommend it unless you've a lot of patience. Besides, the result through ppt is so cool. :+1:

stnolting:
Latex is far too complicated for me (or maybe I'm just too stupid to use it). I would like to "script" the block diagrams. Ascidoctor has rudimentary support for this, but I don't really like the end results. Until I find something better (and open source), I'll stick with PowerPoint for now.

> Besides, the result through ppt is so cool. 👍

Thanks. 😉

Unike267:
> or maybe I'm just too stupid to use it

I don't think so.

> I would like to "script" the block diagrams. 

I think it's good practice. I've got used to doing it this way.

computer_dog

PR #1350, opened 2025-08-21
[demo_semihosting] use 4-byte instead of 16-byte alignment

### LINKED ISSUE #1348: example demo_semihosting fails to link with Zcb extension due to alignment error
**Describe the bug**  
When building the `sw/example/demo_semihosting` example with the new Zcb extension (https://github.com/stnolting/neorv32/pull/1320) enabled in the `-march` string, the build fails at the linking stage due to an alignment error. By default, the example uses `-march=rv32i_zicsr_zifencei`. When using `-march=rv32i_zicsr_zifencei_zcb`, the linker reports that 14 bytes are required for alignment to a 16-byte boundary, but only 12 are available.

**To Reproduce**  
I use the newest RISC-V GNU toolchain and run the following in `sw/example/demo_semihosting`:

```bash
make clean_all exe
```

The build fails with:

```
/path/to/toolchain/riscv32-unknown-elf/bin/ld: build/neorv32_semihosting.c.o(.text.neorv32_semihosting_open+0x12): 14 bytes required for alignment to 16-byte boundary, but only 12 present
/path/to/toolchain/32-bit/lib/gcc/riscv32-unknown-elf/bin/ld: can't relax section: bad value
```

**Expected behavior**  
The example should link successfully with compressed instructions enabled in the `-march` string.

**Environment:**

- OS: Ubuntu 25.04
- GCC Version (RISC-V): riscv32-unknown-elf-gcc 15.1.0
- GCC Version (native): 15.1.0
- Libraries used: newlib

**Additional context**  
The error points to an alignment issue. In [neorv32/sw/lib/include/neorv32_semihosting.h](https://github.com/stnolting/neorv32/blob/75b2eeac6474106ed1c0b6eb085075a34cb19499/sw/lib/include/neorv32_semihosting.h#L109) there is an inline assembly directive `.align 4`. I think on RISC-V this means align to a 2^4 = 16-byte boundary. When I change `.align 4` to `.balign 4`, which explicitly aligns to 4 bytes, the error disappears and the example links successfully with Zcb enabled. 

I am unsure if this code truly requires 16-byte alignment. If only word alignment is needed, replacing `.align 4` with `.balign 4` (or `.align 2`) would avoid this problem. 

I am happy to provide a one character pull request if the `.balign 4` fix is the correct solution.

vogma:
Closes #1348

### DIFF
--- a/sw/lib/include/neorv32_semihosting.h
+++ b/sw/lib/include/neorv32_semihosting.h
@@ -106,7 +106,7 @@ inline int __attribute__ ((always_inline)) neorv32_semihosting_req(int id, void*
   asm volatile (
     " .option push      \n"
     " .option norvc     \n" // we need 32-bit instruction words here
-    " .align 4          \n" // this has to be aligned
+    " .balign 4         \n" // this has to be aligned
     " slli x0, x0, 0x1f \n" // magic triplet: entry NOP
     " ebreak            \n" // magic triplet: break to debugger
     " srai x0, x0, 0x07 \n" // magic triplet: exit NOP


stnolting:
Thanks for fixing!

Issue #1373, opened 2025-09-12
Illegal instruction access fault issues from formal verification


mikaelsky:
**Describe the bug**
During formal verification of the NeoRV32 core 2 issues where uncovered related to illegal instruction access.

Case 1.
Unaligned uncompressed instruction only checks for bus errors on the high part of the unaligned instruction:
E.g. neorv32_cpu_frontend
<img width="596" height="94" alt="Image" src="https://github.com/user-attachments/assets/986d987a-ef47-422d-94bd-adfb7e9ea8ca" />

Case 2.
Illegal instruction access instructions are allowed to execute which results in various side effects.
Now we are still running on an older version of the core, so this seems to be less of an issue now :)
It still hits the FSRs as a float instruction still gets kicked off and won't be stopped even though there is a trap. It still not clear whether we can have on odd-ball write back from the FPU post trap.

**To Reproduce**
For case 1 you need to have an unaligned instruction where the lower half is from a location that reports a bus error and the upper half doesn't report a bus error. Its a bit tricky to setup.
0x0000 c.nop      | nop(high)
0xFFFC nop(low) | c.nop

**Expected behavior**
An illegal instruction access trap should be called.

**Screenshots**

**Environment:**
 - Libraries used newlib

**Hardware:**
 - Hardware version (1.9.3)

**Additional context**
NA

stnolting:
> Case 2.
Illegal instruction access instructions are allowed to execute which results in various side effects.
Now we are still running on an older version of the core, so this seems to be less of an issue now :)
It still hits the FSRs as a float instruction still gets kicked off and won't be stopped even though there is a trap. It still not clear whether we can have on odd-ball write back from the FPU post trap.

I think we more or less solved the problem some time ago. Anyway, ALU co-processors are still being triggered - even if an invalid instruction or an instruction resulting from a bus error is to be executed.

However, this should not be a problem for the FPU, as the exception is detected before the FPU can generate any results or change any statuses (the CFU needs a lot of cycles to commit a signle operation).

However, I think the ALU co-processors should not be triggered at all—that seems cleaner to me. I'll rework that. :+1:

> Case 1.
Unaligned uncompressed instruction only checks for bus errors on the high part of the unaligned instruction:
E.g. neorv32_cpu_frontend
>
> To Reproduce
For case 1 you need to have an unaligned instruction where the lower half is from a location that reports a bus error and the upper half doesn't report a bus error. Its a bit tricky to setup.
[...]

PR #1376, opened 2025-09-13
:bug: fix unaligned instruction fetch bus error; do not trigger co-processors if instruction exception

stnolting:
* (minor :bug:) do not trigger any ALU co-processor if there is a pending instruction-related exception (to prevent any co-processor-internal state change like the FPU's flag register) -> #1373
* :sparkles: update testbench to provide a generic RAM with programmable error-response for each memory word
* :bug: if a 32-bit instruction is unaligned (split accross two 32-bit accesses) the front end has to check if **any** of the two bus accesses caused a bus error; in case of a bus error the **first** error repsonse has to be reported for the exception -> #1373

### DIFF
--- a/sw/example/processor_check/main.c
+++ b/sw/example/processor_check/main.c
@@ -2466,11 +2466,11 @@ void test_ok(void) {
 
 
 /**********************************************************************//**
- * Test results helper function: Shows "[FAIL]" and increments global cnt_fail
+ * Test results helper function: Shows "[fail]" and increments global cnt_fail
  **************************************************************************/
 void test_fail(void) {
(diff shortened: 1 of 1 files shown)


stnolting:
Thanks to the great description by @mikaelsky in https://github.com/stnolting/neorv32/issues/1373#issue-3412003454, I was able to create a new test case. The processor_check fails - which is good, because it shows that this is indeed a bug that is now being detected:

```
[12] IF unaligned access EXC [fail(12)]
```

stnolting:
Thanks for reviewing!

PR #1387, opened 2025-09-26
[rtl] use entity-instantiation for all internal modules

stnolting:
Introduced with VHDL 93 and recommended since VHDL 2008.

### DIFF
--- a/docs/datasheet/overview.adoc
+++ b/docs/datasheet/overview.adoc
@@ -277,8 +277,7 @@ The code makes minimal use of VHDL 2008 features in order to remain compatible w
 The specific user-defined configuration is done entirely via the generics of the top entity.
 * Internally, all generics are checked to ensure correct configuration.
 Asserts and sanity checks are used to inform the user about the actual processor configuration and possible invalid settings.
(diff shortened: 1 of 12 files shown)


stnolting:
Does not work for single-pass analysis tools (like Synplify) as modules are instantiated before being declared.

tmeissner:
Mhm, I use direct instantiation a lot, also with Synplify without problems 🧐 Only with mixed language designs you have to define Verilog modules as vhdl components before instantiating them in vhdl designs.

stnolting:
> [...] you have to define Verilog modules as vhdl components before instantiating them in vhdl designs.

That could be the problem. Some files contain multiple modules, but these are first instantiated and then declared.

E.g.:

Instantiation:

https://github.com/stnolting/neorv32/blob/c197934e3eb86f9448b822229fcaa4a4ebef51c4/rtl/core/neorv32_cache.vhd#L317

Declaration:

https://github.com/stnolting/neorv32/blob/c197934e3eb86f9448b822229fcaa4a4ebef51c4/rtl/core/neorv32_cache.vhd#L360

360 > 317

So maybe we just need to rearrange the modules within a file. 🤔

PR #1393, opened 2025-10-04
🐛 [dtm] fix bypass register; further code optimizations

stnolting:
* 🐛 fix JTAG bypass register (pass-through was broken; bug introduced in v1.12.2.5 / #1390)
* further logic optimization / simplifcation: replace individual `dtmcs`, `dmi`, `idcode` and `bypass` registers by a single shift register

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 04.10.2025 | 1.12.2.8 | :bug: fix broken DTM bypass register; further DTM logic optimization | [#1393](https://github.com/stnolting/neorv32/pull/1393) |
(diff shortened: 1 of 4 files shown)


NikLeberg:
Hi @stnolting, this has been merged since a long time. But I just stumbled over it ... why can we just omit the read only `dmi.op` alias from `dtmcs` register? Is it just because OpenOCD seems to not use it? Everything working for me so this is not an issue. But I'm adapting it to use the Lattice `JTAGG` primitive (stay tuned :wink:) and saw it.

stnolting:
Good question... it's been a long time...

`dtmcs.op` is a read-only alias of `dmi.op`. According to the spec, the two bits of `dmi.op` mean the following: 

grafik

-> https://docs.riscv.org/reference/debug/v1.0/dtm.html#dmi-op

The debug-module interface on the NEORV32 is really minimalistic. Each access takes only two clock cycles (making it faster than any JTAG access) and is always successful (no access errors).

* 1 cannot occur because it is reserved.
* 2 cannot occur because DM accesses are always successful.
* 3 cannot occur because DM accesses are significantly faster than any JTAG access.

That's why it is hardwired to zero (I think).

> But I'm adapting it to use the Lattice JTAGG primitive

Cool! Keep us (me!) updated! ;)

PR #1406, opened 2025-10-14
[sw/example] mark output as early-clobber in unaligned-load test

vogma:
**Problem**  
When building `processor_check` with a GCC 15.1.0 RISC-V cross compiler I built from source, the “Unaligned load address” test fails due to register allocation differences in the inline asm block. The compiler reuses the same register for the output and the address input, causing the test to fail.

**Before (works with xPack GCC 14.2.0, fails with GCC 15.1.0)**

xPack (GCC 14.2.0) generated assembly:

```asm
li   a5, 1
lui  a3, 0xcafe1
addi a3, a3, 560
lw   a3, 0(a5)
```

GCC 15.1.0 reused `a5` for both operands:

```asm
li   a5, 1
lui  a5, 0xcafe1
addi a5, a5, 560
lw   a5, 0(a5)
```

In the second case, `a5` is both the destination register and the base address register, which executes a load from the wrong address.
  
If i understand the gcc documentation correctly, GCC’s extended asm assumes inputs are consumed before outputs are produced, so it may legally allocate the same register to an input and an output unless told otherwise. For outputs that are written before all inputs are used (as in this two-instruction sequence), GCC requires the output to be marked early-clobber with `&` so it cannot overlap any input register. ([gcc.gnu.org](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html))

**Change**  
I added  `&` to the output constraint to enforce distinct registers for `%[da]` and `%[ad]` ([gcc.gnu.org](https://gcc.gnu.org/onlinedocs/gcc/Modifiers.html))
With this change, the unaligned-load test passes with both toolchains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 10.10.2025 | 1.12.3.4 | :warning: CFU: remove R4-type instruction support (three source registers); add support for I-type instructions | [#1402](https://github.com/stnolting/neorv32/pull/1402) |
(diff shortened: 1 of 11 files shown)


stnolting:
Now this is interesting! I didn't even know about the "&" option. Thanks for the adjustment!

Issue #633, opened 2023-06-21
feature request - support for Zc* extensions


biosbob:
the proposed `Zc*` extensions described [here](https://github.com/riscv/riscv-code-size-reduction) have recently been ratified....

the `Zca` extensions is of particular interest in "small cores" with limit memory resources....

just a placeholder for what appears to be a non-trivial improvement....

stnolting:
The standard `C` extension is already implemented (-> `CPU_EXTENSION_RISCV_C` generic). In this particular case `C` = `Zcf`, but all the compressed floating-point operations are mapped to normal integer load/store because `Zfinx` is used instead of `F`.

I had a closer look at the `Zc*` ISA extension. It think it is quite promising. However, in terms of the NEORV32 I am not fully convinced if it would be a good idea to implemten all of them.

* `Zca` - this is what we have if the FPU is disabled (`Zfinx` disabled)
* `Zcf` - this is what we have if the FPU is enabled
* `Zcd` makes no sense as the FPU is single-precision only
* `Zcmp` (list-based push/pop similar to ARM) is quite interesting, but would require a lot of additional hardware. Furthermore, precise exception trapping is complex here as there are several memory loads/stores invoked by a single instruction.
* `Zcmt` (table-based jumps) might be nice thing to have. But this would have a high latency - so the only gain would be further code size reduction (not performance).
* `Zcb` - I really like this extension because it adds 16-bit variants for common operations (like multiplication). This would be quite easy to implement I think. So, yeah, maybe this sub-extension might be integrated in the future. :wink:

This is just my opinion. Any thoughts?

biosbob:
`Zcb` looks promising, in that it is relatively easy to implement....  with disciplined declarations of integer types in EM (`uint8`, `int16`, `uint32`, etc) this would mesh quite well....  future CPU implementations that use (say) an internal 8- or 16-bit ALU would also benefit; reducing ALU width obviously saves gates....

as for `Zcmp`, the EM runtime would generally place some common push/pop code fragments (used by LLVM) into the boot ROM....  even with the smallest boot ROM (say, 2K), there is plenty of "free and fast" instruction memory....

i'm not entirely sure what motivates `Zcmt`....  but honestly, reducing code size without a performance gain doesn't seem worth it anyway....

bottom line -- `Zcb` would be first, when we get around to it....

stnolting:
> bottom line -- Zcb would be first, when we get around to it....

[...]

PR #1417, opened 2025-10-30
WIP/RFC: Add RISC-V `Zcmp` code-size reduction extension to NEORV32

vogma:
This draft PR adds initial support for the ratified RISC-V **Zc** subset **Zcmp** (PUSH/POP and double-move macros) to NEORV32. The implementation of this extension has been discussed in #633, but was closed as not planned. I think this extension is interesting and so I gave the implementation a shot anyway. This is not ready to be merged but I would like to start a discussion on how the implementation can be refined. The current implementation executes all Zcmp instructions in simulation and on hardware. More specifically, the `processor_check` passes in simulation and the examples in `sw/example` run on my Arty A7-100T board.

I have summarized some information below. I would appreciate any feedback :)

## Synthesis results (Vivado 2025.1)

Baseline configuration: `rv32imc_zicntr`
With Zcmp enabled: `rv32imc_zicntr_zcmp`

**Top level**

| Config   | Slice LUTs | Slice Registers | LUT as Logic | LUT as Memory |
| -------- | ---------: | --------------: | -----------: | ------------: |
| Baseline |      1,939 |           1,468 |        1,915 |            24 |
| + Zcmp   |      2,132 |           1,511 |        2,108 |            24 |

Delta (top): +193 LUTs (+9.95%), +43 FFs (+2.92%)

## Code-size impact (`sw/example` executables)

| Program            | Zcmp (B) | No Zcmp (B) | Saved (B) | Saved (%) |
| ------------------ | -------: | ----------: | --------: | --------: |
| `processor_check`  |   25,992 |      26,352 |       360 | 1.37% |
| `game_of_life`     |    4,636 |       4,908 |       272 | 5.54% |
| `coremark` (`-Os`) |   32,560 |      33,252 |       692 | 2.08% |
| `bus_explorer`     |    8,532 |       9,016 |       484 | 5.37% |

The code size results are rather underwhelming for the example software shipped with the NEORV32. I guess because the examples are quite small, only a few cm.push and pop instructions can be generated by the compiler. I tried to compile the embench benchmark suite which was referenced in the Issue #633, but for now I have not been successful.

## Short overview (Zcmp)

Zcmp adds six 16-bit macros:

* `cm.push` / `cm.pop`
* `cm.popret` / `cm.popretz`
* `cm.mvsa01` / `cm.mva01s`

### Frontend

* Two existing FSMs remain: fetch (fills two 16-bit FIFOs) and issue (forms valid instruction words).
* New `uop_fsm` is synthesized only when `RISCV_ISA_Zcmp` generic is set.
  When the decompressor detects Zcmp, `issue_fsm` enters a Zcmp state and `uop_fsm` emits the micro-op sequence.
* The frontend outputs are multiplexed:

  * `frontend_bus_issue` (normal path)
  * `frontend_bus_zcmp` (uOp path)
    `frontend_o` selects between them via `zcmp_uop_in_progress`.

New signals from frontend to control-unit:

* `zcmp_in_uop_seq` – Zcmp micro-op sequence active
* `zcmp_start` – sequence begins next cycle
* `zcmp_atomic_tail` – atomic tail section of the sequence (no traps allowed)

### Control unit

* On `zcmp_start`, the PC is latched. During the micro-op sequence `exe_engine.pc` / `pc2` are held.
* Trap logic is inhibited when `zcmp_atomic_tail` is asserted to honor POP/POPRET atomicity.

The trap/exception handling across the atomic tail likely needs refinement. I’d appreciate feedback.

### Testing 

I created test benches that execute every valid combination of the six Zcmp instructions and compare the affected architectural state against known-good outputs from the RISC-V Spike (riscv-isa-sim) reference simulator. For each sequence, the benches verify destination register values, stack-pointer updates, and control-flow effects (cm.popret/cm.popretz, including a0 zeroing). 

The Hardware and Software implementation is based on the Verilog Zcmp implementation of the [Hazard3](https://github.com/Wren6991/Hazard3) RISC-V CPU

Thanks for taking a look. I am happy to iterate based on your feedback.


stnolting:
Hey @vogma, thanks for the PR and the detailed information!
I saw your work in the network view here on GitHub and followed it with great interest.

> The implementation of this extension has been discussed in https://github.com/stnolting/neorv32/issues/633, but was closed as not planned.

That's right. To be honest, I'm not a fan of the Zcmp extension. An instruction that breaks down into many different instructions is somehow the opposite of what RISC originally stood for. But that's just me. 😉

Anyway, your results look pretty good, and I think that makes a discussion more than appropriate!

> Synthesis results

+10% hardware costs - that's not bad for the code saving that you have provided!
Can you tell us anything about the maximum clock speed? Does the new extension affect the critical path?

> Code-size impact

That's not bad at all. But code size isn't that important to me here. Since many instructions are replaced by a single (16-bit wide) instruction, this should also have a noticeable impact on bus traffic. Have you checked whether, for example, the processor check or Coremark run a little faster as a result?

> Testing

That's really great work! I hope the [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test) TG comes up with some similar tests one day!

Did you check exception handling? What happens if any of the resulting memory instructions cause an exception? What happens when a debug request arrives during a `Zcmp` instruction?

(This is another point that bothers me about the `Zcmp` extension: exceptions are no longer precise and can no longer be assigned to a single operation.)

> Frontend

I noticed that your code affects many modules in the CPU. Have you checked whether the entire code is optimized away when the ISA extension is disabled?

As I said, I'm not a fan of the Zcmp extension. But I also think that this should be left up to the end user. So I think your PR is good!

I'm just wondering if the RTL changes could be condensed a little more. Basically, you need an FSM between the issue logic and the CPU's execution microsequencer. As soon as a Zcmp instruction is executed, the PC of the execution unit is stalled and the update signal to the issue logic or instruction prefetch buffer is suppressed. The FSM can then send as many instructions as it likes to the execution unit.

So I'm wondering if we could move most of the `Zcmp` logic into a new module that would be placed within the fron-end bus system. 🤔

vogma:
Thank you for the feedback @stnolting

> To be honest, I'm not a fan of the Zcmp extension. An instruction that breaks down into many different instructions is somehow the opposite of what RISC originally stood for. But that's just me. 😉

I agree it's quite different from all the other available RISC-V extensions. I only chose this extensions because it hasn't been implemented yet into this cpu core and i wanted to work with the NEORV32 :) 

> Can you tell us anything about the maximum clock speed? Does the new extension affect the critical path?

I haven't done extensive tests for Fmax yet, but when synthesized with a 100MHz clock signal the zcmp-enabled implementation has a worst-negative-slack (WNS) of +1.112 ns which is a bit better than without the zcmp extension (WNS=+1.03ns).

> Since many instructions are replaced by a single (16-bit wide) instruction, this should also have a noticeable impact on bus traffic. Have you checked whether, for example, the processor check or Coremark run a little faster as a result?

Thats a good point. I just noticed that in my branch the coremark executable throws a illegal instruction exception since the latest merge so i will have to investigate that before i can answer your question. 

> Did you check exception handling? What happens if any of the resulting memory instructions cause an exception? What happens when a debug request arrives during a `Zcmp` instruction?

Only in theory. The specification states that if the zcmp micro-op instruction sequence is not inside the atomic tail section, the sequence can be aborted and retried. If a trap occurs during these memory operations, the zcmp sequence is aborted. But i have not tested if the sequence is retried after that. 

The behaviour during debugger requests is a great point. Thanks for bringing that up. 

 
> (This is another point that bothers me about the `Zcmp` extension: exceptions are no longer precise and can no longer be assigned to a single operation.)

True. Because the PC is held at the address of the Zcmp instruction, only this macro instruction will be flagged. 


> I noticed that your code affects many modules in the CPU. Have you checked whether the entire code is optimized away when the ISA extension is disabled?

Good point. I compared my implementation (Zcmp generic => false) with the main branch and 21 LUTs and 8 Registers are not optimized away by my implementation.

> So I'm wondering if we could move most of the `Zcmp` logic into a new module that would be placed within the fron-end bus system. 🤔

Yes i will move as much logic as possible into a separate module. This will probably also fix the unaccounted LUTs and Registers from above. 

Thank you for your feedback. I gained some valuable information and will work on my implementation in the coming weeks.

stnolting:
> I only chose this extensions because it hasn't been implemented yet into this cpu core and i wanted to work with the NEORV32 :)

That's cool, and I really appreciate it!

> I haven't done extensive tests for Fmax yet, but when synthesized with a 100MHz clock signal the zcmp-enabled implementation has a worst-negative-slack (WNS) of +1.112 ns which is a bit better than without the zcmp extension (WNS=+1.03ns).

That sounds good!

> Yes i will move as much logic as possible into a separate module. This will probably also fix the unaccounted LUTs and Registers from above.

Thanks a lot! Let me know if I can help in any way.

vogma:
Hi, sorry for keeping this stale for so long. Here is a quick update on the progress.
                                                                                                                                                                                                                                                              
As you suggested, I moved almost all of the Zcmp logic into a dedicated module (`neorv32_cpu_zcmp.vhd`). The only changes remaining in the other CPU files are the signal wiring and a generate block in the control unit that holds the PC during micro-op sequences. All of this logic is gated behind the `RISCV_ISA_Zcmp` generic.

I re-ran synthesis on an Arty A7-35T to verify. With Zcmp disabled, the overhead compared to upstream is down to 3 LUTs and 0 registers. Enabling Zcmp adds 201 LUTs and 47 registers on top of that.

| Configuration | ISA string | WNS | LUTs | Registers |
|---|---|---|---|---|
| Upstream | `rv32imc_zicntr` | 1.089 ns | 1939 | 1439 |
| This branch (Zcmp disabled) | `rv32icm_zicntr` | 0.993 ns | 1942 (+3) | 1439 (+0) |
| This branch (Zcmp enabled) | `rv32icm_zcmp_zicntr` | 0.813 ns | 2143 (+201) | 1486 (+47) |

I also tried to address the open question about exception and interrupt handling. I wrote two tests in `sw/example/zcmp/`:

- **`cm_exc.c`** verifies that load/store exceptions triggered by Zcmp micro-ops report the correct mepc (pointing at the original Zcmp instruction, not a micro-op), and that the Zcmp FSM recovers cleanly so subsequent Zcmp instructions still work.
- **`cm_irq.c`** verifies that when a machine timer interrupt fires during the non-atomic portion of a Zcmp sequence, mepc is set correctly and the micro-op sequence restarts from scratch on mret. 

Both tests pass in simulation. I looked at the waveforms during simulation and verified that the interrupt stops the zcmp sequence and it gets executed again after the trap handler.  

I hope the implementation for these two tests is sane. It was my first time working with these features so i appreciate feedback.

stnolting:
Sorry for the late reply... I haven't had much time to work on this project lately.

> As you suggested, I moved almost all of the Zcmp logic into a dedicated module (neorv32_cpu_zcmp.vhd). The only changes remaining in the other CPU files are the signal wiring and a generate block in the control unit that holds the PC during micro-op sequences. All of this logic is gated behind the RISCV_ISA_Zcmp generic.

That sounds great! I'll have a look.

> With Zcmp disabled, the overhead compared to upstream is down to 3 LUTs and 0 registers. Enabling Zcmp adds 201 LUTs and 47 registers on top of that.

That's cool and speaks for a well-encapsulated design! :+1:

> I also tried to address the open question about exception and interrupt handling. I wrote two tests in sw/example/zcmp/:

I'll check that out. Btw, do you know if there are any riscv-arch-test test suites for this ISA extension? I don't think I've seen any of that yet.

There seems to be a problem with the Icarus-based verilog simulation. 🤔
Did you check your hardware modifications on a real FPGA?

vogma:
> There seems to be a problem with the Icarus-based verilog simulation. 🤔

Yes that's right. The failing test happened to catch a bug in my design which was fixed in https://github.com/stnolting/neorv32/pull/1417/changes/2364bdd82debbd204198e865970289179204cd7e. I forgot the reset assignment for two signals.

> Did you check your hardware modifications on a real FPGA?

Yes i tried the design on my Arty A7-100T and ran hello-world, game of life and coremark without issues

stnolting:
> Yes i tried the design on my Arty A7-100T and ran hello-world, game of life and coremark without issues

Awesome! Thanks for the update!

I'm currently working on writing back the data cache, and it's more complicated than I expected... As soon as I'm mostly done with that, I'll take a closer look at your updates. :+1:

Issue #1412, opened 2025-10-24
[feature request] add Ada support


ohenley:
Hi!

I maintain [GNAT-Academic-Program/neorv32-setups](https://github.com/GNAT-Academic-Program/neorv32-setups), a fork providing Ada support.

It includes a working Ada BIOS (usable as uploadable app or bootloader), and several student teams are extending it with Ada drivers for their own projects, one even building an NPU over the XBUS! We also added Basys3 support for easier teaching use.

Would be great to give Ada a small mention among the supported environments, it’s already running nicely!

Thanks for all your work on NEORV32.

Olivier

stnolting:
Hey @ohenley.

I have been following your project for some time now and am very impressed!

> Would be great to give Ada a small mention among the supported environments, it’s already running nicely!

Absolutely!

If you like, you are welcome to open a pull request and link your project. I think a link would fit well in the features list:

https://github.com/stnolting/neorv32/blob/d48913c4acf0564de5abbed13828ad69673f4e63/docs/datasheet/overview.adoc?plain=1#L77-L89

And maybe even directly on the front page:

https://github.com/stnolting/neorv32/blob/d48913c4acf0564de5abbed13828ad69673f4e63/README.md?plain=1#L30-L44

stnolting:
Completed in #1420.

PR #1420, opened 2025-11-04
Add Ada mention to READMEs

ohenley:
Reference: https://github.com/stnolting/neorv32/issues/1412


stnolting:
Thanks for the PR!

Just a note. The kite (`:kite:` = 🪁) is the logo of Zephyr, so maybe we should use a different icon here. I couldn't find a silhouette of Lady Ada here 😅, but maybe you have a cool idea: https://github.com/ikatyang/emoji-cheat-sheet

PR #1441, opened 2025-12-05
Added optinal trace log directory

jeras:
This enables running multiple RISCOF testcases in parallel, each writing the trace log into a separate testcase directory.

The generic has a default which behaves exactly like the previous version, there should be no backward compatibility issues, regressions.


jeras:
I found a better solution, will explain it elsewhere.

stnolting:
Maybe you can just instantiate the trace logger in the RISCOF testbench and ignore the processor-internal one?

https://github.com/stnolting/neorv32/blob/ee716a5a2ec1d592be19d77c537b6e074fa6e6f2/rtl/core/neorv32_tracer.vhd#L354-L363

jeras:
I did exactly that, and connected it using hierarchical paths.

PR #1453, opened 2025-12-15
Feature/hub

JesusMLastreUMA:


Unike267:
Hi Jesus!

Your work looks so interesting!

Please, let me say a couple of things:

In my opinion, you could provide a description to tell us what you are doing. It seems that you are including support to perform floating points ops through the [optimization method](https://github.com/HUBformat) that you are developing at your university. I think that you should explain a little bit what it consists of.

Aside from Stephan, the person who has contributed most to the development of the current floating point support is @mikaelsky. We can expect feedback from him if he feels like sharing his opinion. 🤔

I remember that when I developed my Master's Thesis, the NEORV32's FPU couldn't stand divisions. I've seen that you are not including this kind of operation. It would be interesting to know if your method/contribution could solve this problem.

BTW, this is an international project, so the language that we use is English. Please talk with us and make comments on the code in that language. It doesn't matter that your English isn't perfect, but we need to understand each other 😄

Cheers!

stnolting:
Hey @JesusMLastreUMA.

Could you provide some information about your code? Or was the upstream push just a mistake?

@Unike267 

Full ACK 😉

JesusMLastreUMA:
Hi @stnolting and @Unike267!

First of all, apologies for the noise! 😅 As @stnolting  suspected, this PR was opened by mistake against the upstream main branch instead of my own fork's branch. I am still getting used to the GitHub workflow.

Regarding your questions, @Unike267: I am currently working on a university research project at the University of Málaga. The project consists of modifying the FPU to implement the "FPHUB" format. This is an experimental, simplified floating-point format (Bias 128, treating NaNs/Subnormals as normal numbers range) designed to reduce hardware complexity and area while maintaining precision for specific workloads.

About the division: You are spot on! I noticed the original FPU relies on software emulation for FDIV and FSQRT. In fact, we have already developed hardware modules for these operations compliant with the HUB format. However, since they were implemented in SystemVerilog, we are currently evaluating the feasibility of integrating them into the NeoRV32 structure (as a mixed-language design) within our fork.

Sorry about the Spanish comments in the code; since this is for a local project, I kept the documentation in my native language, but I will make sure to use English for any interaction here.

I will close this PR now to clean up your dashboard. Thanks for the kind words and the feedback!

Best regards, Jesus.

stnolting:
No problem. Happens to me sometimes too.

> This is an experimental, simplified floating-point format (Bias 128, treating NaNs/Subnormals as normal numbers range) designed to reduce hardware complexity and area while maintaining precision for specific workloads.

That sounds really great! Denormalized numbers are quite annoying...

> About the division: You are spot on! I noticed the original FPU relies on software emulation for FDIV and FSQRT. In fact, we have already developed hardware modules for these operations compliant with the HUB format.

Are you planning to publish these?

PR #1466, opened 2025-12-31
Add description to TOPCMP in svd

kurtjd:
This PR just adds a description for the TOPCMP register in the svd file mainly to silence a warning [svd2rust](https://docs.rs/svd2rust/latest/svd2rust/) generates for missing register descriptions (and prevents a minor issue in generated Rust code). Just went with the description from the datasheet.

If this warrants an entry in the CHANGELOG, let me know and I'll gladly add one.

### DIFF
--- a/sw/svd/neorv32.svd
+++ b/sw/svd/neorv32.svd
@@ -462,6 +462,7 @@
           <description>Channel configuration register</description>
           <register>
             <name>TOPCMP</name>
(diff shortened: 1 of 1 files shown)


stnolting:
Thanks for the adjustment!

> If this warrants an entry in the CHANGELOG, let me know and I'll gladly add one.

No, I don't think that's necessary. :wink:

Issue #837, opened 2024-03-01
[feature request] add Rust support


Chris44442:
Recently I have tried converting a Rust elf binary to the  neorv32_application_image.vhd, by making a minor hack to the common.mk file. To my surprise it worked pretty much right away. My Rust program was very simple, essentially I only wrote to a couple of physical addresses (GPIOs) in order to toggle the LEDs on my board. But even calculating some trigonometric floating point stuff worked! But as soon as I added things like loops and more complex stuff, it didn't work anymore. My guess is that the instructions in the  neorv32_application_image.vhd were not valid for some reason. I haven't tried using the bootloader yet due to my hardware setup.

Anyway this got my hopes up, that we may have Rust support in the future in some way?

stnolting:
Hey @Chris44442!

That's an interesting idea! I've never worked with Rust before, but I see it being adopted more and more - even in tiny embedded systems (which have a focus on security).

So, yes, I would love to see this project support Rust. Unfortunately, I'm far too inexperienced to tackle it.

> To my surprise it worked pretty much right away. My Rust program was very simple, essentially I only wrote to a couple of physical addresses (GPIOs) in order to toggle the LEDs on my board.

That sounds great! Could you make the sources available some where? Maybe that would be a good starting point.

umarcor:
> My guess is that the instructions in the neorv32_application_image.vhd were not valid for some reason.

Several CPU extensions are disabled by default: https://github.com/stnolting/neorv32/blob/main/rtl/core/neorv32_top.vhd#L59-L71. You might try enabling them all.

Alternatively, when compiling your Rust program, you might try checking how to specify `-march` to avoid the compiler generating instructions which are not supported by the hardware.

Chris44442:
This is my main.rs

```rust
#![no_std]
#![no_main]

use core::panic::PanicInfo;
const GPIO_BASE: usize = 0xffff_fc08;

#[no_mangle]
pub extern "C" fn _start() -> ! {
    let gpio_output_val = GPIO_BASE as *mut u32;
    let mut result: u32 = 0;
    unsafe {gpio_output_val.write_volatile(result);}
    result = result + 5;
    unsafe {gpio_output_val.write_volatile(result);}
    result = 3 * result;
    unsafe {gpio_output_val.write_volatile(result);}
    result = result - 2;
    unsafe {gpio_output_val.write_volatile(result);}
    if result == 1 {
        unsafe {gpio_output_val.write_volatile(0x0000_0001);}
[...]

PR #1469, opened 2026-01-02
Add Rust support to user guide

kurtjd:
This PR adds a section mentioning Rust support to the NEORV32 user guide, which provides links to these crates on [crates.io](https://crates.io) (which contain further links to the repos and docs).

Related: #837


stnolting:
That's great! I've actually already received some emails asking when Rust support will be available. So, thank you for your work!

stnolting (approved):

PR #1475, opened 2026-01-14
[svd] add fields to UART_DATA

kurtjd:
Adds fields missing from `UART_DATA` in the svd file.


stnolting:
Thanks a lot!

kurtjd:
@stnolting Heads up, after merging it looks like CI fails because:

https://github.com/ghdl/ghdl/releases/download/nightly/inventory.json

No longer exists?

Issue #1478, opened 2026-01-18
`csrrci` does not seem to write to mstatus


kurtjd:
**Describe the bug**
Machine mode interrupts are globally enabled prior to this, then the following asm executes:

```asm
csrrci a1, mstatus, 0x8
csrr  a2, mstatus
andi a2, a2, 0x8
bnez a2, 0xca0c <branch>
```

The branch is taken although it shouldn't be (the results in `a1` are used elsewhere after the branch).

**To Reproduce**
Execute the assembly above.

**Expected behavior**
I would expect the branch to not be taken. The `csrrci` instruction should clear the MIE bit (bit 3) of mstatus, then the `csrr` read followed by the mask should set reg `a2` to zero. But the branch is taken anyway, showing the MIE bit was never cleared.

Interestingly, replacing the first line with `csrrci zero, mstatus, 0x8` produces expected behavior. It seems that the zero register is triggering the write to mstatus to take place, but any other register does not.

**Screenshots**
N/A

**Environment:**
 -  NEORV32 v1.12.6 (but the relevant HDL in `neorv32_cpu_control.vhd` doesn't appear to have changed much in later versions)
 - N/A
 - N/A

**Hardware:**
 - Hardware version (x.x.x.x)
 - Implemented CPU extensions, peripherals, ...
 - Hardware modifications

**Additional context**
The problem seems it could be here: https://github.com/stnolting/neorv32/blob/08d54472de757fab2c9cc881521185f73564621d/rtl/core/neorv32_cpu_control.vhd#L462

Although I'm not knowledgeable in VHDL. It looks like the write happens if `csrrw` or `csrrwi` instruction, OR if dest reg is zero, but doesn't possibly account for a `csrrci` with non-zero dest reg?

kurtjd:
Changing the linked line to:

```vhdl
(((funct3_v = funct3_csrrw_c) or (funct3_v = funct3_csrrwi_c)) or (exec.ir(instr_rs1_msb_c downto instr_rs1_lsb_c) /= "00000")) then
```
appears to fix the issue. I tried this since it seems to match what the comment is describing and matches this logic: https://github.com/stnolting/neorv32/blob/08d54472de757fab2c9cc881521185f73564621d/rtl/core/neorv32_cpu_control.vhd#L622

Not confident enough this is the proper fix since it could just happen to be working for unrelated reasons.

stnolting:
I just tested it and you're right! It's a bug. Thanks for finding it!

[...]

PR #1479, opened 2026-01-18
🐛 [cpu] fix csrr[r/c][i] instructions

stnolting:
Fixing #1478.

The `csrr[r/s][i]` instructions must not write to the respective CSR if `rs1/imm5` is **NOT** zero.

grafik

It is interesting that this bug has not been noticed yet. Apparently, these instructions are rarely used in "normal" code. However, it is unfortunate that the official RISC-V ISA tests do not check them.

Thanks again to @kurtjd for finding this bug!


kurtjd:
Thanks for the fix!

PR #1485, opened 2026-01-29
[sdi] Update docs and svd

kurtjd:
This just does some minor cleanup of the docs and svd for the SDI peripheral. Noticed the docs say the clear fifo bits are `r/-` but I'm assuming that is a typo as https://github.com/stnolting/neorv32/pull/1358/ calls them write-only (and this also matches my intuition).

Also just added a discrete field for the lower 8 bits of the data field in the SVD to match other peripherals.


stnolting:
> Noticed the docs say the clear fifo bits are r/- but I'm assuming that is a typo

Thanks for correcting this.

> Also just added a discrete field for the lower 8 bits of the data field in the SVD to match other peripherals.

Good idea!

PR #1488, opened 2026-01-31
[sdi] Fix svd and docs typos

kurtjd:
Noticed the bit for `SDI_CTRL_TX_FULL` in the svd was incorrect, so gave that a fix. While I was at it realized I missed a few typos from my last PR for SDI so just cleaned up some of those as well.


stnolting:
Thanks again for fixing!

stnolting (approved):

PR #1521, opened 2026-04-05
[lsu] add unaligned support

bewimm:
This is gated by the `Zicclsm` extension, which is mentioned in "RISC-V Profiles" (v1.0):
"Misaligned loads and stores to main memory regions with both the cacheability and coherence PMAs must be supported"
The extension explicitly mentions that unaligned accesses can be emulated, but the Linux kernel at least assumes that this extension implies HW support (https://lwn.net/Articles/978350/)

### DIFF
--- a/rtl/core/neorv32_cpu.vhd
+++ b/rtl/core/neorv32_cpu.vhd
@@ -42,6 +42,7 @@ entity neorv32_cpu is
     RISCV_ISA_Zcb       : boolean;                        -- additional code size reduction instructions
     RISCV_ISA_Zfinx     : boolean;                        -- 32-bit floating-point extension
     RISCV_ISA_Zibi      : boolean;                        -- branch with immediate
+    RISCV_ISA_Zicclsm   : boolean;                        -- misaligned loads/stores to main memory
     RISCV_ISA_Zicntr    : boolean;                        -- base counters
     RISCV_ISA_Zicond    : boolean;                        -- integer conditional operations
     RISCV_ISA_Zihpm     : boolean;                        -- hardware performance monitors
@@ -166,6 +167,7 @@ begin
       sel_string_f(riscv_zcb_c,         "_zcb",       "" ) &
       sel_string_f(RISCV_ISA_Zfinx,     "_zfinx",     "" ) &
       sel_string_f(RISCV_ISA_Zibi,      "_zibi",      "" ) &
+      sel_string_f(RISCV_ISA_Zicclsm,   "_zicclsm",   "" ) &
       sel_string_f(RISCV_ISA_Zicntr,    "_zicntr",    "" ) &
       sel_string_f(RISCV_ISA_Zicond,    "_zicond",    "" ) &
       sel_string_f(true,                "_zicsr",     "" ) & -- always enabled
@@ -445,8 +447,9 @@ begin
   -- -------------------------------------------------------------------------------------------
   neorv32_cpu_lsu_inst: entity neorv32.neorv32_cpu_lsu
   generic map (
-    HART_ID => HART_ID,  -- hardware thread ID
-    AMO_EN  => any_amo_c -- enable atomic memory accesses
+    HART_ID      => HART_ID,          -- hardware thread ID
+    AMO_EN       => any_amo_c,        -- enable atomic memory accesses
+    UNALIGNED_EN => RISCV_ISA_Zicclsm -- enable misaligned load/store support
   )
   port map (
     -- global control --
--- a/rtl/core/neorv32_cpu.vhd
+++ b/rtl/core/neorv32_cpu.vhd
@@ -44,6 +44,7 @@ entity neorv32_cpu is
     RISCV_ISA_Zcmop     : boolean                        := false;       -- compressed may-be-operations
     RISCV_ISA_Zfinx     : boolean                        := false;       -- 32-bit floating-point extension
     RISCV_ISA_Zibi      : boolean                        := false;       -- branch with immediate
+    RISCV_ISA_Zicclsm   : boolean                        := false;       -- misaligned loads/stores to main memory
     RISCV_ISA_Zicntr    : boolean                        := false;       -- base counters
     RISCV_ISA_Zicond    : boolean                        := false;       -- integer conditional operations
     RISCV_ISA_Zihpm     : boolean                        := false;       -- hardware performance monitors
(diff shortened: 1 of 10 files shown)


bewimm:
I hope this is somewhat reviewable.\
Feel free to nitpick, since I'm not super familiar with the code-base.

The motivation for making this is that I'm using the core in a project where I'm processing networking packets and the requirement to have aligned loads/stores really increases the code size there (as well as making it a lot slower).

The extension itself is a bit confusing (since it's only mentioned in the profiles doc afaik) so I'm not sure if I'm missing something.
I'm also not sure if there are other test cases I should update or run.

stnolting:
Hey @bewimm, thanks a lot for this PR!

> This is gated by the Zicclsm extension, which is mentioned in "RISC-V Profiles" (v1.0):

I had no idea there was an official ISA extension for that. On the other hand, RISC-V has an extension for just about everything.

It is probably described in A profile, right? NEORV32 is actually targeting the M profile (once it’s ratified).

Anyway. I'm not really a fan of CISC-like ISA extensions - that is, extensions that break a single instruction down into several others (like `Zcmp` -> in #1417). But just as with `Zcmp`, I realize that even an embedded-class core could really benefit from this!

> The motivation for making this is that I'm using the core in a project where I'm processing networking packets and the requirement to have aligned loads/stores really increases the code size there (as well as making it a lot slower).

Good example!

So, yeah, I’m totally in favor of it! :+1:

I just skimmed through your code. You're splitting an unaligned memory access into two consecutive narrower and aligned bus accesses, right?

I'm wondering if we could move the logic / state machine for this into a new module (let's say `neorv32_cpu_lsu_aligner.vhd` or something), which would then essentially be inserted into the data bus within the LSU. The cool thing about this would be that you could completely remove the module by disabling the according generic, effectively resulting in zero hardware overhead when the ISA extension is not used. What do you think about this?

> I'm also not sure if there are other test cases I should update or run.

Your test code looks really good, but we should take care of the hardware first.

In addition, we still need to extend the [`mxisa`](https://stnolting.github.io/neorv32/#_mxisa) CSR to enable software-based discovery of additional ISA extensions.

bewimm:
> Hey @bewimm, thanks a lot for this PR!
> 
> > This is gated by the Zicclsm extension, which is mentioned in "RISC-V Profiles" (v1.0):
> 
> I had no idea there was an official ISA extension for that. On the other hand, RISC-V has an extension for just about everything.
> 
> It is probably described in A profile, right? NEORV32 is actually targeting the M profile (once it’s ratified).

Yes, as far as I can. I found some mentions of it also here https://github.com/riscv/riscv-arch-test/blob/act4/docs/ctp/src/zi.adoc#zicclsm-misaligned-loadsstores & https://github.com/riscv/riscv-arch-test/blob/act4/coverpoints/norm/Misalign.yaml (though the naming seems a bit inconsistent)

> Anyway. I'm not really a fan of CISC-like ISA extensions - that is, extensions that break a single instruction down into several others (like `Zcmp` -> in #1417). But just as with `Zcmp`, I realize that even an embedded-class core could really benefit from this!
> 

Understandable. I saw your comment on that PR before, but decided to give it a shot anyway :sweat_smile: 

> > The motivation for making this is that I'm using the core in a project where I'm processing networking packets and the requirement to have aligned loads/stores really increases the code size there (as well as making it a lot slower).
> 
> Good example!
> 
> So, yeah, I’m totally in favor of it! 👍
> 
> I just skimmed through your code. You're splitting an unaligned memory access into two consecutive narrower and aligned bus accesses, right?
> 

Correct, but only if it actually requires two separate words.\
I have a bunch of AXI components attached to the CPU and that requires that the low address bits are consistent with wstrb for writes and always returns full words for reads (which the host then has to shift, etc...).
I'm not sure if that's the same for WB but I figured this was the cleanest approach.\
maybe generating bursts could make sense, but at least in my case they are all AXI-LITE components so bursts have to be split anyway.

> I'm wondering if we could move the logic / state machine for this into a new module (let's say `neorv32_cpu_lsu_aligner.vhd` or something), which would then essentially be inserted into the data bus within the LSU. The cool thing about this would be that you could completely remove the module by disabling the according generic, effectively resulting in zero hardware overhead when the ISA extension is not used. What do you think about this?
> 

I'm not sure I understand exactly what you mean. could you describe it a bit more?
If it helps readability then I could separate the "unaligned" vs "aligned only" code into two separate generate blocks.

> > I'm also not sure if there are other test cases I should update or run.
> 
> Your test code looks really good, but we should take care of the hardware first.
> 

makes sense

> In addition, we still need to extend the [`mxisa`](https://stnolting.github.io/neorv32/#_mxisa) CSR to enable software-based discovery of additional ISA extensions.

I'll add it for the next revision. any constraints I should keep in mind?

stnolting:
> Yes, as far as I can. I found some mentions of it also here https://github.com/riscv/riscv-arch-test/blob/act4/docs/ctp/src/zi.adoc#zicclsm-misaligned-loadsstores & https://github.com/riscv/riscv-arch-test/blob/act4/coverpoints/norm/Misalign.yaml (though the naming seems a bit inconsistent)

Thanks for the reference!

> Understandable. I saw your comment on that PR before, but decided to give it a shot anyway 😅

I'm trying to stop being so narrow-minded. 😅

> Correct, but only if it actually requires two separate words.

Ah, good point. A 16-bit/word access to, say, address 0x81 is still unaligned, but it only requires a single access.

> I'm not sure if that's the same for WB but I figured this was the cleanest approach.
maybe generating bursts could make sense, but at least in my case they are all AXI-LITE components so bursts have to be split anyway.

I don't think the bus protocol (Wishbone/AXI) matters at this point, as long as the actual bus accesses are always aligned and always involve (at most) 32 bits.

> I'm not sure I understand exactly what you mean. could you describe it a bit more?

A separate optional HDL module that handles unaligned accesses.

```
CPU
---------------------------+
                           |
LSU              optional  |
------+          Aligner   |
      |         +-------+  |
      |---REQ-->|       |----REQ-->
      |         |       |  |        BUS
      |<--RSP---|       |<---RSP---
      |         +-------+  |
------+                    |
                           |
---------------------------+
```

> I'll add it for the next revision. any constraints I should keep in mind?

Thanks for that! But I already have that in my pipeline and will implement it in the PR after the next one. 😉

bewimm:
Sorry for the long break, hopefully I can find more time for this going forward (if you're still willing to review/help/merge).

I've tried to separate it now as you suggested. It duplicates a bit of code, but hopefully it's not too bad.\
I also ran into an issue where a the second transaction in a split transfer ignores the PMP (there is a currently failing test for it). I don't really know how to best fix that, so I'd appreciate a suggestion.\
I've also not added any bits to mxisa yet.

stnolting:
> Sorry for the long break

Don't worry about it. I don't always get around to addressing issues and requests right away either...

> if you're still willing to review/help/merge

Absolutely. I still think there are definitely applications where unaligned accesses are helpful.

> I've tried to separate it now as you suggested. It duplicates a bit of code, but hopefully it's not too bad.

That's looking pretty good already!
I just need a little time to take a closer look at the code.

> I also ran into an issue where a the second transaction in a split transfer ignores the PMP (there is a currently failing test for it). I don't really know how to best fix that, so I'd appreciate a suggestion.

That's actually a problem I was expecting. I don't have a good idea yet, but I'll take a closer look at your code first.

PR #1532, opened 2026-04-24
Please initialise SPRAM array to zero

CodeWarrior1241:
## Summary

I'd like to add an explicit zero-initialiser to the storage signal inside
`neorv32_prim_spram`:

```vhdl
signal spram : ram_t := (others => (others => '0'));
```

## Motivation

Without the initialiser, `spram` starts at `'U'` in VHDL simulation.
Reads before the first write return `'U'`in simulation, which propagates into loads
and address calculations and can trigger spurious access faults in
simulators — observed in NVC during NEORV32 + newlib bring-up, where
the C runtime reads from uninitialised `.bss`/heap regions during
malloc initialisation. This was obnoxious to see in simulation until I dug into it.

Initialising to zero matches inferred-BRAM power-up behaviour on
mainstream FPGA targets, no-op for synthesis: the
initialiser folds into the default BRAM `INIT_*` strings (already 0).

## Scope

`neorv32_prim_spram` is used by `neorv32_dmem_ram`, `neorv32_imem_ram`,
and `neorv32_cache_ram`, so DMEM, IMEM (RAM mode), and cache arrays
all benefit.

## Test plan

- [x] Vivado 2025.2 / xcau15p synthesis: resource report unchanged
- [x] NVC 1.20-devel: DMEM reads of previously-unwritten addresses now
      return `0`; no more phantom access faults during crt0 bring-up

### DIFF
--- a/rtl/core/neorv32_prim.vhd
+++ b/rtl/core/neorv32_prim.vhd
@@ -186,7 +186,13 @@ end neorv32_prim_spram;
 architecture neorv32_prim_spram_rtl of neorv32_prim_spram is
 
   type ram_t is array ((2**AWIDTH)-1 downto 0) of std_ulogic_vector(DWIDTH-1 downto 0);
-  signal spram : ram_t;
+  -- Initialise to zero so simulators don't see 'U' on reads before first write.
+  -- 'U' propagates into loads/addresses and can trigger spurious access faults
+  -- (e.g. in newlib malloc's heap-metadata reads) that do not reflect real
+  -- hardware behaviour. On FPGA, inferred BRAMs power up to zero, so this
+  -- initialiser folds into the default BRAM INIT values and is a no-op for
+  -- synthesis / resource usage.
+  signal spram : ram_t := (others => (others => '0'));
   signal rdata : std_ulogic_vector(DWIDTH-1 downto 0);
 
 begin


stnolting:
I know this kind of problem all too well... Some memory areas aren't initialized properly, and the software runtime goes crazy. 🙈 

Resetting the internal RAM to zero via hardware is a good workaround, but I see a few issues here.

Not every FPGA - nor every type of BRAM - supports initialization by the bitstream (e.g., some Max 10 FPGAs or the UltraRAM blocks in AMD FPGAs). The situation is even worse with an ASIC implementation. There is, of course, no bitstream, and I don't know of a single SRAM IP that has a global reset for the memory array.

You will face the same problem if, for example, you use external DDR RAM instead of the internal RAM. Therefore, this problem should be solved where it arises: at the software level.

CRT0 actually initializes the `.bss` section to all-zero during boot.

https://github.com/stnolting/neorv32/blob/4641181ba96132316c895118a68a79620006c772/sw/common/crt0.S#L121-L130

So perhaps there is an issue with the linker script, causing the relevant memory area to not be located in `.bss`. Do you have a minimal software example that reproduces the error so I can test it?

CodeWarrior1241:
I did some testing between #1532 and #1536. The phantom `'U'` reads were `_impure_ptr` chasing wild pointers because `crt0`'s `.data` copy was pulling shifted bytes out of IMEM. Root cause was a bug in `image_gen` dropping the alignment padding the linker inserts between `.text` and `.rodata` — fixed in #1536. With #1536 applied, the FMCOMMS2 sim runs cleanly through `ad9361_init()` with SPRAM left at its default `'U'` init. Closing this in favour of the proper fix. Also — figuring out why smaller builds simulated great and "real" builds didn't was _wild_.

stnolting:
> Root cause was a bug in image_gen dropping the alignment padding the linker inserts between .text and .rodata — fixed in https://github.com/stnolting/neorv32/pull/1536.

I'll have a look at that. 👍

PR #1536, opened 2026-04-25
[image_gen] Squashed ugly silent-corruption bug — linker alignment padding between sections was being dropped from the IMEM image

CodeWarrior1241:
Identified and squashed an ugly silent-corruption bug: the image generator was concatenating ELF sections by size and silently dropping the alignment padding the linker inserts between `.text` and `.rodata`, shifting every byte after the gap by −N in ROM and corrupting boot.

## Summary

`sw/image_gen/image_gen.c` builds the IMEM/BOOTROM image by raw concatenation of the ELF's `.text`, `.rodata`, and `.data` sections using each section's *size*, ignoring each section's *LMA*. If the linker inserts alignment padding between sections, that padding is silently dropped from the image and every byte after the gap is shifted by −N. The CPU then reads `.rodata` at the wrong offset; `crt0`'s `.data` copy pulls the wrong bytes from `LOADADDR(.data)`; newlib's `_impure_ptr` and the malloc heap metadata get corrupted; boot dies in an RTE access fault.

The offending lines on `main` are:

```c
memcpy(raw_image,                           text,   text_size);   // start with .text
memcpy(raw_image + text_size,               rodata, rodata_size); // append .rodata
memcpy(raw_image + text_size + rodata_size, data,   data_size);   // append .data
```

## Trigger condition

`.rodata`'s required alignment > `.text`'s end alignment.

The default linker script (`sw/common/neorv32.ld`) declares `.rodata : ALIGN(4)`, but the output section's actual alignment is `max(ALIGN(4), max(input section alignment))`. So as soon as the wildcard `*(.rodata .rodata* .rodata.* ...)` pulls in a `.rodata.cst8` input section (8-byte alignment) — which happens with:

- any `double` constant in code,
- `printf`/`scanf` with `%f`/`%lf` (newlib's float format tables are 8-byte aligned),
- merged `.rodata.cst8` constants,

…and `.text` happens to end at a 4-byte-but-not-8-byte boundary, the linker inserts 4 bytes of padding and the bug fires. Whether you hit it depends on `.text`'s size parity, so two compiles of "almost the same" program can behave differently — which is exactly why it's so nasty. Simple bare-metal blinkies that never touch float printf will likely never trip it; anything pulling in newlib heap/stdio is at risk.

## Fix

Walk the ELF program headers, resolve each section's LMA via `p_paddr + (sh_addr - p_vaddr)`, and lay sections out in a zero-filled image at their LMA offsets. `.text`/`.rodata`/`.data` `sh_addr` fall through PT_LOAD lookup; if no PT_LOAD covers a section the LMA defaults to its VMA. A diagnostic prints the gap size when one is detected so silent regressions surface at image-gen time.

Behaviour is byte-identical to the old generator when no gap exists.

## Test plan

- [x] Build a program that does *not* pull in `.rodata.cst8` (e.g. `sw/example/demo_blink_led`) — image must be byte-identical to pre-fix output.
- [x] Build a program that *does* pull in `.rodata.cst8` (any newlib float printf, or a TU containing a `double` constant). Pre-fix: silent corruption / RTE fault on boot. Post-fix: clean boot, and the gap diagnostic prints at image-gen time.
- [x] CI sanity (existing image-gen invocations across the standard examples should be unaffected).

### DIFF
--- a/sw/image_gen/image_gen.c
+++ b/sw/image_gen/image_gen.c
@@ -201,6 +201,7 @@ int main(int argc, char *argv[]) {
   unsigned int text_size = 0;
   unsigned int rodata_size = 0;
   unsigned int data_size = 0;
+  uint32_t text_vma = 0, rodata_vma = 0, data_vma = 0;
 
   // scan section headers
   for (i = 0; i < elf.e_shnum; i++) {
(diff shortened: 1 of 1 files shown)


stnolting:
I hadn't realized that there could also be gaps between sections. Apparently, I haven't come across a program where that was the case so far.

Anyway, thanks for the fix! I'll take a closer look at it.

I do wonder if objdump couldn't simply generate a "flat binary" which would then just be converted to the appropriate format by the image generator. Would that allow us to move all the ELF handling back to GCC? 🤔

CodeWarrior1241:
This makes sense in the long term, would have prevented the bug entirely. Two things I'm seeing:

1. The exe header's `base_addr` currently comes from `e_entry` — would need a CLI flag or a hardcoded value (it's effectively `__neorv32_rom_base` per build).
2. It's a CLI-breaking change for anyone running image_gen directly on `.elf` files. Is that important?

Bonus: the `#ifdef __APPLE__` / `<libelf.h>` macOS workaround in image_gen.c also goes away.

My thinking was that this PR goes in as an immediate fix, and I can open a separate PR that wires `objcopy -O binary` into common.mk and strips the ELF handling out of image_gen. What do you think?

stnolting:
> 1. The exe header's base_addr currently comes from e_entry — would need a CLI flag or a hardcoded value (it's effectively __neorv32_rom_base per build).

That is actually a problem. Using `readelf -h` gives the base address as a hexadecimal number. With a little console magic, you could extract that and pass it as an additional argument (actually only for the EXE format) to the image generator.

> 2. It's a CLI-breaking change for anyone running image_gen directly on .elf files. Is that important?

Yeah, that's too bad. But it wouldn't be the first breaking change (and it won't be the last, either). So I think that would be okay.

I was just about to clean up the image generator anyway. I'd put that, along with the "flat binary" thing, into a new branch. Could you help with the setup and testing?

CodeWarrior1241:
> I was just about to clean up the image generator anyway. I'd put that, along with the "flat binary" thing, into a new branch. Could you help with the setup and testing?

Neat — happy to help. What do you think about hitting these two changes:

**`common.mk`**
- Add `OBJCOPY = $(RISCV_PREFIX)objcopy` next to the existing `OBJDUMP`/`READELF`.
- New intermediate target (e.g. `main.flat.bin`) via `$(OBJCOPY) -O binary $< $@`. All `image_gen` rules (`$(APP_EXE)`, `$(APP_VHD)`, `$(APP_BIN)`, `$(APP_COE)`, `$(APP_MIF)`, `$(APP_MEM)`, and `bl_image`) feed the flat binary instead of the ELF.
- For `-t exe` only: `ENTRY=$$($(READELF) -h $(APP_ELF) | awk '/Entry point/ {print $$NF}')`, pass as `--base-addr $$ENTRY`.
- Add the flat binary to the `clean` rule.

**`image_gen.c`**
- Drop `<libelf.h>` (and the `__APPLE__` shim with it), all the `Elf32_*` parsing, `read_section()`, the PT_LOAD/LMA walker.
- Input becomes a single `fread` of the flat buffer.
- New `--base-addr <hex>` arg, required for `-t exe`, error otherwise.
- Check that format emitters didn't change.

Questions:

1. Branch base: off `main` after #1536 merges, or layered on `fix-image-gen-lma-gap`?
2. Do you want to drive the C/Makefile changes and have me test, or the other way around?
3. `objcopy -O binary` zero-fills inter-section gaps on RISC-V — I want to confirm with a hexdump that its layout matches what #1536 produces byte-for-byte before we commit to the rewrite.

I'm gonna test it by:

- Byte-compare every output format (`exe`, `raw_exe`, `vhd`, `coe`, `mif`, `mem`) old-vs-new on a no-gap program (`demo_blink_led`) and a gap-bearing one (anything pulling `.rodata.cst8` — `double` literal or `printf %f`). Both must match `cmp`-clean against post-#1536 output.
- `xxd -l 12 neorv32_exe.bin` to verify `base_addr` matches `readelf -h` entry.
- Boot the gap-bearing program on real hardware (the original #1536 regression test).
- `make bl_image` and re-synth bootloader.
- I'll test image_gen on Linux, don't have a Mac to test the `__APPLE__` removal and will mark it as untested.

Push the branch when ready and I'll start on the test pass.

stnolting:
> Neat — happy to help.

Great, thanks!

> Add OBJCOPY = $(RISCV_PREFIX)objcopy next to the existing OBJDUMP/READELF.

Check.

> New intermediate target (e.g. main.flat.bin)

Check: `APP_FLT`. Maybe we should change that name.

> For -t exe only: ENTRY=$$($(READELF) -h $(APP_ELF) | awk '/Entry point/ {print $$NF}'), pass as --base-addr $$ENTRY.

I don't know how to handle `awk`, so I tried with `sed`.

> Add the flat binary to the clean rule.

Check.

> Drop <libelf.h>

Check.

> Input becomes a single fread of the flat buffer.

Check.

> New --base-addr <hex> arg, required for -t exe, error otherwise.

Check. It's `-b address` now.

> Check that format emitters didn't change.

The VHDL targets seem fine. However, the other targets still need testing.

> I'm gonna test it by:

Thanks a lot again! :)

> Push the branch when ready and I'll start on the test pass.

-> #1541

stnolting:
Superseeded by #1541.

PR #1540, opened 2026-04-26
Two write-back D-cache correctness fixes (fence.i + AMO bypass)

14sea:
Two coherence bugs in the cache architecture introduced by #1513
(write-back & write-allocate). Found while bumping the `neorv32`
submodule in [see_neorv32_run_linux](https://github.com/14sea/see_neorv32_run_linux)
from `v1.12.9` to `origin/main` for an FPGA Linux boot.

The bugs are independent and small. Squashed into one PR because they
both stem from the same write-back-vs-bypass interaction and neither
boots Linux without the other; happy to split into two PRs if preferred.

---

## 1. `fence.i` must drain the D-cache (Zifencei)

**File:** `rtl/core/neorv32_cpu_control.vhd`, `opcode_fence_c` decode.

Zifencei requires that prior stores to instruction memory are made
visible to subsequent instruction fetches. The current decode asserts
`if_fence` (I-cache invalidate) but **not** `lsu_fence` (D-cache drain)
for `fence.i`:

```vhdl
when opcode_fence_c =>
  ctrl_nxt.lsu_fence <= not exec.ir(instr_funct3_lsb_c);  -- '0' for fence.i
  ctrl_nxt.if_fence  <=     exec.ir(instr_funct3_lsb_c);
```

Pre-#1513 this was harmless (write-through cache, no dirty lines).
Post-#1513 dirty stores can sit in the D-cache indefinitely, so the
I-cache refill reads stale memory and the CPU executes garbage.

### Bisect

`git bisect` between `v1.12.9` (good) and `origin/main` (bad) on a
write-then-jump-to-RAM testcase identifies `f774dac6 [cache] first
draft of write-back architecture` as the first bad commit. The bus
reservation logic, CPU control, and LSU are byte-identical across the
bisect window — only the cache write policy changed.

### Reproducer

1. Build a NEORV32 SoC with `ICACHE_EN=true`, `DCACHE_EN=true`,
   write-back cache (current default), and external memory via XBUS.
2. From software running in IMEM/local memory: write a few RV32
   instructions to a cached external-RAM address, issue `fence.i`,
   `jalr` to that address.
3. CPU hangs on the first I-fetch from external RAM.

Disabling D-cache makes the test pass — confirms the dirty stores
never reach memory before the I-cache refills.

### Fix

Always assert `lsu_fence` for `opcode_fence_c`; leave `if_fence` gated
on `funct3 LSB` so plain `fence` still does not bother I.

```vhdl
when opcode_fence_c =>
  ctrl_nxt.lsu_fence <= '1';                            -- always drain D
  ctrl_nxt.if_fence  <= exec.ir(instr_funct3_lsb_c);    -- only fence.i invalidates I
```

One-line cost; no FF/LE change.

---

## 2. AMO/uncached bypass must coordinate with dirty cache lines

**File:** `rtl/core/neorv32_cache.vhd`, `S_CHECK` / `S_BYPASS` /
`S_WRITE_DONE`.

AMO and uncached requests bypass the cache (`pnd_bp = 1` ->
`S_BYPASS`). With a write-back D-cache that breaks coherence in two
directions:

1. **Bypass-read sees stale memory.** A regular store can leave the
   target word dirty in the cache; the AMO bypass reads memory and
   gets the pre-store value.
2. **Cache holds stale value after bypass-write.** The AMO bypass
   writes memory; the cache still holds the pre-AMO value, so a later
   plain load returns it.

Linux exercises this on every spinlock / atomic / rwsem update.
Without the fix the kernel hits `bad: scheduling from the idle thread!`
loops and lock deadlocks during early boot; with D-cache disabled the
same kernel boots cleanly.

### Fix

Before falling through to `S_BYPASS` from `S_CHECK`, look at the cache
line at the same index:

| State                | Action                                                                         |
| -------------------- | ------------------------------------------------------------------------------ |
| HIT, dirty           | Write the line back (`S_WRITE_START` -> `S_WRITE_DONE`), invalidate, then bypass. |
| HIT, clean           | Invalidate the line inline, then bypass. (Memory is already current.)          |
| MISS                 | Bypass directly. (No cache state to repair.)                                   |

`pnd_bp` becomes a sticky flag (default `<= ctrl.pnd_bp` instead of
`<= '0'`) so it survives the write-back detour, and is cleared in
`S_BYPASS` on ACK. `S_WRITE_DONE` gains an `elsif (ctrl.pnd_bp = '1')`
branch that invalidates the line and re-arms the bypass.

`READ_ONLY=true` (I-cache) is untouched: there is no bypass-write case
to fix, and the I-cache never has dirty lines to write back.

### Cost

A handful of LE for the new `S_WRITE_DONE -> S_BYPASS` transition and
one extra mux on `pnd_bp`. No new RAM, no new ports.

---

## Validation

- Both patches apply cleanly on top of current `origin/main` HEAD
  (`e0739e63`); branch synthesizes for Cyclone-IV EP4CE6 with `0`
  errors / standard warnings only.
- The full diff (these two fixes plus a kernel-side LR/SC reservation
  fix and `DCACHE_EN=false` workaround for a separate non-burst-SDRAM
  performance issue) lets Linux 6.6.83 boot to a userspace shell on a
  Cyclone-IV EP4CE6 FPGA in ~36 s wall time.
- Tested on a Heijin AX301 board (50 MHz, 32 MB SDRAM, non-burst
  controller).

Both commits include detailed commit messages with bisect / repro /
fix structure for direct cherry-picking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 29.04.2026 | 1.13.0.1 | rework cache handling of atomic memory operations: enforce memory synchronization by hardware | [#1540](https://github.com/stnolting/neorv32/pull/1540) |
 | 27.04.2026 | [**1.13.0**](https://github.com/stnolting/neorv32/releases/tag/v1.13.0) | :rocket: **New release** | |
 | 25.04.2026 | 1.12.9.9 | `C` ISA extension: decompressor cleanups and logic optimizations (dead code elimination) | [#1537](https://github.com/stnolting/neorv32/pull/1537) |
 | 24.04.2026 | 1.12.9.8 | minor rtl (corner case) fixes and logic optimizations | [#1534](https://github.com/stnolting/neorv32/pull/1534) |
--- a/docs/datasheet/cpu_isa.adoc
+++ b/docs/datasheet/cpu_isa.adoc
@@ -290,9 +290,11 @@ Example: `MARCH=rv32imc_zca_zcb_...`
 
 ==== `Zifencei` ISA Extension
 
-This instruction is the only standard mechanism to ensure that stores visible to a hart will also be visible to its
-instruction fetches. When executed, the CPU flushes the instruction prefetch buffer and reloads the CPU's
-instruction cache (if enabled). See section <<_memory_coherence>> for more information.
+This instruction is the only standard mechanism to ensure that stores visible to a hart will also be visible to
+its instruction fetches. When executed, the CPU flushes the data cache (if enabled) to main memory, clears the
+instruction prefetch buffer and invalidates the instruction cache (if enabled). The next instruction fetch request
+from the CPU will result in an instruction cache miss that will return up-to-date data from main memory.
+See section <<_memory_coherence>> for more information.
 
 .Instructions and Timing
 [cols="<2,<5,<4"]
--- a/docs/datasheet/cpu_isa.adoc
+++ b/docs/datasheet/cpu_isa.adoc
@@ -101,14 +101,9 @@ extension is shorthand for the following set of other extensions:
 * <<_zaamo_isa_extension>> - Atomic read-modify-write instructions.
 * <<_zalrsc_isa_extension>> - Atomic reservation-set instructions.
 
-.Atomic Variable in C Languages
-[NOTE]
-The C `_Atomic` specifier should be used for atomic variables.
-See section <<_coherence_example>> for more information.
-
 .`A` ISA Extension Exceptions
 [NOTE]
-Note that load and load-reserved instructions generate load exceptions,
+Note that according to the RISC-V spec. load and load-reservate instructions generate load exceptions,
 whereas store, store-conditional, and AMO instructions generate store exceptions.
 
 
@@ -242,6 +237,9 @@ are handled by the processor's <<_atomic_memory_operations_controller>>.
 .`aq` and `rl` Bits
 [NOTE]
 The instruction word's `aq` and `lr` memory ordering bits are not evaluated by the hardware at all.
+However, the CPU and especially the <<_data_cache_dcache>> treats all atomic memory operation as if
+the `aq` and `lr` bits were set (for the targeted address). See <<_memory_coherence>> for more
+information.
 
 
 ==== `Zalrsc` ISA Extension
@@ -262,6 +260,9 @@ are handled by the processor's <<_atomic_reservation_set_controller>>.
 .`aq` and `rl` Bits
 [NOTE]
 The instruction word's `aq` and `lr` memory ordering bits are not evaluated by the hardware at all.
+However, the CPU and especially the <<_data_cache_dcache>> treats all atomic memory operation as if
+the `aq` and `lr` bits were set (for the targeted address). See <<_memory_coherence>> for more
+information.
 
 
 ==== `Zcb` ISA Extension
--- a/docs/datasheet/soc.adoc
+++ b/docs/datasheet/soc.adoc
@@ -652,7 +652,7 @@ that are executed by processor-external agents.
 :sectnums:
 ==== Memory Coherence
 
-Depending on the configuration, the NEORV32 memory hierarchy consists of several levels.
+Depending on the configuration, the NEORV32 memory hierarchy provides several levels/layers of memory.
 
 [start=1]
 . The CPU execution pipeline and its _instruction prefetch buffer_ (<<_cpu_front_end>>)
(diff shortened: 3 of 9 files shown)


14sea:
Thanks Copilot — both findings are correct. Pushed an amend to the AMO commit (`3352cd97` -> `ca1a4350`) addressing them belt-and-suspenders:

1. **Default `pnd_bp` cleanup in `S_IDLE`** (review comment 1 / line 163):
   ```vhdl
   ctrl_nxt.pnd_bp <= '0' when (ctrl.state = S_IDLE) else ctrl.pnd_bp;
   ```
   So any path that returns to `S_IDLE` (bus error, future early-terminate paths, etc.) drops stale bypass state before the next request enters `S_CHECK`. `S_IDLE` itself still sets `pnd_bp <= '1'` explicitly for uncached/AMO requests, so this default is safe there.

2. **Explicit `pnd_bp <= '0'` in `S_WRITE_DONE` bus-error branch** (review comment 2 / line 495):
   Aborts the now-meaningless bypass when the writeback fails, instead of relying solely on (1). Makes the intent obvious to readers and removes a one-cycle window where `pnd_bp` would still be observable in the IDLE-entry combinational logic.

GHDL `tb_bus_amo_rvs` still passes; bitstream rebuilt; kernel boot still reaches `nommu#`.

stnolting:
Thanks for another great bug report (and fix)!

It's clear that I-cache flushes require a D-cache flush first. I had already suspected as much. 🤔 

The part about atomic accesses and cache-line flushes makes sense, too. Have you tested the timing? Somehow I get the feeling that the additional hit/miss/dirty checks aren't good for the critical path...

Anyway. If it's okay with you, I'd like to tweak a few minor things. I like the detailed comments, but maybe a one-liner plus a link to this PR would suffice. 😉

14sea:
Thanks for the kind reply!

Honestly, **no — I haven't run any timing analysis on this**. I only verified it builds in Quartus for my AX301 (50 MHz, EP4CE6) target and that the kernel boots; I didn't look at Fmax or the critical-path report at all. Your concern is very likely well-founded — adding hit/miss/dirty checks into the AMO bypass path almost certainly costs something on the critical path, and I'm not in a position to evaluate the impact across the configurations you care about.

To be upfront: this patch is really a **minimal "make my own project boot" workaround**, not a properly engineered fix. I'd rather not have it land on `main` as-is and impose my hacks on the wider project framework. The reason I sent it as a PR rather than a plain issue is just that the interaction (write-back D-cache vs. fence.i / vs. AMO bypass) was hard to describe cleanly in prose — the diff was easier to point at.

So please feel free to **close this PR** (or take it purely as a reproducer) and replace it with whatever proper fix fits the architecture — one-liner comments, different structure, or an entirely different approach. I'll happily re-test any candidate fix on the AX301 + Linux boot path and report back. Looking forward to your version!

stnolting:
I ran both your original version and my modifications through the synthesis. Neither version affects the timing of the cache module.

However, I've made a few more adjustments. The AMO logic is actually pretty complicated at this point... I've also run a few tests, and they look good so far. It would be great if you could test the changes on your end again. Then I'll update the documentation accordingly.

14sea:
Thanks for the lightning-fast turnaround!

Verified `1b183040` on AX301 + nommu Linux:
- Quartus build clean (EP4CE10).
- `DCACHE_EN=false`: boots to `nommu#` shell in ~36 s — matches our pre-PR baseline.
- `DCACHE_EN=true`: your `fence.i` drain + AMO bypass coherence paths are exercised end-to-end through early init (`Calibrating delay loop`, `devtmpfs initialized`, `Switched to clocksource clint_clocksource` all pass), then hangs at `subsys_initcall` #04.

The `DCACHE_EN=true` hang is independent of your fix — it's a `ktime_get_coarse_real_ts64` seqcount-retry livelock. Our `wb_sdram_ctrl.v` is non-burst (`CACHE_BURSTS_EN=false`), so every dirty eviction is 16 single-word stores instead of one burst transaction; reads on hot kernel paths get slow enough that the seqcount window spans a timer IRQ and `read_seqcount_retry` never converges. We get the write-back cost without the burst upside. Adding burst support to the SDRAM controller is on our side, not yours — so we keep `DCACHE_EN=false` in our project, where your patches are runtime no-ops but correctness is verified by a clean kernel boot.

Ready to merge from our side. Thanks again!

stnolting:
Thanks for the feedback!

I've updated the documentation and run a few tests of my own. While doing so, I realized that it might be a good idea to actually use the `aq` (acquire) and `rl` (release) bits of the AMO instruction. Especially the `rl` bit - this could be used to automatically trigger a full d-cache flush before the AMO instruction. But that might be something for another PR.

> Adding burst support to the SDRAM controller

I'm just curious: what SDRAM controller do you use?

14sea:
Awesome — thanks for merging and for the v1.13.0.1 bump!

**`aq`/`rl` for AMO d-cache coherence:** agreed, that's a clean follow-up. Using `rl=1` to trigger the flush is spec-aligned with release-consistency, and it lets the common `rl=0` AMO paths (e.g. plain `atomic_add` without ordering) skip the flush entirely. Worth a quick measurement of how often the kernel actually emits `rl=0` vs `rl=1` AMOs — RV Linux uses both, so the win could be material on hot atomic-counter paths. Happy to hardware-test on AX301 + nommu Linux whenever you open that PR.

**SDRAM controller — it's a custom one in this repo, not the upstream `wb_sdram` core:**
- [`rtl/sdram_ctrl.v`](https://github.com/14sea/see_neorv32_run_linux/blob/main/rtl/sdram_ctrl.v) — HY57V2562GTR (32 MB, 16-bit) on the Heijin AX301 board, 50 MHz, CL=3, **BL=1 (no bursts)**.
- [`rtl/wb_sdram_ctrl.v`](https://github.com/14sea/see_neorv32_run_linux/blob/main/rtl/wb_sdram_ctrl.v) — small XBUS/Wishbone bridge.

Each 32-bit word costs two single-word 16-bit SDRAM accesses with full row/col addressing each time, so a dirty line eviction is 16 individual stores. That's why `CACHE_BURSTS_EN=false` and why the seqcount-livelock kicks in on hot kernel paths with `DCACHE_EN=true`. Fixing it on our side really does mean adding burst support (or moving to a burst-capable SDRAM core), nothing on you.

On our side, next step is rebasing our submodule pin to post-merge `main` and dropping the now-upstream `0001`/`0002` patches. Will hardware-verify before bumping. Thanks again for the fast turnaround and the cleanups along the way!

PR #1541, opened 2026-04-28
[image_gen] Pass a flattened ELF binary instead of parsing the ELF itself

stnolting:
* use `objcopy` to generate a flattened binary from the ELF
* the image generator uses this binary to convert it to different format
* no more ELF parsing by the generator itself
* fixes portential sections offsets and gaps

(based on #1536)


CodeWarrior1241:
Results — pulled `pr-1541` into a clean checkout and ran against both a no-gap program (`demo_blink_led`) and a gap-bearing variant (added a `static const double[]` referenced from `main()` to force `.rodata.cst8`; verified a real 4-byte LMA gap shows up between `.text` and `.rodata`).

**The rewrite works.** Compared against post-#1536 baseline output:

| format | demo_blink_led | gap-bearing |
| --- | --- | --- |
| `vhd` | `cmp` MATCH | `cmp` MATCH |
| `bin` | `cmp` MATCH | `cmp` MATCH |
| `coe` | `cmp` MATCH | `cmp` MATCH |
| `mif` | `cmp` MATCH | `cmp` MATCH |
| `mem` | `cmp` MATCH | `cmp` MATCH |
| `exe` | mismatch ⚠ | mismatch ⚠ |

Also checked: the `objcopy -O binary` flat output (`elf.bin`) is byte-identical to #1536's `neorv32_raw_exe.bin` for the gap-bearing program — `objcopy` zero-fills the inter-section gap exactly as the in-process LMA walker did. So things worked out.

**One bug to fix** — `$(APP_EXE)` rule:

```make
$(APP_EXE): $(APP_ELF) $(APP_FLT) $(IMAGE_GEN)
	...
	$(Q)$(IMAGE_GEN) -t exe -b $(shell ...) -i $< -o $@
```

`$<` is the first prereq, which is `$(APP_ELF)`, so image_gen gets fed `main.elf` instead of the flat binary. The 796-byte demo produces a 42056-byte `neorv32_exe.bin` whose payload starts with `7f 45 4c 46` (ELF magic). VHD/bin/coe/mif/mem rules all have `$(APP_FLT)` as the first prereq so they're unaffected — that's why VHD looked fine.

One-line fix: change `-i $<` to `-i $(APP_FLT)` (or reorder the prereqs). After patching by hand, `cmp` against the post-#1536 baseline EXE is clean.

**Cosmetic** — the help text example:
```
image_gen -i elf.bin -o main_exe.bin -b A0000000 -t exe
```
`strtoul(arg, NULL, 0)` parses non-`0x`-prefixed input as decimal and silently returns 0 if it has no leading digit. The makefile path is fine (`readelf` emits `0x0`-style), but the example as written would silently produce a zero base addr. Maybe `-b 0xA0000000` in the help.

**Untested:** macOS build (no Mac on hand, will mark as untested in the PR), hardware boot of the gap-bearing image.

stnolting:
> The rewrite works. Compared against post-https://github.com/stnolting/neorv32/pull/1536 baseline output:

Thanks for testing!

> objcopy zero-fills the inter-section gap exactly as the in-process LMA walker did. So things worked out.

:+1:

> One bug to fix — $(APP_EXE) rule: [...]

Oh, right. Thanks for spotting that. I'll fix that.

> strtoul(arg, NULL, 0) parses non-0x-prefixed input as decimal and silently returns 0 if it has no leading digit. The makefile path is fine (readelf emits 0x0-style), but the example as written would silently produce a zero base addr. Maybe -b 0xA0000000 in the help.

Good point. I think we can add another test using the third argument (now NULL)... I'll take another look at it.

PR #1558, opened 2026-05-20
[slink] Assert ready signal only when enabled

tkundrat:
The slink_rx_ready_o signal of the RX FIFO was
always the same as the rx_fifo.free signal.
When the signal rx_fifo.clr to clear the FIFO is
asserted, free will be asserted aswell.
As the clear signal is defined as `not ctrl.enable` the slink_rx_ready_o signal is always asserted,
when the RX module is NOT enabled.
If a module attached to the SLINK interface wants
to send data to it while the RX module is not
enabled, it will send them anyway, as the ready
signal is asserted. Thus the data sent to the
RX module during this state will be dismissed.

My change changes this concurrent signal assignment to `rx_fifo.free and ctrl.enable`, so data will
only be received when the RX fifo is free AND the
module is enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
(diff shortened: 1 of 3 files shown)


stnolting:
You're right, that's a bug.
Thanks for fixing!

stnolting (approved):

PR #1564, opened 2026-05-31
[sw/example] fixup twi demo SDL & SCL signal sense

NikLeberg:
Hi there fabulous @stnolting.

First of all:
 - Huge respect to the amount of work you pour into this absolute fantastic beast of a softcore every day. I had to fast-forward my stale fork by about 1000 commits. Insane! :smile:
 - Oh and congrats on more than 1000 closed PRs! :fireworks: :rocket:

I've noticed that in `demo_twi` the sensing of the SDL and SCL lines was a bit weird. I guess `neorv32_twi_sense_{scl,sda}()` used to return a 1?. Instead now it directly returns the bit from the register which is `> 1`. So I've changed the demo such that it considers `== 0` as `0` and anything else as `1`. This then fixes the erroneous warning and it now prints nice values when `sense`ing via the command line.

Cheers,
Nik

### DIFF
--- a/docs/datasheet/soc.adoc
+++ b/docs/datasheet/soc.adoc
@@ -782,7 +782,7 @@ However, only the specified memory-mapped registers addresses should be used.
 
 .Unimplemented Modules / Address Holes
 [NOTE]
-When accessing an IO device tha hast not been implemented (disabled via the according generic)
+When accessing an IO device that has not been implemented (disabled via the according generic)
 or when accessing an address that is actually unused, a load/store access fault exception is raised.
 
 .Writing to Read-Only Registers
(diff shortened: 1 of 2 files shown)


stnolting:
Hey @NikLeberg!
Great to hear from you, and thanks for the kind words! :)

> So I've changed the demo such that it considers == 0 as 0 and anything else as 1. This then fixes the erroneous warning and it now prints nice values when senseing via the command line.

At some point, the sensing mechanism changed for some reason, and I (or whoever) forgot to update it everywhere. Anyway, thanks for the fix!

stnolting (approved):

PR #1603, opened 2026-07-21
[cpu] Add optional CPU_FAST_MUL_REG generic for the fast multiplier

CodeWarrior1241:
# Add optional pipeline register for the fast (DSP) multiplier

## Summary

Solves issue that came up with using non-Xilinx and non-Altera FPGAs: 
their DSP slices are much slower. Adds an opt-in generic 
**`CPU_FAST_MUL_REG`** (default `false`) that inserts
a second pipeline register into the DSP-based fast multiplier (`neorv32_prim_mul`),
with matching one-cycle wait state in the `M`/`Zmmul` control FSM. 
Default is off, so every existing configuration is bit-identical.

## Motivation

`neorv32_prim_mul` has a single register after the `opa * opb` product. On FPGAs
whose DSP blocks are narrower than the 32-bit operands, that product is inferred
as a **cascade** of DSP slices - e.g. three 18×18 MACCs on Microchip PolarFire,
versus two 27×18 DSP48E2s (one cascade hop) on Xilinx UltraScale+. With only one
register the whole cascade is a single combinational path.

On a test PolarFire MPF300 (−1 speed grade) at 125 MHz this multiplier path was the
worst setup path at **8.66 ns**, ~6.8 ns of it inside the DSP hard
macros. It's fixed delay that placement/routing/retiming cannot reduce (retiming
relocates registers, it cannot create the second one a 3-deep cascade needs).

Enabling the pipeline lets synthesis retime a register into the DSP cascade
(the DSP's own output register), which dropped the multiplier path to ~4.95 ns
and moved the CPU clock domain from **−0.99 ns (fails, ~112 MHz) to +1.14 ns
(meets 125 MHz)**, post-layout / multi-corner. This is a portability fix for
narrow-DSP families (PolarFire, Lattice ECP5/Nexus, …);
wide-DSP targets are unaffected at the default.

## Cost

`T_mul_latency` goes 1 → 2 when enabled, i.e. `mul`/`mulh`/`mulhsu`/`mulhu` take
5 instead of 4 cycles (datasheet formula `3 + T_mul_latency`). No ISA change, no
software change (multiply is fixed-latency behind the existing valid handshake);
division, area, and all other instructions are unaffected.

## Why it is correct

The CPU already stalls the pipeline on a co-processor completion handshake
(`valid_o` → `done_o`), not on a fixed cycle count. This is why the same core
tolerates a 4-cycle fast multiply, a 35-cycle serial multiply, a divide, and a
CFU of up to 512 cycles. Adding a cycle needs no new control interface.

The extra register stays correct because latency and `valid` move together: the
new `res_pipe` register adds one data-delay cycle, and the new `S_PIPE` state
adds exactly one cycle before `valid_o`/`out_en` assert in `S_DONE`. So when
`valid_o` fires, the product has fully propagated through the register. (Adding
the register *without* `S_PIPE` would fire `valid_o` a cycle early — that is the
bug this state prevents.) The FPU, which also uses `neorv32_prim_mul` but times
it with a fixed-length shift register rather than a handshake, keeps
`PIPELINE = false` and is untouched.

## Verification

- **`sw/example/processor_check`** (full M-extension self-check, verifies
  multiply *results*) run under NVC with `CPU_FAST_MUL_REG = true`:
  **PASS 62/62, FAIL 0/62 — `PROCESSOR CHECK COMPLETED SUCCESSFULLY!`**
- **`sw/example/performance_tests/M`** (instruction timing), baseline vs.
  pipelined, same RTL and image, generic toggled at elaboration:
  `mul`/`mulh`/`mulhsu`/`mulhu` 4 → 5 cycles; `div`/`divu`/`rem`/`remu`
  unchanged at 35 — change isolated to the fast multiplier.

## Files

RTL: `neorv32_prim.vhd`, `neorv32_cpu_alu_muldiv.vhd`, `neorv32_cpu_alu.vhd`,
`neorv32_cpu.vhd`, `neorv32_top.vhd`, `neorv32_package.vhd` (component
declaration + `hw_version_c` bump to `1.13.2.9`).
Docs: `CHANGELOG.md`, `docs/datasheet/{soc,cpu,cpu_isa}.adoc`.
Testbench: `sim/neorv32_tb.vhd` exposes the generic (default `false`) so it can
be toggled for verification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 20.07.2026 | 1.13.2.9 | :sparkles: add optional `CPU_FAST_MUL_REG` generic: pipeline register for the fast multiplier (shortens critical path on narrow-DSP FPGAs; +1 multiply latency cycle) | [#1603](https://github.com/stnolting/neorv32/pull/1603) |
 | 20.07.2026 | 1.13.2.8 | comprehensive VHDL coding style edits | [#1602](https://github.com/stnolting/neorv32/pull/1602) |
 | 17.07.2026 | 1.13.2.7 | :sparkles: add new SoC module: serial memory controller (SMC) to attach external serial memory (PSRAM or flash; supports XIP) | [#1599](https://github.com/stnolting/neorv32/pull/1599) |
 | 11.07.2026 | 1.13.2.6 | :bug: fix another RVFI memory address & data signal alignment bug | [#1597](https://github.com/stnolting/neorv32/pull/1597) |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
-| 20.07.2026 | 1.13.2.9 | :sparkles: add optional `CPU_FAST_MUL_REG` generic: pipeline register for the fast multiplier (shortens critical path on narrow-DSP FPGAs; +1 multiply latency cycle) | [#1603](https://github.com/stnolting/neorv32/pull/1603) |
+| 23.07.2026 | 1.13.2.9 | :sparkles: add optional `CPU_FAST_MUL_REG` generic: pipeline register for the fast multiplier (shortens critical path on narrow-DSP FPGAs; +1 multiply latency cycle) | [#1603](https://github.com/stnolting/neorv32/pull/1603) |
 | 20.07.2026 | 1.13.2.8 | comprehensive VHDL coding style edits | [#1602](https://github.com/stnolting/neorv32/pull/1602) |
 | 17.07.2026 | 1.13.2.7 | :sparkles: add new SoC module: serial memory controller (SMC) to attach external serial memory (PSRAM or flash; supports XIP) | [#1599](https://github.com/stnolting/neorv32/pull/1599) |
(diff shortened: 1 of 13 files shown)


CodeWarrior1241:
Hi @stnolting ,

Thank you for taking a look. I'm trying to add a register in to your multiplier, this really helped with timing on PolarFire. Please let me know what you think.

stnolting:
Hey @CodeWarrior1241,

I think an optional register level for the DSP blocks is a good idea. And thanks for this pretty comprehensive PR!

Could you maybe adjust the name of the generic? `CPU_FAST_MUL_REG` is a bit shorter and matches the other generics (such as `XBUS_REGSTAGE_EN`). It would also ensure proper alignment with the other top generics. 😅

Maybe we should also add this option to the Vivado IP setup. That might help with timing closure there, too. I can take care of that if you like.

CodeWarrior1241:
> Hey @CodeWarrior1241,
> 
> I think an optional register level for the DSP blocks is a good idea. And thanks for this pretty comprehensive PR!
> 
> Could you maybe adjust the name of the generic? `CPU_FAST_MUL_REG` is a bit shorter and matches the other generics (such as `XBUS_REGSTAGE_EN`). It would also ensure proper alignment with the other top generics. 😅
> 
> Maybe we should also add this option to the Vivado IP setup. That might help with timing closure there, too. I can take care of that if you like.

Awesome, thank you for taking a look :)

I made the change to the generic and retested. No biggie.

I'd be happy to make the change to the Vivado IP setup also, that shouldn't be much... Would you consider me doing  that, but in a separate PR?

stnolting:
> I made the change to the generic and retested. No biggie.

Looks good! :+1:

> I'd be happy to make the change to the Vivado IP setup also, that shouldn't be much... Would you consider me doing that, but in a separate PR?

There are only two files to edit (with ~three lines each). So I think we can definitely add them here.

stnolting (approved):

CodeWarrior1241:
Gotcha on the Vivado IP setup, I made those changes and tossed them in for your approval.

stnolting (approved):

PR #1609, opened 2026-08-01
[cpu] remove reset from data path registers

stnolting:
All registers in the control path use an asynchronous reset. However, many registers in the data path now do not have a dedicated reset any more if their state is gated by a valid signal (and is therefore architecturally invisible while it is still undefined).

Without a dedicated reset, smaller FFs can be used (ASIC-only). In addition, the complexity of the reset network and its routing is simplified.

This is experimental.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 01.08.2026 | 1.13.3.3 | :test_tube: remove dedicated reset logic from CPU data path FFs | [#1609](https://github.com/stnolting/neorv32/pull/1609) |
 | 26.07.2026 | 1.13.3.2 | minor rtl edits and cleanups; :warning: simplify CFS bus interface | [#1606](https://github.com/stnolting/neorv32/pull/1606) |
 | 24.07.2026 | 1.13.3.1 | :bug: fix PMP `pmpaddr*[31:30]` bits: must be read-only and zero | [#1605](https://github.com/stnolting/neorv32/pull/1605) |
 | 24.07.2026 | [**1.13.3**](https://github.com/stnolting/neorv32/releases/tag/v1.13.3) | :rocket: **New release** | |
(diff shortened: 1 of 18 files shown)


tmeissner:
I assume one will get synthesis warnings in Synplify for these registers now when synthesizing for flash based Microchip FPGAs - but I have to check at work next week.

stnolting:
Hey @tmeissner,

I've never really worked with Flash-based FPGAs. Why do you think that would make a difference?

We already have FFs without a reset (instruction prefetch buffer, DSP-based multipliers, etc.). Have there been any problems with them so far?

tmeissner:
I remember messages more or less like "register X has no reset/preset port. Add reset/preset to improve timing ..." But I'm not sure if it were warnings or just notes. And if these come for all registers without (p)reset. And if it really influenced timing.

I'm not into neorv code base lately, so I'll check on monday.

Nevertheless, as we use different FPGA architectures, I always add explicit resets to registers with implicit sets (by initialization when declaring signals) in code which was written for SRAM FPGAs before.

stnolting:
> I remember messages more or less like "register X has no reset/preset port. 

But then the warning would be specific to that technology.

> I'm not into neorv code base lately, so I'll check on monday.

Thanks, I'm looking forward to it!

> I always add explicit resets to registers with implicit sets (by initialization when declaring signals) in code which was written for SRAM FPGAs before.

I've done that with FPGAs, too. But ASIC tools really don't like that at all, for obvious reasons. 😅

ekigwana:
> I remember messages more or less like "register X has no reset/preset port. Add reset/preset to improve timing ..." But I'm not sure if it were warnings or just notes. And if these come for all registers without (p)reset. And if it really influenced timing.
> 
> I'm not into neorv code base lately, so I'll check on monday.
> 
> Nevertheless, as we use different FPGA architectures, I always add explicit resets to registers with implicit sets (by initialization when declaring signals) in code which was written for SRAM FPGAs before.

Resets on registers definitely affect Xilinx FPGAs. I have a 500 MHz design and this certainly has an impact. Would be nice if this could be controlled via generic but I am firmly in the no reset camp for the technology that I am using. I do acknowledge that that is a separate religious war and not an attempt to start a new front.

stnolting:
> Resets on registers definitely affect Xilinx FPGAs. I have a 500 MHz design and this certainly has an impact.

In what way? Does the reset improve the timing or make it worse? And most importantly: Why? Is it the routing?

> Would be nice if this could be controlled via generic but I am firmly in the no reset camp for the technology that I am using.

I've been thinking about that, too... Maybe that would be the way to go if registers without a reset actually cause more problems than I thought.

```vhdl
process(rstn_i, clk_i)
begin
  if (rstn_i = '0') and RESET_EN then
    reg_q <= '0';
  elsif rising_edge(clk_i) then
    reg_q <= reg_d;
  end if;
end process;
```

But I have no idea if the (ASIC) synthesis would be able to resolve that properly.

> but I am firmly in the no reset camp for the technology that I am using.

ASIC or (SRAM/flash) FPGA? 

> I do acknowledge that that is a separate religious war and not an attempt to start a new front.

What about synchronous reset versus asynchronous reset? 🙈

PR #1620, opened 2026-08-08
[SoC] remove reset from registers where it is not strictly necessary

stnolting:
This PR extends #1609 (_remove reset from data path registers_) to the entire SoC:

* remove device's IRQ buffer reset (they have a synchronous reset via the module's "enable" flag)
* remove synchronizer reset (even with a reset these can have undefined values)
* remove clock gen async reset (they also have a sync reset via the module's enable flag)
* remove reset of FFs that have a sync reset (like the enable flag)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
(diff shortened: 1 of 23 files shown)


stnolting:
I'm wondering if we should perhaps also remove the bus response reset. That would save (depending on the configuration) about 400 FFs, which would allow to use smaller FFs (ASIC). 🤔

ekigwana:
> I'm wondering if we should perhaps also remove the bus response reset. That would save (depending on the configuration) about 400 FFs, which would allow to use smaller FFs (ASIC). 🤔

They are good candidates for removal if initial state is irrelevant to operations. The reset networks are just like clock and clock enable networks in that they add timing and routing pressure.

stnolting:
Good point, and thanks for confirming that!

Issue #1584, opened 2026-06-20
List of superlint warnings


short-stuff88:
Generated from Cadence Superlint 2026.3
Can't tell which you would want to fix, and which are deliberate and should be waived.
If you post the list of issues into two files:
1) those that should be waived (they're intended to be that way, or harmless)
2) those that you intend to fix (but not right now)

Then I'll update the Superlint scripts to reflect both.
As you fix issues, let me know - I'll update and see if any new issues arise.

[violations_detailed_basic.txt](https://github.com/user-attachments/files/29165694/violations_detailed_basic.txt)

stnolting:
Wow, thanks a lot!
That's quite a few warnings... 😅

I'll go through them and try to at least assess the different types.

stnolting:
I took another closer look at the warnings. They're just warnings, not critical errors. For me, I think the linter's requirements are very very strict (not even my VHDL professor was that strict). Anyway, here's my assessment.

I'd be super happy if more people would share their thoughts on these points here.


## CODINGSTYLE

> ARY_MS_DRNG: Inconsistent ordering of bits in range declarations in design-unit ... -- should be all Descending ranges

It seems like using `to` instead of `downto` is fine for memory arrays:

```vhdl
type rom_t is array (0 to 1023) of std_ulogic_vector(31 downto 0);
```

But bus "arrays" _should_ use `downto` instead of `to`. Example:

```vhdl
type port_bool_list_t is array (0 to 3) of boolean;
```

So, yeah, maybe it would be cleaner to change these to `downto`.


> ARY_NR_DFDR: In design unit 'neorv32_cpu_alu', bits of bus signal 'cp_result' have different types of drivers (/neorv32/rtl/core/neorv32_cpu_alu.vhd :147)
> ARY_NR_DFDR: In design unit 'neorv32_sysinfo', bits of bus signal 'sysinfo' have different types of drivers (/neorv32/rtl/core/neorv32_sysinfo.vhd :92)

I cannot identify the problem here. Is it complaining about assigning signals and constants to another signal array?!


> ARY_NR_LBND: Lower bound '1' of 'onehot' is not '0' in design-unit neorv32_cpu_regfile:neorv32_cpu_regfile_rtl

That is by design. Bit zero would correspond to register zero, which is not a physical register
(hardwired to zero).


> CAS_NR_DEFX: Signal ... assigned in case statement is not assigned to X in default case

Caused by code like this:

```vhdl
case req.addr(1 downto 0) is
[...]

PR #1622, opened 2026-08-09
Normalize bus/list array ranges to downto (ARY_MS_DRNG)

john-kearney:
Following up on #1584, this normalizes the non-memory array declarations from ascending `(0 to N)` to descending `(N downto 0)` range direction to clear the Superlint ARY_MS_DRNG warnings.

Memory/ROM arrays (`rom_t`, `code_rom_t`, `key_mem_t`, etc.) are kept as `to` since, as you noted, that is fine for memory.

A few arrays use positional aggregate initialization (`port_bool_list_t`, `port_tmo_en_list_t`, `dev_en_list_t`, `inst_c`, `ID`, etc.) so I have left those as `to` for now to avoid reversing the element mapping. Easy to do a follow-up that migrates them to named association and flips them too if you want.

33 declarations across 17 files. No functional changes, same indices and access patterns throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 14.08.2026 | 1.13.4.3 | normalize bus/list array ranges to `x downto y` | [#1622](https://github.com/stnolting/neorv32/pull/1622) |
 | 14.08.2026 | 1.13.4.2 | bitmanip: simplify register enables | [#1626](https://github.com/stnolting/neorv32/pull/1626) |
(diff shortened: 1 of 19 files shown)


stnolting:
Thanks for the PR!

> A few arrays use positional aggregate initialization (port_bool_list_t, port_tmo_en_list_t, dev_en_list_t, inst_c, ID, etc.) so I have left those as to for now to avoid reversing the element mapping. Easy to do a follow-up that migrates them to named association and flips them too if you want.

I think we should go ahead and adjust that as well, so we have everything in one PR.

I also think that the current implementation is causing errors, which is why the CI pipeline is failing. But I haven't looked into it in detail yet.

stnolting:
Thanks for the update!

Is this ready to be merged?
Should we adjust the remaining ranges in a subsequent PR?

john-kearney:
Yes, ready to merge. I'll do the remaining ranges (positional aggregate types) in a follow-up PR.

stnolting (approved):

PR #1623, opened 2026-08-11
[prim] declare FIFO/SPRAM memory arrays per generate-branch to fix RAM extraction on Verific-based synthesis tools

CodeWarrior1241:
While porting a NEORV32-based design to an Efinix Titanium (Efinity 2026.1), I
found that the processor's FIFO and SPRAM storage was being implemented in
discrete registers instead of block RAM, blowing up utilization.

The problem: in `neorv32_prim_fifo` and `neorv32_prim_spram` the memory array is
declared at architecture scope but driven from two mutually exclusive
`memory_large` / `memory_small` generate branches. Verific-based synthesis
front-ends (Efinity, and likely other tools using Verific elaboration) see a
signal with multiple processes driving it and don't map it to RAM — the
mutual exclusivity of the generate conditions is not considered.

My proposed fix: let's declare the array inside each generate branch instead. Since exactly one
branch is elaborated, this functions the same way and all tools see that only a single process
drives signals.

With this change Efinity extracts the FIFO/SPRAM memories to block RAM properly;
Xilinx/Microchip flows are unaffected (verified no change in inference there).
There's no functional change — things elaborate and the design boots and runs
identically in simulation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
+| 12.08.2026 | 1.13.4.1 | declare FIFO/SPRAM memory arrays per generate-branch to fix RAM extraction on Verific-based synthesis tools | [#1623](https://github.com/stnolting/neorv32/pull/1623) |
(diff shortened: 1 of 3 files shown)


stnolting:
That's an interesting fix!
I even didn't realize at all that you could declare (local) signals directly in a `generate` statement.

stnolting:
I did a quick synthesis using Vivado, Quartus, and Radiant (Synplify) and they all look fine. 👍

stnolting (approved):

CodeWarrior1241:
Awesome, thank you for the quick turnaround :)

PR #1625, opened 2026-08-13
[cpu] generalize fast multiplier pipeline

ekigwana:
## Background

This builds on #1391, which centralized multiplication in
`neorv32_prim_mul`, and #1603, which introduced `CPU_FAST_MUL_REG`.

The existing optional output register relies on synthesis retiming it into a
multi-DSP multiplier. This works well in some toolchains, but mapping and timing
remain vendor-dependent when native DSP blocks are narrower than the 33x33
signed product.

As discussed, this PR generalizes the boolean register option and adds an
explicitly pipelined implementation that avoids multi-DSP multiplier cascades.

## Summary

`CPU_FAST_MUL_REG` is replaced by:

```vhdl
CPU_FAST_MUL_REGS : natural range 1 to 3 := 1
```

The configurations are:

| `CPU_FAST_MUL_REGS` | Implementation | Primitive latency |
|:-------------------:|----------------|:-----------------:|
| `1` | Original inferred multiplier | 2 cycles |
| `2` | Original multiplier plus output register (`CPU_FAST_MUL_REG`) | 3 cycles |
| `3` | Explicit half-word partial-product pipeline | 4 cycles |

Thus, existing configurations map as follows:

- `CPU_FAST_MUL_REG = false` -> `CPU_FAST_MUL_REGS = 1`
- `CPU_FAST_MUL_REG = true` -> `CPU_FAST_MUL_REGS = 2`

The new three-register implementation uses:

```text
registered operands
  -> four registered half x half partial products
  -> two registered partial sums
  -> registered final sum
```

For `DWIDTH = 32`, each partial product is at most 17x17 signed. This fits one
native multiplier block on the mainstream FPGA families targeted here, avoiding
tool-dependent composition and retiming across multiple DSP blocks.

The data-path registers remain reset-free and CE-free, except for the
enable-controlled operand registers.

## Controller changes

`neorv32_cpu_alu_muldiv` derives the fast-multiplier latency from the selected
register count:

```vhdl
constant mul_latency_c : natural := FAST_MUL_REGS + 1;
```

The existing cycle counter handles the additional wait cycles without adding
more FSM states.

The documented ISA timing is:

```text
T_mul_latency = CPU_FAST_MUL_REGS
```

when `CPU_FAST_MUL_EN` is enabled.

## Latency contract

I also considered moving the latency contract into either:

- a shared package function; or
- a `done_o` strobe generated by `neorv32_prim_mul`.

Neither seemed worth the additional machinery in this change.

A package function would only wrap the simple `NUM_REGS + 1` relationship while
still requiring each consumer to implement its own scheduling. It would also
couple the otherwise generic primitive more closely to the NEORV32 package.

A primitive completion strobe would require validity state and reset handling.
More importantly, the integer M unit currently provides a one-cycle look-ahead
`valid_o`, whereas a natural multiplier `done_o` would indicate that the result
is valid in the current cycle. Adopting it cleanly would therefore require a
wider ALU handshake refactor or could add another multiplication cycle.

The latency remains documented at the primitive interface and derived locally
by the controller. The FPU continues to use the default one-register
configuration (`DWIDTH = 24`) and is unaffected.

## Integration and documentation

The renamed generic and its 1..3 range are propagated through:

- the CPU and processor top entities;
- the package component declaration;
- the simulation testbench;
- the Vivado IP wrapper and GUI; and
- the CPU, SoC, and ISA timing documentation.

## Verification

Completed locally:

- [x] Full GHDL compilation and elaboration
- [x] Full-system elaboration with `CPU_FAST_MUL_REGS = 1`, `2`, and `3`
- [x] Directed primitive tests for:
  - zero operands
  - `0x80000000 * 0x80000000`
  - negative signed x unsigned multiplication
  - `0xffffffff * 0xffffffff` unsigned
  - signed boundary combinations
- [x] 1,000 randomized reference cross-checks covering all signedness combinations
- [x] All three register configurations checked against the same plain signed-product reference
- [x] Controller result and completion timing checked for `MUL`, `MULH`, `MULHSU`, and `MULHU`
- [x] 10 ms full-system GHDL simulation with `CPU_FAST_MUL_REGS = 3`
- [x] `git diff --check`

Still outstanding / useful upstream testing:

- [x] RISC-V ACT tests for the M extension
- [ ] Post-route timing confirmation
- [x] Intel synthesis results
- [x] Lattice/Synplify synthesis results
- [ ] Microchip synthesis results

The focused primitive and controller checks used temporary local GHDL
testbenches and are not added to this PR.

## Synthesis evidence

On an AMD Zynq UltraScale+ RFSoC (`xczu48dr`, speed grade `-2`) with a 2.0 ns
clock constraint, Vivado absorbed each partial product into a DSP multiplier
register without attributes.

All multiplier paths had positive synthesis timing slack. The remaining
core-internal violations were in the prefetch-buffer-to-decode path, around
-0.3 ns, rather than in the multiplier.

The Intel, Microchip, and Lattice portability argument is currently structural:
each partial product fits one native multiplier block and no DSP post-adder
cascade is required. Measurements from those toolchains would still be very
valuable.

### DIFF
--- a/rtl/system_integration/neorv32_vivado_ip.tcl
+++ b/rtl/system_integration/neorv32_vivado_ip.tcl
@@ -310,10 +310,12 @@ proc setup_ip_gui {} {
   add_params $group {
     { CPU_CONSTT_BR_EN  {Constant-time branches} {Identical execution times for taken and not-taken branches} }
     { CPU_FAST_MUL_EN   {DSP-based multiplier}   {Use DSP block instead of bit-serial multipliers} }
-    { CPU_FAST_MUL_REG  {Multiplier register}    {Add a pipeline register to the DSP-based multiplier to improve timing closure} }
+    { CPU_FAST_MUL_REGS {Multiplier registers}   {Number of multiplier register stages (1..3)} }
     { CPU_FAST_SHIFT_EN {Barrel shifter}         {Use full-parallel shifters instead of of bit-serial shifters} }
     { CPU_RF_ARCH_SEL   {Register file style}    {Select implementation style of CPU register file} }
   }
+  set_property value_validation_range_minimum 1 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
+  set_property value_validation_range_maximum 3 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
   set_property widget {comboBox} [ipgui::get_guiparamspec -name "CPU_RF_ARCH_SEL" -component [ipx::current_core] ]
   set_property value_validation_type pairs [ipx::get_user_parameters CPU_RF_ARCH_SEL -of_objects [ipx::current_core]]
   set_property value_validation_pairs {{Block RAM} 0 {Distributed RAM} 1 {FFs with reset} 2 {Latches} 3} [ipx::get_user_parameters CPU_RF_ARCH_SEL -of_objects [ipx::current_core]]
--- a/rtl/system_integration/neorv32_vivado_ip.tcl
+++ b/rtl/system_integration/neorv32_vivado_ip.tcl
@@ -314,8 +314,6 @@ proc setup_ip_gui {} {
     { CPU_FAST_SHIFT_EN {Barrel shifter}         {Use full-parallel shifters instead of of bit-serial shifters} }
     { CPU_RF_ARCH_SEL   {Register file style}    {Select implementation style of CPU register file} }
   }
-  set_property value_validation_range_minimum 1 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
-  set_property value_validation_range_maximum 3 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
   set_property widget {comboBox} [ipgui::get_guiparamspec -name "CPU_RF_ARCH_SEL" -component [ipx::current_core] ]
   set_property value_validation_type pairs [ipx::get_user_parameters CPU_RF_ARCH_SEL -of_objects [ipx::current_core]]
   set_property value_validation_pairs {{Block RAM} 0 {Distributed RAM} 1 {FFs with reset} 2 {Latches} 3} [ipx::get_user_parameters CPU_RF_ARCH_SEL -of_objects [ipx::current_core]]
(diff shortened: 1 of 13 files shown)


rtl/system_integration/neorv32_vivado_ip.tcl
```diff
@@ -310,10 +310,12 @@ proc setup_ip_gui {} {
   add_params $group {
     { CPU_CONSTT_BR_EN  {Constant-time branches} {Identical execution times for taken and not-taken branches} }
     { CPU_FAST_MUL_EN   {DSP-based multiplier}   {Use DSP block instead of bit-serial multipliers} }
-    { CPU_FAST_MUL_REG  {Multiplier register}    {Add a pipeline register to the DSP-based multiplier to improve timing closure} }
+    { CPU_FAST_MUL_REGS {Multiplier registers}   {Number of multiplier register stages (1..3)} }
     { CPU_FAST_SHIFT_EN {Barrel shifter}         {Use full-parallel shifters instead of of bit-serial shifters} }
     { CPU_RF_ARCH_SEL   {Register file style}    {Select implementation style of CPU register file} }
   }
+  set_property value_validation_range_minimum 1 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
+  set_property value_validation_range_maximum 3 [ipx::get_user_parameters CPU_FAST_MUL_REGS -of_objects [ipx::current_core]]
```
stnolting: Do we actually need this? The range specified in the port declaration should actually limit that.

stnolting:
Thank you very much for the optimization and the detailed PR description!

> RISC-V ACT tests for the M extension

I tested it with `CPU_FAST_MUL_REGS = 3` and there were no errors.

> Intel synthesis results

Tested all 3 configuration options with Quartus. They all use the same number of DSP blocks; option 3 adds additional external adders.

> Lattice/Synplify synthesis results

Tested `CPU_FAST_MUL_REGS = 3` on an ice40up5k using Lattice Radiant (Synplify): correctly infers 4 DSP blocks.

stnolting (approved):

rtl/system_integration/neorv32_vivado_ip.tcl
ekigwana: Not strictly required and is an IP integrator effect. When changing options using the vivado gui this limits the range so the HDL never sees an incorrect value. Since the other options don't have it you can drop it.

I recommend you drop that bit but seeing how you pushed I cant easily amend - So I am leaving that bit to you.

PR #1626, opened 2026-08-13
[bitmanip] simplify register enables

### LINKED ISSUE #1617: [cpu/bitmanip] simplify operand/state register enables to shorten decode CE cone (500 MHz timing)
The cmd funct-decode (or_reduce_f(cmd)) sits in the clock-enable cone of the bitmanip unit's operand and FSM state registers.

At 500 MHz on a Zynq UltraScale+ RFSoC (2  ns clock), that cone — exec_reg[ir] → decode → ~70 register CEs — was our last timing-violating cluster. Two minimal edits close it: capture rs1/rs2/sha/less on  ctrl_i.alu_cp_alu alone (operands are stable during dispatch, and the FSM only consumes them for its own commands), and drop the redundant state <= S_IDLE self-assignment, which narrows the state register's load-enable. No latency or behavioral change; full regression passes, worst slack now positive.

[0001-neorv32-simplify-bitmanip-register-enables.patch](https://github.com/user-attachments/files/30728181/0001-neorv32-simplify-bitmanip-register-enables.patch)

ekigwana:
## Summary

This PR makes two small changes to the bit-manipulation unit's controller:

- capture `rs1`, `rs2`, `shamt`, and `less` on the common
  `ctrl_i.alu_cp_alu` dispatch trigger instead of the fully decoded
  `valid_cmd`; and
- remove the redundant `state <= S_IDLE` self-assignment for single-cycle
  operations.

## Motivation

The `or_reduce_f(cmd)` function-decode was part of the clock-enable cone for
the bit-manipulation unit's operand and FSM state registers.

At 500 MHz on a Zynq UltraScale+ RFSoC with a 2 ns clock constraint, the path
from `exec_reg[ir]` through instruction decode to approximately 70 register
clock enables was the final timing-violating cluster.

The operands are stable during the one-cycle ALU co-processor dispatch. The
bit-manipulation FSM only consumes the captured values when its own command
decode is valid, so capturing them for other ALU co-processor commands has no
behavioral effect. Likewise, omitting the explicit `S_IDLE` self-assignment
retains the current state through normal sequential semantics while reducing
the state register's load-enable logic.

There is no interface, behavior, or instruction-latency change.

With these changes, the previously failing register-enable paths have positive
timing slack at synthesis.

## Verification

- [x] Full maximum-ISA `processor_check` build and GHDL simulation
- [x] All enabled bit-manipulation extensions exercised: Zba, Zbb, Zbc, Zbs,
      Zbkb, Zbkc, and Zbkx
- [x] Processor check: 62/62 passed, 0 failed
- [x] `git diff --check`

Regression command:

```console
make RISCV_PREFIX=riscv32-unknown-elf- \
  USER_FLAGS+="-DUART0_SIM_MODE -DUART1_SIM_MODE" \
  clean_all info all sim-check
```

Closes #1617

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
-| 14.08.2026 | 1.13.4.2 | [bitmanip] simplify register enables | [#1626](https://github.com/stnolting/neorv32/pull/1626) |
+| 14.08.2026 | 1.13.4.2 | bitmanip: simplify register enables | [#1626](https://github.com/stnolting/neorv32/pull/1626) |
 | 12.08.2026 | 1.13.4.1 | declare FIFO/SPRAM memory arrays per generate-branch to fix RAM extraction on Verific-based synthesis tools | [#1623](https://github.com/stnolting/neorv32/pull/1623) |
 | 10.08.2026 | [**1.13.4**](https://github.com/stnolting/neorv32/releases/tag/v1.13.4) | :rocket: **New release** | |
 | 09.08.2026 | 1.13.3.9 | minor CPU updates, optimizations and RISC-V-compliance fixes | [#1621](https://github.com/stnolting/neorv32/pull/1621) |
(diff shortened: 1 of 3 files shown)


stnolting:
Thanks a lot!

stnolting (approved):

PR #1629, opened 2026-08-16
Add a new function to the UART for a transmission with a newline

### LINKED ISSUE #1426: neorv32_uart_puts does not add '\n' at the end of a string.
Hello, I was trying to adapt a program using functions provided in neorv32/sw/lib and noticed that the behaviour of the original `puts` function differs a little from the neorv32 equivalent (`neorv32_uart_puts`) because the original adds a '\n' character at the end. Not a big issue, but I thought I will let you know.

Kampi:


stnolting:
Good idea!
And I think this finally resolves #1426.

Thanks a lot!

stnolting (approved):

PR #1632, opened 2026-08-20
Convert remaining list arrays to downto (named-association aggregates)

john-kearney:
Follow-up to #1622: converts the arrays that were left ascending because their constants use positional aggregate initialization. Reversing the range direction would reverse positional element mapping, so the aggregates are migrated to named association at the same time, which makes the mapping independent of range direction.

Types converted:

- `port_bool_list_t`, `port_tmo_en_list_t`, `port_tmo_bit_list_t`, `dev_en_list_t`, `dev_base_list_t` (neorv32_bus)
- `dev_en_list_t`, `dev_base_list_t`, `dev_size_list_t` (xbus_gateway sim)
- `id_t` (psram_model sim)
- `inst_t` (cpu_trace instruction decode table, 231 entries)

One behavioral note: `decode_mnemonic_f` iterated `inst_c'range` and returns on first match, with the table ordered most-specific first and the catch-all `INVALID` entry last. The loop now uses an explicit ascending range (`0 to inst_c'length-1`) so the search order stays specific-first regardless of the array's direction. Without this, flipping the range would have made every instruction decode as INVALID.

Memory/ROM arrays and the arrays accessed via `'left`/`'right` in neorv32_top stay ascending, as established in #1622.

Verification: every constant was checked index-by-index against the previous positional form (scripted comparison, all 10 constants equivalent, including the 231-entry table). No `'left`/`'right`/`'range` attribute usage remains on any converted type.


stnolting:
Thank you very much for the follow-up PR!

> The loop now uses an explicit ascending range (0 to inst_c'length-1) so the search order stays specific-first regardless of the array's direction. Without this, flipping the range would have made every instruction decode as INVALID.

Good point!

> the arrays accessed via 'left/'right in neorv32_top stay ascending

We should adjust that for consistency as well. I can take care of that.

> Verification: 

:+1:

stnolting (approved):

PR #1643, opened 2026-09-03
test: add GPIO input loopback verification to processor_check

Abolfazl-Azad:
### Description

This PR improves GPIO verification coverage in the processor_check test suite.

The existing tests verify GPIO interrupt behavior, but the GPIO input read path was not explicitly checked.

This change adds a GPIO output-to-input loopback readback test using the existing simulation environment.

### Motivation

The default NEORV32 simulation connects GPIO outputs back to inputs. This allows verification that values written through the GPIO output interface are correctly sampled through the GPIO input interface.

This improves confidence in the GPIO peripheral verification without changing any RTL or software interfaces.

### Testing

- Built processor_check using the RISC-V GCC toolchain
- Executed the NEORV32 GHDL simulation flow
- Verified that the complete processor_check simulation passes successfully

### DIFF
--- a/sw/example/processor_check/main.c
+++ b/sw/example/processor_check/main.c
@@ -1578,10 +1578,20 @@ int main() {
 
   if (neorv32_gpio_available()) {
     neorv32_cpu_csr_write(CSR_MCAUSE, trap_never_c);
+
+    uint32_t gpio_pattern = 0b0101;
+    uint32_t gpio_mask = 0x0000000F;
+
     cnt_test++;
 
     gpio_trap_handler_ack = 0;
-    neorv32_gpio_port_set(0b0101);
+    neorv32_gpio_port_set(gpio_pattern);
+
+    // Test GPIO input readback via simulation loopback
+    neorv32_aux_delay_ms(neorv32_sysinfo_get_clk(), 1);
+    if ((neorv32_gpio_port_get() & gpio_mask) != gpio_pattern) {
+      test_fail();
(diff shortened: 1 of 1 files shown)


stnolting:
Hey @Abolfazl-Azad!

> This change adds a GPIO output-to-input loopback readback test using the existing simulation environment.

That's a good idea!
I think we could expand this to test the entire 32-bit input vector instead of just testing the lowest 4-bits. Maybe we could also test the direction control (e.g. only set some pins as output, the remaining ones should read `L` (= `0`) as we have "pull-down" in the testbench for all inputs).

Could you move your "test_fail" check into the final GPIO test check?

https://github.com/stnolting/neorv32/blob/e11079458b149828fbcc792c0e8124bc86f4a3b7/sw/example/processor_check/main.c#L1609-L1615

That way, we'll only get one error message even if both tests (input/output + interrupt) fail.

Abolfazl-Azad:
Thanks for the review and the helpful suggestions!

I have updated the PR to address all three points:

1. Expanded the GPIO input verification to cover the complete 32-bit vector.
2. Added GPIO direction-control verification using mixed input/output pins.
3. Moved the readback failure into the final GPIO test result so the GPIO test reports a single success/failure result.

I also verified the updated processor_check build and GHDL simulation successfully.

stnolting:
Great, thanks for the update!

stnolting (approved):

Issue #14, opened 2021-03-06
demo_gpio_irq uses firq channel 8. The following changes are needed in main.c to make it work:


stnolting:
demo_gpio_irq uses firq channel 8. The following changes are needed in main.c to make it work:

//install_err += neorv32_rte_exception_install(RTE_TRAP_FIRQ_1, gpio_pin_change_irq_handler);
  install_err += neorv32_rte_exception_install(RTE_TRAP_FIRQ_8, gpio_pin_change_irq_handler);

//install_err += neorv32_cpu_irq_enable(CSR_MIE_FIRQ1E);
  install_err += neorv32_cpu_irq_enable(CSR_MIE_FIRQ8E);

_Originally posted by @razzy34 in https://github.com/stnolting/neorv32/discussions/4#discussioncomment-434671_

stnolting:
Fixed.

Issue #13, opened 2021-02-28
Feature Request: Expanded CFS I/O


vhdlnerd:
Having only two 32-bit ports, one input and one output, for the CFS is a little limiting.  I know I can just modify the design and add whatever I want, But, it would be nice if the core allowed more CFS I/O without core modification.

**Some like:**
Add this type to neorv32_package.vhd:
`type cfs_io_t is array (integer range <>) of std_ulogic_vector(31 downto 0);`

Add new generics to neorv32_top:
```
    IO_CFS_IN_VECS							 : positive := 1;
    IO_CFS_OUT_VECS							 : positive := 1;`
```
Change CFS I/O ports to:
```
    cfs_in_i    : in  cfs_io_t(0 to IO_CFS_IN_VECS-1) := (others => (others => '0')); -- custom CFS inputs conduit
    cfs_out_o   : out cfs_io_t(0 to IO_CFS_OUT_VECS-1); -- custom CFS outputs conduit
```
And, of course, update the port hierarchy, for the CFS to match.

 This way end users could have as many CFS 32-bit I/O ports as they wished.

stnolting:
I should have seen it coming that 32-bit of CFS in/outs might be a limiting factor...

> `type cfs_io_t is array (integer range <>) of std_ulogic_vector(31 downto 0);`

I have to admit, this is an elegant solution!

However, personally I do not like having multi-dimensional array data types in the top entity. Also, I am not sure if high-level designs (like the Vivado IP packager) might struggle with these interface types.

How-about a trade-off? We could transform your [N][M] IO array into a single [N*M] array.

New generic for neorv32_top:
`CFS_IO_SIZE : natural := 32;`

And the modified CFS IO conduits:
```
cfs_in_i    : in  std_ulogic_vector(CFS_IO_SIZE-1 downto 0) := (others => '0');
cfs_out_o   : out std_ulogic_vector(CFS_IO_SIZE-1 downto 0);
```

But I am open for a discussion 😉

vhdlnerd:
Personally, I don't mind port types of multi-dimensional arrays (or arrays of records, record with arrays, records of records,... well, just about anything VHDL allows 😄) at any level of the hierarchy (except the true top that connects to FPGA I/O). But, that is just my coding style; I'm not saying it's better than anyone else's coding style. Plus, if you want to use a higher level IP stitching tool, it's best to avoid crazy port types.

[...]

Issue #11, opened 2021-02-19
UART RTS / CTS HW Flow Control


vhdlnerd:
It would be great if the UART ports supported HW flow control.  I would like to run the UARTs at much higher baud rates and transfer large amounts of data. This will require flow control.

BTW: Nice work on this project!  I had it up and running on a Digilent Nexys A7 board quickly without issues.

stnolting:
Interesting idea!

Obviously, the hardware flow control signals could be implemented using the GPIO ports and some software overhead. But this overhead might compromise some of the "very high" data throughput.

So having dedicated control signals seems to be a nice solution. I have checked the RTS / CTS specifications and if I got them right
`neorv32.uart.RTS_o` (output) would indicate the UART's receiver being ready to receive another byte while `neorv32.uart.CTS_i` (input) would indicate the UART's transmitter is allowed to transmit another byte, correct?

Right now I am not sure about the polarity of these signals, but this isn't really a big deal.

vhdlnerd:
Yes, that is the way I believe it works: CTS is an input and RTS is an output. I think they are high active signals (at least to the TTL level).
Thanks for considering this enhancement.

stnolting:
I have added (experimental) RTS/CTS hardware flow control with the last repository update. Since I have no full-scale UART-USB adapter available right now, I could only simulate the flow control so far - but it looks pretty good.

Maybe you can give some feedback if you plan to test the new features. :wink:

By the way, RTS and CTS seem to be low-active (they *are* low-active in the current NEORV32 version).

vhdlnerd:
I'll try it out this weekend and give you any feedback.

vhdlnerd:
It worked great for me on my Nexys A7 board -- RTS/CTS worked as expected.
I'm now running the UART at 460,800 baud, in the bootloader and app code, with no issues.

stnolting:
I'm glad to hear that! 👍

PR #10, opened 2021-02-17
Fix typo in the Readme

pcotret:

PR #1644, opened 2026-09-04
sw/floating_point_test: fix unreachable special values and use signaling NaN

Abolfazl-Azad:
### Description

This PR fixes two issues in the `floating_point_test` application that prevented complete IEEE-754 special value coverage.

1. **Unreachable special value cases**

The special value selection in `get_test_vector()` used the mask `& 0x3`, which limited the generated values to the range 0-3. As a result, cases 4-7 were unreachable and several predefined corner cases were never tested.

The mask has been changed to `& 0x7` so all eight special value cases can be selected.

2. **Signaling NaN test case**

Case 7 previously contained:

```c
case 7: tmp.float_value = NAN; break; // FIXME signaling_NAN?

### DIFF
--- a/sw/example/floating_point_test/main.c
+++ b/sw/example/floating_point_test/main.c
@@ -1018,7 +1018,7 @@ uint32_t get_test_vector(void) {
   // generate special value "every" ~256th time this function is called
   if ((neorv32_aux_xorshift32() & 0xff) == 0xff) {
 
-    switch((neorv32_aux_xorshift32() >> 10) & 0x3) { // random decision which special value we are taking
+    switch((neorv32_aux_xorshift32() >> 10) & 0x7) { // random decision which special value we are taking
       case  0: tmp.float_value  = +INFINITY; break;
       case  1: tmp.float_value  = -INFINITY; break;
       case  2: tmp.float_value  = +0.0f; break;
@@ -1026,7 +1026,7 @@ uint32_t get_test_vector(void) {
       case  4: tmp.binary_value = 0x7fffffff; break;
       case  5: tmp.binary_value = 0xffffffff; break;
       case  6: tmp.float_value  = NAN; break;
-      case  7: tmp.float_value  = NAN; break; // FIXME signaling_NAN?
+      case  7: tmp.binary_value = FLOAT32_SNAN; break;
       default: tmp.float_value  = NAN; break;
     }
   }

### CODE REVIEW

**Comment by stnolting** [comment]
I'm not sure if these were just typos, or if we _actually_ disabled them... 🤔
Have you been able to test the new corner cases yet?

Btw, this PR also includes the changes from #1643. Is that intentional? If not, could you please remove it?

**Comment by Abolfazl-Azad** [comment]
Thanks for the feedback!

I removed the unrelated changes from #1643 and updated the PR so it now contains only the floating-point test changes.

I also verified that the selector now reaches all eight special-value cases and confirmed that `FLOAT32_SNAN` is the existing signaling-NaN test operand used by the example.

The local build succeeds. My local environment could not complete the full `floating_point_test` GHDL run because the available RISC-V toolchain uses a different runtime/startup configuration from the project's CI toolchain.

**Comment by stnolting** [comment]
Thank you for the updates!

I've checked the floating-point tests and they still work - so your extended test cases didn't discover any errors (phew, luckily 😅). 

```
NEORV32 Zfinx ISA extension (FPU) test

SILENT_MODE enabled (only showing actual errors)
Test cases per instruction: 1000000
NOTE: The NEORV32 FPU does not support subnormal numbers yet. Subnormal numbers are flushed to zero.


#0: FCVT.S.WU (unsigned integer to float)...
Errors: 0/1000000 [ok]

#1: FCVT.S.W (signed integer to float)...
Errors: 0/1000000 [ok]

#2: FCVT.WU.S (float to unsigned integer)...
Errors: 0/1000000 [ok]

#3: FCVT.W.S (float to signed integer)...
Errors: 0/1000000 [ok]

#4: FADD.S (addition)...
Errors: 0/1000000 [ok]

#5: FSUB.S (subtraction)...
Errors: 0/1000000 [ok]

#6: FMUL.S (multiplication)...
Errors: 0/1000000 [ok]

#7: FMIN.S (select minimum)...
Errors: 0/1000000 [ok]

#8: FMAX.S (select maximum)...
Errors: 0/1000000 [ok]

#9: FEQ.S (compare if equal)...
Errors: 0/1000000 [ok]

#10: FLT.S (compare if less-than)...
Errors: 0/1000000 [ok]

#11: FLE.S (compare if less-than-or-equal)...
Errors: 0/1000000 [ok]

#12: FSGNJ.S (sign-injection)...
Errors: 0/1000000 [ok]

#13: FSGNJN.S (sign-injection NOT)...
Errors: 0/1000000 [ok]

#14: FSGNJX.S (sign-injection XOR)...
Errors: 0/1000000 [ok]

#15: FCLASS.S (classify)...
Errors: 0/1000000 [ok]

[Zfinx extension verification successful!]
```

stnolting (approved):

PR #1645, opened 2026-09-05
🐛 [axi bridge] full rework; add data buffering to handle back pressure

stnolting:
The current XBUS-to-AXI bridge works well for single accesses. However, problems can arise with bursts. For example, write bursts cannot be paused, which means the bridge cannot handle back pressure.

This pull request is a complete redesign of the AXI bridge. Among other things, it now features data buffering (FIFO).

I have tested the bridge using AMD Vivado: high bus utilization using AXI Traffic Generator IP and protocol verification using the AXI Protocol Checker IP. Anyway, it would be great if someone else could test the bridge in their own design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mimpid = 0x01040312 -> Version 01.04.03.12 -> v1.4.3.12
 
 | Date | Version | Comment | Ticket |
 |:----:|:-------:|:--------|:------:|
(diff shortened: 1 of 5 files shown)


stnolting:
Tested on a Digilent Nexys A7 board. Programs are executed from DDR2 memory (unsing Vivado's MIG). CPU caches enabled, AXI bursts enabled.

* [x] CoreMark
* [x] FreeRTOS
* [x] MicroPython

Seems to run stable; no protocol errors detected (by Vivado AXI Protocol Checker IP).

PR #1648, opened 2026-09-11
[cpu] improve timing margin without added latency

ekigwana:
### Summary

Improve timing margin across CSR readback, ALU result capture, multiplier partial sums and instruction-fetch control.
The changes redistribute combinational work around existing register edges without adding instruction cycles.

### Standalone μC results

One comparison: unchanged upstream versus this complete branch. Both use the same standalone CPU harness,
Vivado 2026.1, `xczu48dr-ffvg1517-2-e` and a 2.000 ns clock constraint.
There is no application RAM, interconnect or peripheral logic in the synthesis top.

| Metric | Upstream | This branch | Difference |
| --- | ---: | ---: | ---: |
| μC register-to-register WNS | +0.228 ns | +0.367 ns | **+0.139 ns** |
| Setup endpoints below +0.500 ns | 200 | 39 | **-161 (80.5%)** |
| Setup endpoints below +0.300 ns | 63 | 0 | **-63** |
| LUTs | 2,894 | 3,088 | **+194 (6.7%)** |
| FFs | 2,286 | 2,434 | **+148 (6.5%)** |
| BRAM tiles | 0 | 0 | 0 |
| DSPs | 4 | 4 | 0 |
| Latches | 0 | 0 | 0 |

Both designs have zero unconstrained internal endpoints, missing I/O delays and combinational loops.
The remaining worst path is instruction decode to the bitmanip FSM clock enable.
These are synthesis timing estimates, not hardware timing sign-off.

Configuration: RV32I with M, Zba, Zbb, Zbkc, Zbs, Zicntr and Zicond enabled.
RF architecture 2, fast shifting and a three-stage fast multiplier are enabled.
Compressed instructions, constant-time branches, user mode, HPM, PMP and debug are disabled.
CPU buses and interrupt inputs are exposed. Maximum input/output delay is 0.200 ns, minimum delay is zero.
WNS and endpoint counts above cover internal register-to-register setup paths, excluding external I/O timing paths.
One worst path per endpoint is reported, with a 100,000-path limit.

### 1. [cpu/control] separate local and external CSR decode

**Justification.** The CSR read register receives both local CSR data and external contributions, including counters.
Explicit mutually exclusive local/external decode exposes their selection before the existing register
and reduces serial selection on the CSR read path.

**Contract.** Preserve the local address set and decode priority.
Disabled local CSRs return zero rather than admitting external data.
The read-enable, idle-zero output and sampling edge are unchanged.

**Measured CSR-path margin for the complete branch:** +0.228 → +0.441 ns, **+0.213 ns**.

**Optional counter-readback extension.** Select each counter's requested 32-bit word before read-enable masking,
and decode counter reads with full CSR address comparisons. This narrows the readback network and reduces serial address qualification
without changing counter state or the CSR sampling edge.
Measured with this branch, the extension raises CSR margin from +0.441 to +0.583 ns (**+142 ps**)
for **28 additional LUTs and one FF**, and reduces endpoints below +0.500 ns from 39 to seven.
Overall μC WNS remains +0.367 ns. This offers additional CSR margin where that resource tradeoff is useful.

### 2. [cpu/alu_bitmanip] separate result registers

**Justification.** A shared OR reduction combines logic, shift/count and shifted-add results before capture.
Partitioning the result registers by these groups removes part of that reduction from individual capture paths.
The groups are combined after the existing edge, distributing the work across the same register boundary.

**Contract.** Preserve operand capture, live decode, iterative-operation control, valid timing and idle-zero behavior.
Every original OR term remains represented, including overlapping selections. No instruction cycle is added.

**Measured bitmanip-path margin for the complete branch:** +0.278 → +0.367 ns, **+0.089 ns**.

**Tradeoff.** Separate result groups require additional storage and output combination logic.
The remaining worst bitmanip path is instruction decode to the FSM clock enable.

### 3. [cpu/alu_cond] separate result data and control

**Justification.** The conditional move decision otherwise controls selection into a 32-bit result register.
Capture raw data and the one-bit move decision separately, then select data or zero.
This removes the late decode/condition expression from the wide register input selection.

**Contract.** Preserve accepted results, valid timing and the result capture edge.
The unqualified output before the first clock changes from unknown to zero.
The data register captures during idle cycles, which can change switching activity.

**Measured conditional-path margin for the complete branch:** +0.804 → +0.952 ns, **+0.148 ns**.

**Incremental cost.** In the combined μC, this change adds **one FF and no LUTs**.
With all other changes held fixed, conditional-path margin increases from +0.655 to +0.952 ns (**+297 ps**).
Overall WNS remains +0.367 ns, with 39 endpoints below +0.500 ns. Idle data-register switching can change.

### 4. [cpu/frontend] localize prefetch buffer writes

**Justification.** Dynamic indexed writes to the small prefetch buffer can infer storage with less favorable input timing.
Static per-entry write processes expose local write selection without a RAM-style attribute.
This form supports integration with instruction memories whose output-to-buffer path is critical.

**Contract.** Preserve buffer depth, pointer updates, read selection, clear behavior and fetch handshakes.
Normalize the write address once while retaining the unknown-pointer fallback.
Each halfword buffer remains two entries deep. No additional prefetch capacity or fetch cycle is introduced.

**Observation.** The isolated standalone μC measurement showed no WNS or below-target endpoint improvement and increased FF usage.
This harness has no instruction RAM, so it does not establish the intended memory-to-prefetch timing benefit.

**RAM-attached result.** With 512 KiB of synchronous, byte-write-enabled true-dual-port RAM in READ_FIRST mode,
mapped to 128 AMD UltraScale+ RAMB36E2 primitives, RAM-to-prefetch margin improved from +0.134 to +0.423 ns (**+289 ps**).
The RAM uses a one-cycle read with no optional output register. Fetch latency is unchanged.
The prefetch change alone used **82 fewer LUTs and 59 additional FFs**, with unchanged BRAM count.

**RAM-enable integration.** The changed mapping increased loading on the request-qualified RAM enables.
Holding both RAM read enables high removes that enable-path regression, while qualifying each byte-write enable
with the original request enable preserves accepted writes. Together, the prefetch and RAM-enable changes retain
the **+289 ps RAM-to-prefetch gain** at a net cost of **59 FFs and 65 fewer LUTs**, with unchanged BRAM count.
Overall RAM-attached WNS remains +0.132 ns. The RAM-enable adjustment belongs in the external RAM wrapper and is not part of this PR.
It preserves accepted read/write and acknowledgement timing, but allows idle read data to change and increases idle RAM switching.

### 5. [prim] use carry-save multiplier partial sums

**Justification.** Adding aligned partial products with carry-propagate arithmetic creates long partial-sum capture paths.
The low-low and shifted high-high terms occupy non-overlapping bit positions and can be combined directly.
Compress that term and the two cross products into sum and shifted carry, then retain the existing final addition.

**Contract.** Preserve the upstream operand split, signed widths, input-enable behavior and three-stage latency.
Other register-count configurations are unchanged. No extra DSP stage or synthesis attribute is introduced.

**Measured multiplier-path margin for the complete branch:** +0.238 → +0.625 ns, **+0.387 ns**.

**Tradeoff.** Carry-save partial sums use additional **+119 LUTs and +14 FFs**. DSP count remains four.

### 6. [cpu/regfile] capture comparisons at the read edge

**Justification.** Comparing registered operands and then driving fetch/control decisions puts comparison and control decode in one cycle.
For RF architecture 2, the raw operands are already available before the RF read-output edge.
Capture equality, signed/unsigned less-than and immediate equality at that edge, then select the appropriate flag in control.

**Contract.** Comparison flags sample the same register contents and edge as operand data.
Preserve RV32E address truncation, x0 handling and the full five-bit Zibi immediate selector.
Signedness selection occurs after capture.
The RF gains a four-bit internal output, but the CPU external interface is unchanged.
Other RF architectures continue using the ALU comparison path. No branch cycle or operand-data stage is added.

**Measured frontend-path margin for the complete branch:** +0.295 → +0.653 ns, **+0.358 ns**.

**Tradeoff.** Comparison capture adds comparison logic before the RF read-output edge and stores four flags.
In the complete branch, minimum RF destination slack is +0.501 ns.

### Interpretation

Each measurement identifies its comparison: upstream versus the complete branch, a patch applied alone, or the combined design with and
without that patch. Resource costs and timing gains apply to the stated comparison.

### Validation

- Fresh standalone synthesis completed for upstream and the revised six-optimization branch.
  Earlier isolated and combined runs support the observations above.
- The revised run reported zero critical warnings, errors, inferred latches and timing-check deficiencies.
- The branch's seven changed RTL files and restored upstream counter file match the revised synthesis snapshot.
- Prior validation of the optimizations includes processor_check: 62/62 passed with RF architecture 2 and three multiplier stages.
- The real-CPU operand test passed 64 configurations and 128,128 independent signature checks per source set,
  with matching public-bus and retirement traces against its reference.
- Multiplier differential tests passed 45 configurations and 184,230 checked cycles.
- Counter and bitmanip generic elaboration and focused arithmetic/control identity checks passed.
- μC integration pre/post-synthesis Riviera tests passed. XSim pre/post-synthesis passed.

### DIFF
--- a/rtl/core/neorv32_cpu.vhd
+++ b/rtl/core/neorv32_cpu.vhd
@@ -134,6 +134,8 @@ architecture neorv32_cpu_rtl of neorv32_cpu is
   signal alu_res     : std_ulogic_vector(31 downto 0); -- ALU result
   signal alu_add     : std_ulogic_vector(31 downto 0); -- ALU address result
   signal alu_cmp     : std_ulogic_vector(1 downto 0);  -- comparator result
+  signal branch_cmp  : std_ulogic_vector(1 downto 0);
+  signal rf_cmp      : std_ulogic_vector(3 downto 0);
   signal alu_cp_done : std_ulogic;                     -- ALU co-processor operation done
   signal lsu_rdata   : std_ulogic_vector(31 downto 0); -- LSU memory read data
   signal lsu_mar     : std_ulogic_vector(31 downto 0); -- LSU memory address register
@@ -306,7 +308,7 @@ begin
     hwtrig_i      => hwtrig,      -- hardware trigger
     -- data path interface --
     alu_cp_done_i => alu_cp_done, -- ALU iterative operation done
-    alu_cmp_i     => alu_cmp,     -- comparator status
+    alu_cmp_i     => branch_cmp,  -- comparator status
     alu_add_i     => alu_add,     -- ALU address result
     rf_rs1_i      => rs1,         -- register file source 1
     csr_rdata_o   => csr_rdata,   -- CSR read data
@@ -416,8 +418,20 @@ begin
     rs1_data_o => rs1,
     -- read port 1 (rs2) --
     rs2_addr_i => ctrl.rf_rs2,
-    rs2_data_o => rs2
+    rs2_data_o => rs2,
+    cmp_o      => rf_cmp
   );
+
+  rf_compare:
+  if (CPU_RF_ARCH_SEL = 2) generate
+    branch_cmp(0) <= rf_cmp(3) when RISCV_ISA_Zibi and (ctrl.ir_funct3(2 downto 1) = "01") else rf_cmp(0);
+    branch_cmp(1) <= '0' when is_x(ctrl.alu_unsigned) else rf_cmp(1) when (to_x01(ctrl.alu_unsigned) = '1') else rf_cmp(2);
+  end generate;
+
+  alu_compare:
+  if (CPU_RF_ARCH_SEL /= 2) generate
+    branch_cmp <= alu_cmp;
+  end generate;
 
   -- all buses are zero unless there is an according operation --
   rf_wdata <= alu_res or lsu_rdata or csr_rdata or ctrl.pc_ret;
--- a/rtl/core/neorv32_cpu_alu_bitmanip.vhd
+++ b/rtl/core/neorv32_cpu_alu_bitmanip.vhd
@@ -182,6 +182,7 @@ architecture neorv32_cpu_alu_bitmanip_rtl of neorv32_cpu_alu_bitmanip is
   type res_t is array (op_width_c-1 downto 0) of std_ulogic_vector(31 downto 0);
   signal res_int, res : res_t;
   signal xperm4_res, xperm8_res, adder_res, one_hot_res, zip_res, unzip_res, res_out : std_ulogic_vector(31 downto 0);
+  signal logic_reg, shift_reg, add_reg : std_ulogic_vector(31 downto 0);
 
 begin
 
@@ -581,7 +582,9 @@ begin
   begin
     tmp_v := (others => '0');
     for i in 0 to op_width_c-1 loop
-      tmp_v := tmp_v or res(i);
+      if (i /= op_shadd_c) and (i /= op_cz_c) and (i /= op_cpop_c) and (i /= op_rot_c) then
+        tmp_v := tmp_v or res(i);
+      end if;
     end loop;
     res_out <= tmp_v;
   end process;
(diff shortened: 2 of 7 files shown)


stnolting:
Hey @ekigwana!

Thanks for another excellent pull request! I really appreciate it when someone digs so deeply into the micro architecture!

I'm just now getting around to writing this because I wanted to test your changes myself first. However, I'm currently having some toolchain and licensing issues, so I can't use my usual test platform right now to test your code edits.

What do you think about splitting up some of the optimizations across individual PRs?

I fully agree with timing optimizations that don’t affect (or positively affect) the circuit’s size. For all other optimizations, I understand the reasoning, but I’m skeptical for now. 😉 The small size is essentially the core selling point of this project. So I’m hesitant for now when it comes to using additional hardware resources.

As I said, I cannot asses the impact of your changes here in my test environment right now. I'll come back to the individual points once my setup is running again.

Issue #897, opened 2024-05-10
Latency measurement through `CSR(MCYCLE)` adds one extra cycle


Unike267:
Hi @stnolting !

Back here again 😅

I'm performing a CFU (Custom Function Unit) latency measurement using the `CSR(MYCYCLE)`.

For this purpose I've added a multiplier with NEORV32 through this mode of connection.

In this context I've made the following code in C:

```c
  //0000000000000001 x 0000000000000001
  static uint32_t fir = 0x00010001;
  //0000000000000010 x 0000000000000010
  static uint32_t sec = 0x00020002;
  //0000000000000100 x 0000000000000100
  static uint32_t thi = 0x00040004;
  //0000000000001000 x 0000000000001000
  static uint32_t fou = 0x00080008;

  // Intro
  neorv32_uart0_printf("\n CFU-mpw \n");

  // Perform 4 multiplication through custom instruction (funct3=001)
    neorv32_cpu_csr_write(CSR_MCYCLE, 0);
    neorv32_cfu_r3_instr(0b1111111, 0b001, fir, 0); 
    neorv32_cpu_csr_read(CSR_MCYCLE); 

    neorv32_cpu_csr_write(CSR_MCYCLE, 0);
    neorv32_cfu_r3_instr(0b1111111, 0b001, sec, 0); 
    neorv32_cpu_csr_read(CSR_MCYCLE); 

    neorv32_cpu_csr_write(CSR_MCYCLE, 0);
    neorv32_cfu_r3_instr(0b1111111, 0b001, thi, 0); 
    neorv32_cpu_csr_read(CSR_MCYCLE); 

    neorv32_cpu_csr_write(CSR_MCYCLE, 0);
    neorv32_cfu_r3_instr(0b1111111, 0b001, fou, 0); 
    neorv32_cpu_csr_read(CSR_MCYCLE); 

  // End
  neorv32_uart0_printf("\nEND\n");
```

*Note: in the custom instruction the two 16-bit input operators are contained in `rs1`, `rs2` don't care and the 32-bit result is stored in `rd`.*

Obtaining the following assembly code:

```asm
  asm volatile ("csrw %[input_i], %[input_j]" :  : [input_i] "i" (csr_id), [input_j] "r" (csr_data));
 214:	00000793          	li	a5,0
 218:	b0079073          	csrw	mcycle,a5
    neorv32_cfu_r3_instr(0b1111111, 0b001, fir, 0); 
[...]

PR #1654, opened 2026-09-20
[counter CSRs] do not increment during CSR-write

stnolting:
Counter CSRs (`cycle`, `instret`, and HPMs) should not be automatically incremented (as a result of a side effect, such as a retired instruction) if a write access to the corresponding CSR is occurring at the same time.

RISC-V unpriv spec. (section "6.1. CSR Instructions"):

> Some CSRs, such as the instructions-retired counter, instret, may be modified as side effects of instruction execution. In these cases, if a CSR access instruction reads a CSR, it reads the value prior to the execution of the instruction. **If a CSR access instruction writes such a CSR, the explicit write is done instead of the update from the side effect.** In particular, a value written to instret by one instruction will be the value read by the following instruction.


Unike267:
Hey Stephan!

How's it going? It's been a while since I've bothered you haha (I'm still busy with the thesis).

I just wanted to check out of curiosity if this change has anything to do with what's described in [github.com/stnolting/neorv32/issues/897.](https://github.com/stnolting/neorv32/issues/897)

Cheers!

PR #1655, opened 2026-09-20
TWD start/end status; clear only if set

### LINKED ISSUE #1652: Potential loss of status in neorv32_twd_com_started/ended
https://github.com/stnolting/neorv32/blob/da5f595a3528c524eaabf712ec8823cc43375b96/sw/lib/source/neorv32_twd.c#L151

Hello,

This is not a result of any sort of code review; I was just debugging my own broken code and poking around to see what was (supposed) to happen. I did not observe any real misbehaviour, it is just an observation from reading the code.

My doubt with these functions is clearing the status bits: Is it not the case that we read NEORV32_TWD->CTRL twice, first when obtaining the status and second time when clearing it? If it would happen that the status was set just at the correct/unfortunate moment, between the two reads, we would clear it and never observe it.

I am fairly new here, so I am not sure whether I should have just proposed a fix instead of opening an issue - feel free to let me know what is the preferred way of proposing such minor fixes/improvements.

Thanks for all the nice work, really easy to use and trace things around.

ANicheMold:
closes https://github.com/stnolting/neorv32/issues/1652


ANicheMold:
@stnolting, please let me know if this is not how contributions should be done. Or if you disagree with the fix.

